PR #14, opened 2015-01-27
Escape code example within javadoc

xerial:
This fix is necessary to make `./gradlew build` work with Java8.


frsyuki:
Yes, thank you!

PR #16, opened 2015-01-27
TimeStamp.fromString should be public

xerial:
Plugin needs to access this method to create Timestamp from strings.


frsyuki:
In which case you need `Timestamp.fromString`? Why don't you use TimestampParser?

xerial:
I failed to notice there is TimestampParser.

xerial:
My expectation was there is some factory method like `Integer.parseInt(str)'

xerial:
What is the difference of `Timestamp.fromString` and `TimestampParser

PR #17, opened 2015-01-27
Add description field to @Config annotation

xerial:
It would be good if we can add descriptions to `@Config` parameters: 
```
@Config("param", description="parameter description")
```
In future, we can automatically generate documentation of plugin configurations by extracting these annotations.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/config/Config.java
+++ b/embulk-core/src/main/java/org/embulk/config/Config.java
@@ -12,4 +12,9 @@ import java.lang.annotation.Target;
 public @interface Config
 {
     String value();
+
+    /**
+     * Description of this configuration parameter
+     */
+    String description() default "";
 }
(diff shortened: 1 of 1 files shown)


frsyuki:
Sounds good idea to have description!
hmm...but "description" is a bit long. Do you think adding description parameter to `@Config` is better than adding `@ConfigDesc` annotation? Because we already have `@ConfigDefault`.

xerial:
Using `desc` instead of `description` is fine. Defining `desc` as a parameter is better than having `@ConfigDesc` since IDE is good at finding `desc` inside `@Config`.

frsyuki:
That's good to know > IDE support.
Isn't it `@Config(value="foo", desc="bar")` by the way? I think we can't omit value= when there're 2 or more parameters.

xerial:
Ah. Yes. This is a little bit annoying. I will consider a better way. Thanks.

PR #43, opened 2015-01-29
Fix gem name.

takashiyamazaki:
fix gem name "embulk-plugin-postgres-json"


frsyuki:
Good catch! Thank you.

Issue #25, opened 2015-01-29
Resuming failed transaction


frsyuki:
A bulk load likely fails when we import large amount of data (say 1TB). When it fails, Embulk currently requires plugins to rollback entire transaction so that we can retry.
However, retrying the entire bulk load of the large data takes too long time. It's very good if we can resume the failed transaction from partially succeeded state.

Difficulties:

* What's the API for plugins?
* Is a input plugin responsible to skip tasks? Or, is a output plugin responsible to skip records?

xerial:
This feature would be nested transaction, with which we can partially roll back sub tasks.

frsyuki:
@xerial What do you mean nested transaction? Could you explain with an example scenario?

frsyuki:
I'm writing some code at https://github.com/embulk/embulk/tree/resume branch.

frsyuki:
#50 implemented this feature.

PR #50, opened 2015-02-02
Implementing resume and cleanup

frsyuki:
#25
With `-r` option, `run` subcommand writes execution state to a file if the transaction is partially failed:

```
embulk run config.yml -r resume.yml
```

We can retry it using the same command-line.
If we don't want to retry it, use `cleanup` subcommand:

```
embulk cleanup config.yml -r resume.yml
```

If we don't set `-r` option, cleanup is executed automatically.


muga:
It looks good to me.

frsyuki:
thank you. but already merged :-p

PR #53, opened 2015-02-04
Add Enumerable behavior to Page

niku:
We have already implement Page#each.
So, we could use other convenient methods if including Enumerable.


frsyuki:
That's cool idea! Now we can use page.map, page.each_with_index, etc etc.

frsyuki:
Merged. Thank you!

PR #65, opened 2015-02-10
add "preview = yes" to config at preview time

yaggytter:
I want to know it is now preview mode in the input plugin.
for example, such an implementation would you like?
or other better way?


frsyuki:
I think better idea is to add an API to org.embulk.spi.Exec (ExecSession). Exec class is an interface to provide runtime environment information.

yaggytter:
oh, thank you for teaching me!

PR #70, opened 2015-02-14
Minor fixed plugin templates

muga:
* fixed license declarations within the templates of java/build.gradle and gemspec.
* changed the method bodies within the templates of Java plugin classes. By the changing, the generated Java plugin classes can be compiled.


frsyuki:
Good pull-request :+1:

Issue #72, opened 2015-02-16
"new ruby-input" generator does not work properly.


hiroyuki-sato:
"new ruby-input" generator does not work properly.

java -jar ./embulk-0.4.1.jar new ruby-input sample

```
% java -jar ./embulk-0.4.1.jar new ruby-input sample
Creating embulk-input-sample/
  Creating embulk-input-sample/README.md
  Creating embulk-input-sample/LICENSE.txt
  Creating embulk-input-sample/.gitignore
  Creating embulk-input-sample/Rakefile
  Creating embulk-input-sample/Gemfile
  Creating embulk-input-sample/embulk-input-sample.gemspec
  Creating embulk-input-sample/lib/embulk/input/sample.rb
```

rake build (and failed.)

```
% cd embulk-input-sample/
% rake build 
rake aborted!
There was a SyntaxError while loading embulk-input-sample.gemspec: 
/home/embulk/embulk-input-sample/embulk-input-sample.gemspec:7: syntax error, unexpected tIDENTIFIER, expecting keyword_end
...h the output plugins by "embulk-output" keyword."
...                               ^
/home/arch/embulk-input-sample/embulk-input-sample.gemspec:7: syntax error, unexpected tSTRING_BEG, expecting keyword_do or '{' or '('
...tput plugins by "embulk-output" keyword."
...                               ^
/home/embulk/embulk-input-sample/Rakefile:1:in `<top (required)>'
(See full trace by running task with --trace)
```

Can't use double quote "embulk-output" in gemspec file.
```ruby
  spec.description   = "Sample input plugin is an Embulk plugin that loads records from Sample so that any output plugins can receive the records. Search the output plugins by "embulk-output" keyword."
```

Use single quote better
```ruby
  spec.description   = "Sample input plugin is an Embulk plugin that loads records from Sample so that any output plugins can receive the records. Search the output plugins by 'embulk-output' keyword."
```

And one more thing. 

```ruby
module Embulk
  module Input

    class SampleInputPlugin < InputPlugin
      Plugin.register_input(sample, self)
```

need single quote??
```ruby
module Embulk
  module Input

    class SampleInputPlugin < InputPlugin
      Plugin.register_input('sample', self)
```

frsyuki:
Fixed by #73. Thank you!

PR #74, opened 2015-02-16
fix ERB templates

takei-yuya:
fix ERB templates


frsyuki:
Thank you! Fixed [#72]

PR #82, opened 2015-02-18
Maked log level configurable by command line argument

ykubota:
Embulk always sets "INFO"  to the log level of rootLogger.
This PR makes log level configurable by command line arguments as below.

- java -jar pkg/embulk.jar run example.yml -l &lt;LEVEL&gt;
- java -jar pkg/embulk.jar run example.yml --log-level &lt;LEVEL&gt;
- java -Dembulk.logLevel=&lt;LEVEL&gt; -jar pkg/embulk.jar run example.yml

In addition, users can specify the log level by system property: embulk.logLevel.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/LoggerProvider.java
+++ b/embulk-core/src/main/java/org/embulk/exec/LoggerProvider.java
@@ -4,17 +4,21 @@ import java.util.Properties;
 import org.slf4j.ILoggerFactory;
 import org.slf4j.LoggerFactory;
 import org.apache.log4j.PropertyConfigurator;
+import com.google.inject.Inject;
 import com.google.inject.Provider;
+import org.embulk.config.ConfigSource;
 
 public class LoggerProvider
         implements Provider<ILoggerFactory>
 {
-    public LoggerProvider()
+    @Inject
+    public LoggerProvider(@ForSystemConfig ConfigSource systemConfig)
     {
         // TODO system config
         Properties prop = new Properties();
 
-        prop.setProperty("log4j.rootLogger", "INFO,root");
+        String logLevel = systemConfig.get(String.class, "logLevel","INFO");
+        prop.setProperty("log4j.rootLogger", logLevel+",root");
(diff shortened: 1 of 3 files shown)


lib/embulk/command/embulk_run.rb
```diff
@@ -42,10 +42,16 @@ def self.run(argv)
...
       # only for help message. implemented at lib/embulk/command/embulk.rb
     end
+    op.on('-l', '--log-level LEVEL', 'Specify the global log level') do |level|
+      options[:logLevel] = level
+    end
 
     case subcmd.to_sym
     when :bundle
       op.remove  # remove --bundle
+      op.on('-l', '--log-level LEVEL', 'Specify the global log level') do |level|
```
frsyuki: we don't need this line

lib/embulk/command/embulk_run.rb
```diff
@@ -108,6 +114,9 @@ def self.run(argv)
 
     when :new
       op.remove  # remove --bundle
+      op.on('-l', '--log-level LEVEL', 'Specify the global log level') do |level|
```
frsyuki: we don't need this line

frsyuki: instead, `:new` and `:bundle` need `op.remove` call

embulk-core/src/main/java/org/embulk/exec/LoggerProvider.java
```diff
@@ -4,17 +4,21 @@
...
 {
-    public LoggerProvider()
+    @Inject
+    public LoggerProvider(@ForSystemConfig ConfigSource systemConfig)
     {
         // TODO system config
         Properties prop = new Properties();
 
-        prop.setProperty("log4j.rootLogger", "INFO,root");
+        String logLevel = systemConfig.get(String.class, "logLevel","INFO");
```
frsyuki: we need validation here to reject unknown level strings.

frsyuki:
tweaked and merged. thank you!

PR #84, opened 2015-02-18
add isPreview method

yaggytter:
such an implementation would you like?

e.g. use in ruby plugin
```
import 'org.embulk.spi.Exec'
puts Exec.isPreview
```
is this ok?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java
+++ b/embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java
@@ -52,6 +52,7 @@ public class PreviewExecutor
 
     public PreviewResult preview(ExecSession exec, final ConfigSource config)
     {
+        exec.setPreview(true);
         try {
             return Exec.doWith(exec, new ExecAction<PreviewResult>() {
                 public PreviewResult run()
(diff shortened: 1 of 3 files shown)


frsyuki:
I checked other code. In which case do you need Exec.isPreview? Do you have any examples?
PreviewExecutor should stop input plugin automatically when it gets enough number of records.

frsyuki:
I made Exec immutable and merged. Thank you!

yaggytter:
use case is
the final date of the acquired log is recorded,
get the greater than a last date at the next run
such as want to create an input plugins with the state

For example,
in embulk-input-slack-history

```
import 'org.embulk.spi.Exec'

# end of input plugin function
if !Exec.isPreview then
   update_lastlogdate
end
```

I guess
 embulk-plugin-input-sfdc-event-log-files etc...
necessary in use case where read the log if repeatedly executed.

sorry if this is unlike thought of embulk.
thank you again!

frsyuki:
@yaggytter I see. You need isPreview not to update the information at a data source in "preview" mode.
Can you use this code instead?:

```
# input plugin's transaction method
def self.transaction(config, &control)
  ...
  commit_reports = yield(...)

  # do this after yield()
  update_lastlogdate

  return {}
end
```

With this code, `update_lastlogdata` is called only when all `run` succeeded in "run" mode. It's not called in "preview" mode. Because `yield` throws an exception in preview mode.

If this doesn't work, please let me know. It might mean a bug of preview executor.

yaggytter:
thank you for teaching me!
let me try it!

PR #93, opened 2015-02-20
Fix gem name

yamatoya:
from [embulk-plugin-postgres-json] to [embulk-output-postgres-json]


frsyuki:
Nice pull-request! Thank you.

PR #94, opened 2015-02-20
Add @ char for use class variable

hiroyuki-sato:


hiroyuki-sato:
class variable -> instance variable.

frsyuki:
Thank you!

PR #103, opened 2015-02-24
Fix timestamp format '%s'

hiroyuki-sato:
Please review this patch. 

Detail
https://gist.github.com/hiroyuki-sato/61e5da934ccf41a258ca

Epoch time : 1424770450 is '2015-02-24 18:34:10 +0900'
But, preview result is show 47119-03-07 17:26:40 UTC.


frsyuki:
Right. Thank you

frsyuki:
I updated some other code to be able to handle microseconds (there was millisecond limitation before).

PR #107, opened 2015-02-26
set default task as :build to new Rakefile template

takei-yuya:
Default template README.md said 'run "rake" to build', but Rakefile have no default task.
And `rake` will raise an error like below.
```
rake aborted!
Don't know how to build task 'default'
```

// or, should I change README.md?


frsyuki:
Merged....!

PR #110, opened 2015-02-27
Embulk.run: fix timestamp format.

nishidayuya:
In many countries, this PR is invalid. :dizzy_face:
http://en.wikipedia.org/wiki/Decimal_mark#mediaviewer/File:DecimalSeparator.svg

### DIFF
--- a/lib/embulk/command/embulk_run.rb
+++ b/lib/embulk/command/embulk_run.rb
@@ -34,7 +34,7 @@ module Embulk
     op = OptionParser.new
     op.version = Embulk::VERSION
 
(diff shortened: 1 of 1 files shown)


frsyuki:
merged. thank you.

Reasons:
I made it same with log4j's default timestamp format but I changed the log4j's behavior to use dot. So, I needed to change Embulk.run as well but I've been missing it.
Dot is not standard in some countries but it is the standard in SQL timestamp format. I take SQL format because one of the major use cases of Embulk is definitely integrating with RDBMS. 
adea5ae21bfb5aa9a43060298d92c10fd745484c changed TimestampType::DEFAULT_FORMAT to use dot.

PR #111, opened 2015-02-27
Task 'classpath' should clean a destination directory first to avoid a version conflict

komamitsu:
After executing `./gradlew classpath` commands with several versions, I found the different versions of `embulk-core-0.x.x.jar` and `embulk-standards-0.x.x.jar` existed and `bin/embulk` command didn't work expectedly.

classpath dir

So I think `./gradlew classpath` command should clean the directory first.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -139,7 +139,7 @@ dependencies {
 //
 // classpath task
 //
-task classpath(dependsOn: ['build', ':embulk-cli:classpath']) << { }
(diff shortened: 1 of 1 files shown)


frsyuki:
Good catch. thank you. But running "clean" task is too much. It can delete only classpath directory.

frsyuki:
Fixed at ca0c61a1327621cc1ebc93deddd911d7de1b1af4

PR #112, opened 2015-02-28
Fixed FileInputInputStream.skip() to avoid returning -1

hata:
org.embulk.spi.util.FileInputInputStream.skip() returns -1 when there is
no data in the stream. skip() calls read() method and use the returned
value without validation. skip() should return 0 or more positive value
when read() returns -1.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/util/FileInputInputStream.java
+++ b/embulk-core/src/main/java/org/embulk/spi/util/FileInputInputStream.java
@@ -80,7 +80,8 @@ public class FileInputInputStream
     @Override
     public long skip(long len)
     {
-        return read(null, 0, (int) Math.min(len, Integer.MAX_VALUE));
+        int skipped = read(null, 0, (int) Math.min(len, Integer.MAX_VALUE));
+        return skipped > 0 ? skipped : 0;
     }
 
     private boolean nextBuffer()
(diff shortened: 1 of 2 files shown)


frsyuki:
I didn't know this spec...I confirmed that standard Socket InputStream never returns -1.
Merged. Thank you!

Confirmation code:

```java
import java.net.Socket;
import java.io.IOException;
import java.io.InputStream;
import java.io.OutputStream;

public class Isk
{
    public static void main(String[] args) throws Exception
    {
        Socket s = new Socket("127.0.0.1", 8080);  // echo server
        try (OutputStream out = s.getOutputStream()) {
            out.write("ab".getBytes());
            s.shutdownOutput();
            long n;
            try (InputStream in = s.getInputStream()) {
                n = in.skip(1);
                System.out.println("n: "+n);
                n = in.skip(1);
                System.out.println("n: "+n);
                n = in.skip(1);
                System.out.println("n: "+n);
                n = in.skip(1);
                System.out.println("n: "+n);
            }
        }
    }
}
```

Result:
```
$ java -cp . Isk
n: 1
n: 1
n: 0
n: 0
```

Server:
```
$ socat -v tcp-listen:8080,reuseaddr system:cat
```

hata:
Thank you very much for merging my request quickly.

I also checked some classes. java.io.BufferedInputStream also return 0 or positive value. In any case, I think it doesn't make sense -1 for "the actual number of bytes skipped." in javadoc.
reference:
http://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#skip(long)

Thanks again!

Issue #119, opened 2015-03-06
Add an option to CSV parser plugin to skip first some lines


frsyuki:
Some CSV files have some garbage lines.
Those lines are informative for some cases but they're not a part of table.

For example,

```
created date,2015-03-05
created by,me
device,xyz
time,key1,key2,key3,key4
2015-03-05 20:40:14 -0800,a,b,c,d
2015-03-05 20:40:14 -0800,a,b,c,d
2015-03-05 20:40:14 -0800,a,b,c,d
2015-03-05 20:40:14 -0800,a,b,c,d
2015-03-05 20:40:14 -0800,a,b,c,d
2015-03-05 20:40:14 -0800,a,b,c,d
...
```

frsyuki:
It would be great if csv guess plugin can guess them.

frsyuki:
An example:
https://gist.github.com/shun0102/bdf732df8470b3a73143

PR #126, opened 2015-03-07
Add skip_header_lines option to CSV parser plugin

frsyuki:
New `skip_header_lines` (int, optional) replaces `header_line` parameter. `header_line` parameter is obsoleted but kept for backward compatibility.


frsyuki:
For #119.

PR #128, opened 2015-03-10
Add delimiter config in CsvFormatter.

hiroyuki-sato:
It can use another delimiter instead of commna(',').

Reported-by hito4t <thitoshi@cac.co.jp>

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/CsvFormatterPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/CsvFormatterPlugin.java
@@ -31,6 +31,10 @@ public class CsvFormatterPlugin
         @Config("header_line")
         @ConfigDefault("true")
         public boolean getHeaderLine();
+
+        @Config("delimiter")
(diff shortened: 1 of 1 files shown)


embulk-standards/src/main/java/org/embulk/standards/CsvFormatterPlugin.java
```diff
@@ -124,7 +128,7 @@ public void timestampColumn(Column column)
                         private void addDelimiter(Column column)
                         {
                             if (column.getIndex() != 0) {
-                                encoder.addText(",");
+                                encoder.addText(task.getDelimiterChar());
```
frsyuki: Getter methods of Task classes are very slow. Can you change the code to cache the String to a local variable?

hiroyuki-sato:
Thank you for your comment.

I’m not familiar Java. How about this change?

And I missed to replace header field delimiter.

frsyuki:
Good! Merged.

PR #129, opened 2015-03-11
build.gradle: create and release embulk-xxx-tests.jar

muga:
Since test utilities in embulk-core project are useful for plugin development, we should create and release embulk-xxx-tests.jar files.


frsyuki:
Very good :+1:

PR #141, opened 2015-03-18
Using BufferedOutputStream

hito4t:
Using BufferedOutputStream will improve performance.


frsyuki:
How does this improve the performance? I'm assuming that using BufferedOutputStream causes double buffering because `#add(Buffer buffer)` method receives a buffered data. I suspect that FormatterPlugin or EncoderPlugin uses too small buffer size. Which plugins did you use?

hito4t:
Plugins I used were file input, csv parser, csv formatter and file output.

Default size of Buffer class is 256. 
It's too small, so using BufferedOutputStream (whose buffer size is 8192 bytes) improves performance.

But, as you wrote, using BufferedOutputStream causes double buffering, so it's not best solution.
Same as https://github.com/embulk/embulk/pull/142, passing large buffer size when calling BufferAllocator#allocate in FileOutputOutputStream class will improve performance more.
I'll create another pull request.

My test results are as follows:
  [current implementation]
    26.5s, 26.0s, 25.9s (avg. 26.1s)
  [BufferedOutputStream]
    16.5s, 15.6s, 16.2s (avg. 16.1s)
  [BuffereAllocator.allocate(8192)]
    11.2s, 11.3s, 11.5s (avg. 11.3s)
(input file size is 430MB)

frsyuki:
Great job.

PR #142, opened 2015-03-18
Increase buffer size of InputStreamFileInput

hito4t:
When FileInputInputStream#read(byte[], off, len) called by sun.nio.cs.StreamDecoder#readBytes, len is 8192.
But the capacity of FileInputInputStream#buffer is 256 (default), it's too small.
Setting 8192 to the capacity will improve performance.

In my environment, reading 1,000,000 lines file (430MB) takes 24s when the capacity is 256, and takes 9s when the capacity is 8192.
(We don't need to use BufferedInputStream (https://github.com/embulk/embulk/pull/140) if increasing buffer size)


frsyuki:
Awesome!!!

PR #143, opened 2015-03-18
Null check of plugin type

hito4t:
When I intended to use NullOutputPlugin and wrote "type: null" in yml, NullPointerException is thrown.
I should write "type: 'null'", and PluginManager class should check if type is null.


frsyuki:
Good improvement :-)

PR #147, opened 2015-03-21
Fix NullPointerException after PageBuilder#doFlush fails

hito4t:
PageBuilder#flush method calls doFlush, then calls newBuffer.
But if doFlush throws an exception, newBuffer won't be called and bufferSlice will be null.
That causes NullPointerException when setXXX is called.
I think newBuffer should be called even if doFlush throws an exception.

And should we call output.finish in PageBuilder#finish even if doFlush throws an exception?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
+++ b/embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
@@ -200,10 +200,13 @@ public class PageBuilder
 
     public void flush()
     {
-        doFlush();
-        if (buffer == null) {
-            newBuffer();
-        }
+    	try {
+            doFlush();
+    	} finally {
+            if (buffer == null) {
+                newBuffer();
+            }
+    	}
     }
 
     public void finish()
--- a/embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
+++ b/embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
@@ -200,13 +200,13 @@ public class PageBuilder
 
     public void flush()
     {
-    	try {
+        try {
             doFlush();
-    	} finally {
+        } finally {
             if (buffer == null) {
                 newBuffer();
             }
-    	}
+        }
     }
 
     public void finish()


frsyuki:
in which cache you call setXxx after doFlush throws exception?
You shouldn't call finish if you don't commit the transaction.

hito4t:
I use CsvParserPlugin.
When doFlush throws exception, CSVParserPlugin catches it (see below).
Then CSVParserPlugin parses next line, and call PageBuilder#setXXX, and NullPointerException thrown.

If doFlush in PageBuilder#flush throws exception, output.finish will be called.
But if doFlush in PageBuilder#finish throws exception, output.finish won't be called.

I think CSVParserPlugin should catch an exception when finishing, same as parsing.
<pre>
CSVParserPlugin#run

    ...
    try (final PageBuilder pageBuilder = new PageBuilder(Exec.getBufferAllocator(), schema, output)) {
        while (tokenizer.nextFile()) {
            ...
            while (true) {
                try {
                    ...
                            pageBuilder.setString(column, v);
                    ...
                    pageBuilder.addRecord();

                // exception from PageBuilder#doFlush will be caught
                } catch (Exception e) {
                    // TODO logging
                    long lineNumber = tokenizer.getCurrentLineNumber();
                    String skippedLine = tokenizer.skipCurrentLine();
                    log.warn(String.format("Skipped (line %d): %s", lineNumber, skippedLine), e);
                    //exec.notice().skippedLine(skippedLine);
                }
            }
        }

        //pageBuilder.finish();
        // ↓
        try {
            pageBuilder.flush();
        } catch (Exception e) {
            // will catch excpetion from PageBuilder#doFlush
            ...
        } finally {
            // PageBuilder#doFlush in PageBuilder#finish won't be called
            pageBuilder.finish();
        }
    }
</pre>

frsyuki:
It must be a bug in CsvParser. I created this fix: #149

frsyuki:
Please reopen if there is still the same problem.

hito4t:
Thanks!
It works well.

PR #148, opened 2015-03-22
Add allow_optional_columns option

kamatama41:
I want to parse the CSV with the optional columns, such as the following.
```
id,name,country,favorite(optional)
1,shinichi,JP,karaoke
2,scott,US
```

This pull-request add the option to allow CSV to have insufficient columns.(default is "false")
```
  parser:
    type: csv
    delimiter: ','
    allow_optional_columns: true
    columns:
    - {name: id,       type: long}
    - {name: name,     type: string}
    - {name: country,  type: string}
    - {name: favorite, type: string}
```


frsyuki:
I see...I merged.

PR #149, opened 2015-03-23
CsvParserPlugin should not ignore all exceptions #147

frsyuki:
CsvParserPlugin stupidly ignores all exceptions. But it should catch only CsvTokenizer exceptions.


frsyuki:
@muga I fixed this exception handling.

PR #152, opened 2015-03-29
Refactor ColumnVisitor instances in loop for CsvParserPlugin and CsvFormatterPlugin

hata:
# Description

This is a refactoring request relates to CsvParserPlugin and CsvFormatterPlugin relates to creating new instances in each loop.

I wonder if I should make this pull request or not. So, it is no problem for me to reject this request because this change can affect very little performance.

# Details

Commonly, it may be high cost to create instances in a loop. Current CsvFormatterPlugin and CsvParserPlugin creates ColumnVisitor for each record. It may be better to create the instance outside of the loop.

When using hprof profile for heap=site with my test and it shows 0.04% accumulate.(It depends on the number of records.) So, it affects very few memory allocation and it may be difficult to measure the performance improvement. (I guess the current majority of memory allocations are jruby code.)

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/CsvFormatterPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/CsvFormatterPlugin.java
@@ -45,27 +45,12 @@ public class CsvFormatterPlugin
         control.run(task.dump());
     }
 
-    private Map<Integer, TimestampFormatter> newTimestampFormatters(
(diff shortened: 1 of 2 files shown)


frsyuki:
I don't think this is necessary at least for now.

hata:
OK. No problem. It may be better to revisit when the instances becomes majority of memory allocation.

PR #176, opened 2015-05-14
Changing the default escape character to \" to be compatible with RFC 4180

frsyuki:
[RFC 4180](https://www.ietf.org/rfc/rfc4180.txt) says:

>   7.  If double-quotes are used to enclose fields, then a double-quote
>       appearing inside a field must be escaped by preceding it with
>       another double quote.  For example:
>
>       "aaa","b""bb","ccc"

So, I think default `escape` parameter should be `\"`.


frsyuki:
@sakama can I ask you to review this change?

sakama:
@frsyuki
 > So, I think default escape parameter should be \".

I think so too.
I used `\\` to be the same as the default option of CSV parser , but almost application or CSV writer use `\"` as default escape char.

frsyuki:
:+1: 
I merged.

PR #186, opened 2015-05-22
ProcessTask uses SchemaConfig as a workaround

frsyuki:
Workaround for https://github.com/embulk/embulk-executor-mapreduce/issues/1


muga:
:+1:

Issue #123, opened 2015-03-06
Add an option to CSV parser plugin to skip a line if it starts with "#"


frsyuki:
Some CSV files have `#` character at the beginning of a line as a comment.
It would be great if guess-csv can guess it.

repeatedly:
Maybe, adding option like skip_pattern is better.
CSV has no common view ;p

frsyuki:
Help wanted:
If you have an example CSV file which needs this option or similar option, please comment here.

randyh0329:
in our CSV format, we have `#CHECKSUM 972540854` as end of line.

I modified the CSV parser related function to support skip_pattern in configuration.
reference: https://goo.gl/jtcB8r

I know it's not the same idea as using guess-csv.

PR #189, opened 2015-05-27
Add skippattern support in Csv Parser Plugin

randyh0329:
Add Skip Pattern as an option. #123


frsyuki:
I think this code lacks some consideration such as unreadLines.

frsyuki:
I created #190 which adds code to `CsvTokenizer#nextLine` method.

randyh0329:
Thanks for the clean code. Really learned a lot from you. :)

Issue #120, opened 2015-03-06
Support array and map types


frsyuki:
Some storages can store nested values.
Many input data formatted in JSON or XM have nested valuesL
To transfer those storage, embulk needs to support nested types.

Discussions are:

* Array:
   * Fixed type array or variable type array
   * Fixed-type array: `array(T)` like `array(long)`. [1, 2, 3]
   * Variable-type array: `array`. [1, "str", 0.3]
* Map:
   * Fixed type or variable map
   * Allow non-string keys or not
* Struct:
   * Fixed-length map with fixed-types for each field.
   * For example, `struct(int, string))`

frsyuki:
#306 should be implemented first.

samuraraujo:
is this implemented? Does embulk support array types?

hiroyuki-sato:
Hello, @samuraraujo. Could you try `json` type?

PR #191, opened 2015-05-28
Added SwitchPageReader and SwitchPageBuilder utilities.

frsyuki:
The goal is to make type system extensible without breaking plugin API
compatibility.

* SwitchPageReader calls ValueSwitch#whenXxx methods where Xxx is name
  of a type. If a plugin does not know type Xxx, whenElse method is
  called. Plugins must implement whenElse and whenNull methods at least.
* SwitchPageBuilder calls ColumnSwitch#whenXxx methods where Xxx is name
  of a type. If a plugin does not implement those methods, calling those
  methods cause a exception. But as long as the methods are not called,
  plugin loading itself succeeds and plugin runs without exception.

The next step is to add array/map/json/decimal type support (#120).

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/type/BooleanType.java
+++ b/embulk-core/src/main/java/org/embulk/spi/type/BooleanType.java
@@ -1,5 +1,9 @@
 package org.embulk.spi.type;
 
+import org.embulk.spi.Column;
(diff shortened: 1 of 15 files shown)


frsyuki:
This is redesigned and taken over by #306.

PR #192, opened 2015-05-29
Implement selfrun.sh for Windows

hito4t:
The Windows version is not completely same as the Linux version.
If an argument contains '=', it should be enclosed by double quotations, because otherwise the left side and the right side are divided as two arguments in the bat program.

<pre>
embulk.bat "-J-Dembulk.max_threads=1" run mysql.yml
</pre>


frsyuki:
Amazing! This is a great achievement. 
Thank you!

PR #209, opened 2015-06-23
add DynamicPageBuilder utility

frsyuki:
This utility class makes it easy to develop a filter plugin.


frsyuki:
Ruby's Embulk::PageBuilder uses DynamicPageBuilder by default.

PR #216, opened 2015-06-27
Fix unavailable URL to access to 'How to acquire bintray API Keys'

monkey-mas:
Access to the current URL gives us 404 status code.


frsyuki:
Thank you!

PR #218, opened 2015-06-29
Fix trivial increment bug at CsvFormatterPlugin

huydx:
Hi @frsyuki, I just found a very simple bug which is increment of a counter in loop is missing at CsvFormatterPlugin, when create timeformatters array.
The bug cause for every input which have timestamp field with index bigger then 0 to be Null Exception.


frsyuki:
ouch! Thank you!

PR #225, opened 2015-07-01
PluginClassLoader loads classes from child classloader first

frsyuki:
To prevent problems, list of packages which should be loaded from parent
class loader first is generated by a script.
This script (./gradlew updateResources) reads all dependent jars of
embulk-cli and creates list of common package names.


muga:
:+1:

PR #233, opened 2015-07-08
Expose @java_file_input for parser plugins

uu59:
`@java_file_input` is needed by ruby parser plugins.

e.g.:
https://github.com/treasure-data/embulk-parser-query_string/pull/6/files#diff-1e7234223ad807f67a382ee2242b79d5R40
https://github.com/shun0102/embulk-parser-jsonl/blob/master/lib/embulk/parser/jsonl.rb#L26

So I added `attr_reader` for it.

### DIFF
--- a/lib/embulk/file_input.rb
+++ b/lib/embulk/file_input.rb
@@ -3,6 +3,8 @@ module Embulk
   require 'embulk/buffer'
 
   class FileInput
+    attr_reader :java_file_input
+
     def initialize(java_file_input)
       @java_file_input = java_file_input
       @buffer = nil


frsyuki:
I think method name is `to_java`

yoshihara:
@frsyuki (Cc: @uu59 ) Do you mean `#to_java` instead of attr_reader returns `@java_file_input` value? (I think `to_java` name seems to be good because other values has same name method for same purpose)

frsyuki:
Added at https://github.com/embulk/embulk/commit/6848239f4a354dfdf5686ed6c13951ad333c5215

uu59:
Thanks!

yoshihara:
Thanks!!

PR #248, opened 2015-08-05
CSV parser plugin supports non-quoted values which contains "

frsyuki:
This change includes 3 changes:

* CSV parser supports null to `quote` and `escape` options. It means
  disabling quoting or escaping.
* CSV guess plugin uses `quote: null` if some values contain '"' in
  middle of non-quoted string.
* <del>CSV parser assumes `quote: ''` (or `escape: ''`) mean disabling
  quoting (or escaping). This is backward incompatible because previous
  behavior was assuming quote='"' (or escape="\\")</del>

### DIFF
--- a/embulk-docs/src/built-in.rst
+++ b/embulk-docs/src/built-in.rst
@@ -23,7 +23,7 @@ Embulk uses a YAML file to define a bulk data loading. Here is an example of the
         type: csv
         delimiter: ','
         quote: '"'
-        escape: ''
+        escape: '"'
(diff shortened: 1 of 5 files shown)


frsyuki:
I added another commit to make the behavior backward compatible.

muga:
:+1:

frsyuki:
squashed & merged:
https://github.com/embulk/embulk/commit/a21e6134786883001bf0091588826ab1d998957b

PR #255, opened 2015-08-09
Add EmbulkEmbed with JSR 250 Lifecycle support

frsyuki:
EmbulkEmbed is a new interface to embed Embulk in an application. This
class will replace EmbulkService once the API is designed.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
+++ b/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
@@ -1,0 +1,216 @@
+package org.embulk;
+
+import java.util.Arrays;
+import java.util.List;
+import com.google.common.base.Function;
+import com.google.common.base.Throwables;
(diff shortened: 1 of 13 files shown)


cosmo0920:
Good stuff. :+1: 

For now, this test helper code https://github.com/treasure-data/embulk-parser-query_string/blob/cb2e1abb0c3daedc37b5fcf06bb4b78acf5f528e/test/prepare_embulk.rb becomes simpler than original as follows:

```ruby
require "embulk/command/embulk_run"

# stable API
Embulk.require_classpath

props = java.util.Properties.new
props.setProperty("embulk.use_global_ruby_runtime", "true")

# Using beta EmbulkEmbed API
config_loader = org.embulk.EmbulkEmbed.newSystemConfigLoader
system_config = config_loader.fromPropertiesYamlLiteral(props, "embulk.")
org.embulk.EmbulkEmbed.new(system_config)

require "embulk"
```

PR #257, opened 2015-08-10
DataSource should raise ConfigException on config error

frsyuki:


frsyuki:
This change rescues ArgumentError. It is wrong behavior.
This should be fixed with #258.

PR #260, opened 2015-08-10
Try to clean up temp files recursively in TempFileSpace.java

cosmo0920:
For JDK 1.7 or later, it can use useful pattern which is combining `java.nio.file.Files` and `java.nio.file.SimpleFileVisitor` to do file operation.


frsyuki:
Great! Thank you!

cosmo0920:
Thanks!!

frsyuki:
I added this follow-up commit: https://github.com/embulk/embulk/commit/e94308854eaa4c09d2a0bb1dd50cc7205c490d31
This is the code style embulk is/will be using: https://github.com/airlift/codestyle

Issue #228, opened 2015-07-07
Schema.builder API for Ruby


frsyuki:
I added spi.Schema.Builder to Java plugin API.
I want similar API for Ruby so that we can avoid this kind of code: https://github.com/treasure-data/embulk-parser-query_string/blob/7aec718f5f056301d23ac054c1c3cf3e8b7d7469/lib/embulk/parser/query_string.rb#L15-L22

dmikurube:
We have no plans for expanding Ruby-based Embulk plugins. Closing this issue.

PR #262, opened 2015-08-12
Import Schema#builder API from Java to Ruby style

cosmo0920:
Related to #228.

Following code 

```ruby
 columns = []
 schema = config.param(:schema, :array, default: [])
 schema.each do |column|
   name = column["name"]
   type = column["type"].to_sym
   
   columns << Column.new(nil, name, type)
 end
```

will become:

```ruby
column_config = config.param(:schema, :array, default: [])
columns = Builder.build(column_config)
```

Or this API can more simplize?

### DIFF
--- a/lib/embulk/builder.rb
+++ b/lib/embulk/builder.rb
@@ -1,0 +1,21 @@
+module Embulk
+
+  require 'embulk/column'
+
+  class Builder
+    def self.add(name, type_str_or_sym)
+      Column.new(nil, name, type_str_or_sym.to_sym)
+    end
+
+    def self.build(column_config)
+      columns = []
+      column_config.each do |column|
+        name = column["name"]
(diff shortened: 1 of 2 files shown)


cosmo0920:
Hmm, it seems to be something wrong.
Please wait a moment.

lib/embulk/schema.rb
```diff
@@ -88,6 +88,15 @@ def write_record(page_builder, record)
       @record_writer.call(page_builder, record)
     end
 
+    def builder!(columns)
+      self.each do |column|
```
cosmo0920: Umm..., I'm no confidence this code.... :sweat:

cosmo0920:
I introduced `Embulk::Builder` class to build and add Schema.
Does it deviates from the original purpose?

PR #278, opened 2015-08-18
Add DataException and UserDataException tag interface

frsyuki:
An exception class DataException is expected to be thrown if input data
has a problem (e.g. unexpected timestamp format).

An interface UserDataException is implemented by exception classes which
represent a problem of input data by users. For example,
ConfigException, DataException and TimestampParseException. If those
exceptions are thrown, applications are suggested not to retry this bulk
load because retried bulk import will fail again unless user's input
data is fixed.


muga:
:+1:

Issue #27, opened 2015-01-29
Writing broken records or texts to a separated destination


frsyuki:
Data include a lot of broken records. A bulk import can skip them but we want to load them later as an exceptional case. To do it, we want to get those error records written to other files or databases.

Difficulty in terms of API design is that format of the records can be different depending on plugin types.

* Encoder, Decoder, some Parser plugins
** These plugins read with buffer. They can't recognize "records". When they detect broken data, they *skip the entire file*
* Line-based parser plugins
** Some parser plugins are based on lines (e.g. csv). They can *skip a line* and continue parsing from the next line.
* Formatter and Output plugins
** Formatter plugins read records. They can *skip a record* whose schema is fixed by the previous plugins.
* Filter plugins (#26)
** Filter plugins read records. They can *skip a record* whose schema is fixed by the previous plugins.

So, depending on plugins, error output needs to store 3 kinds of data:

a) file from a certain position
b) line
c) record with various schema

frsyuki:
A possible idea:

```java
XxxPlugin extends interface Plugin;
ErrorReporter Exec.getErrorReporter(Plugin instance);
void ErrorReporter.reportError(XxxException cause, Record record);

static Record Record.copyOf(Schema, PageReader);
static Record Record.ofText(String text);
static RecordBuilder Record.builder();
Schema Record.getSchema();
RecordReader Record.getRecordReader()?

interface ErrorPlugin extends Plugin;
TaskSource ErrorPlugin.configure(ConfigSource config);
ErrorReporter ErrorPlugin.open(TaskSource taskSource, PluginType pluginType);
void ErrorReporter.report(Exception cause, Record record);

static PluginType PluginType.of(Plugin instance);
```

An example `CsvErrorPlugin` writes error records to `{exec.getTransactionUniqueName()}.{pluginType}.{schema.hashCode()}.{cause.getClass().getSimpleName()}.{seqId}.csv` file.

Another example `JsonErrorPlugin` writes error in `{"cause":"{cause.getClass().getSimpleName()}", "message":"{cause.getMessage()}", "schema":{schema}, "record": {record}}` format.

frsyuki:
I found that this idea by itself doesn't work because it doesn't consider distributed execution environment.
Each `ErrorReporter` (each `ErrorPlugin.open` calls) needs an unique identifier. The identifier is a kind of task ID. Such ID is not available in ExecSession.

frsyuki:
Here is another idea.

[...]

PR #279, opened 2015-08-20
Add stop_on_invalid_record option to CsvParserPlugin

frsyuki:
In some cases, stopping bulk load is more useful than silently skipping.
This option will be necessary until #27 is available.


frsyuki:
Merged.

Issue #280, opened 2015-08-20
selfupdate subcommand is corrupted?


cosmo0920:
I freshly installed latest Embulk(0.7.1) into Ubuntu machine with Embulk README instruction: https://github.com/embulk/embulk#linux--mac--bsd

And then I execute `embulk selfupdate` command:

### Expected

Embulk selfupdate subcommand success.

### Actual

```bash
% embulk selfupdate
2015-08-20 16:25:55.645 +0900: Embulk v0.7.1
ArgumentError: wrong number of arguments (0 for 1)
      run at /home/hhatake/.embulk/bin/embulk!/embulk/command/embulk_run.rb:256
    <top> at /home/hhatake/.embulk/bin/embulk!/embulk/command/embulk_main.rb:2
  require at org/jruby/RubyKernel.java:940
   (root) at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1
    <top> at file:/home/hhatake/.embulk/bin/embulk!/embulk/command/embulk_bundle.rb:55
```

frsyuki:
strange...

cosmo0920:
I tried other version of Embulk.
I got different result 0.6.25 and 0.7.0.

### Embulk 0.6.25:

I got expected result.

```bash
% embulk selfupdate
2015-08-20 16:39:34.461 +0900: Embulk v0.6.25
Checking the latest version...
Found new version 0.7.1.
Downloading https://dl.bintray.com/embulk/maven/embulk-0.7.1.jar ...
Updated to 0.7.1.
```

### Embulk 0.7.0

I got same error as above:

```bash
% embulk selfupdate
2015-08-20 16:40:56.142 +0900: Embulk v0.7.0
ArgumentError: wrong number of arguments (0 for 1)
      run at /home/hhatake/.embulk/bin/embulk!/embulk/command/embulk_run.rb:256
    <top> at /home/hhatake/.embulk/bin/embulk!/embulk/command/embulk_main.rb:2
  require at org/jruby/RubyKernel.java:940
   (root) at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1
    <top> at file:/home/hhatake/.embulk/bin/embulk!/embulk/command/embulk_bundle.rb:55

```

frsyuki:
[...]

PR #282, opened 2015-08-20
Fix selfupdate subcommand

cosmo0920:
To avoid curious result, I use `options` in Embulk.selfupdate argument instead of `system`.

Related to #280.


frsyuki:
Thank you :+1:

PR #288, opened 2015-08-27
Add out_columns parameter definition (ruby-filter template)

hiroyuki-sato:
ruby-filter template does not define ``out_columns``

alternative idea

```ruby
yield(task, in_schema)
```


frsyuki:
:+1:

Issue #286, opened 2015-08-25
version 0.7.3 on windows8.1 fail "./gradlew" after "embulk migrate".


mikoto2000:
Embulk version 0.7.3 on windows8.1 fail "./gradlew" after "embulk migrate".

Is there something that is considered to be the cause of that?

Best regards.

## Detailes

I tried two environments.

- Debian8, OpenJDK-7, embulk-0.7.3 -> OK.
- Windows8.1 on msys2 bash, Oracle Java 8 JDK, embulk-0.7.3 -> NG.

## Debian8, openjdk-7, embulk-0.7.3

```sh
$ wget https://dl.bintray.com/embulk/maven/embulk-0.7.3.ja
$ chmod u+x embulk-0.7.3.jar
$ git clone https://github.com/mikoto2000/embulk-input-salesforce_bulk
$ ./embulk-0.7.3.jar migrate embulk-input-salesforce_bulk/
2015-08-26 00:32:50.031 +0900: Embulk v0.7.3
Detected Java plugin for Embulk 0.6.21...
  Modified src/main/java/org/embulk/input/salesforce_bulk/SalesforceBulkInputPlugin.java
  Modified gradle/wrapper/gradle-wrapper.jar
  Modified gradle/wrapper/gradle-wrapper.properties
  Modified build.gradle
Done. Please check modifieid files.$ cd embulk-input-salesforce_bulk/
$ chmod u+x gradlew
...(start gradlew success)

$ ls -l embulk-input-salesforce_bulk/gradle/wrapper/gradle-wrapper.jar
-rw-r--r-- 1 root root 51018 Aug 25 23:40 embulk-input-salesforce_bulk/gradle/wrapper/gradle-wrapper.jar
```

## Windows8.1 on msys2 bash, OracleJDK8, embulk-0.7.3

```sh
mikoto@homewin MINGW64 /c/Users/mikoto/project/tmp
$ curl --create-dirs -o ./bin/embulk -L "https://dl.bintray.com/embulk/maven/embulk-0.7.3.jar"
$ git clone https://github.com/mikoto2000/embulk-input-salesforce_bulk
$ ./bin/embulk migrate embulk-input-salesforce_bulk/
2015-08-25 22:55:31.184 +0900: Embulk v0.7.3
Detected Java plugin for Embulk 0.6.21...
  Modified src/main/java/org/embulk/input/salesforce_bulk/SalesforceBulkInputPlugin.java
[...]

PR #289, opened 2015-08-27
Fix fail migrate on windows.

mikoto2000:
Use File#binread, when copy 'gradle-wrapper.jar'.
see #286 issue.


frsyuki:
Nice. Thank you!

Issue #124, opened 2015-03-06
Splitting an input file into multiple tasks by lines (or size)


frsyuki:
File input plugins create 1 task for each file. But a file can be very large. If it loads a big one file, emublk can't take advantage of multiple CPU cores.

But splitting a file into multiple files depends on the file format. For example, CSV file can have multi-line value. Splitting a file by lines can cause invalid output if a multi-line value is at a boundary.
However, at some cases, people want parallel loading even with the limitation.

General implementation is hard because it's assumed that number of input tasks is same with number of output tasks. If FileInputRunner and executor is flexible enough, we can implement the file splitter at the decoder layer.
Otherwise, each file input plugins need to support an option to split a file by size.

Issue #46, opened 2015-01-31
Make the transaction fail if the error ratio is too large


muga:
I want to set the error ratio for rolling back the transaction.

frsyuki:
Difficulties are:

* What's error ratio? (how does embulk calculates "number of error records"?)
* What's the default threshold ratio?

PR #293, opened 2015-08-30
Add experimental Mixin mechanism (WIP)

frsyuki:
Mixin is a meta-plugin that wraps another plugin. A mixin enhances
functionality of a plugin, and/or provides API to a plugin.

Plugins with fields annotated by @PluginMixin can use mixin:

```
@PluginMixin
ParserErrorMixin error;
```

or

```
@PluginMixin
void setError(ParserErrorMixin error)
```

This field is filled by PluginManager. Plugins can call methods of the
mixin instance at transaction and run/open methods.

A mixin class (such as LineParserErrorMixin) implements `T mixin(T)`
method where T is an instance of the plugin. Typically, it returns a
wrapped plugin instance.

Mixin classes can use TaskReport mechanism. At transaction stage, mixin
class calls `Exec.newMixinId()` to get a MixinId. At task stage, mixin
sets TaskReport using `Exec.reportMixinTask(MixinId, TaskReport` method.
Then at transaction stage again, mixin gathers TaskReport of tasks using
`Exec.getMixinReports(MixinId)` method.

### DIFF
--- a/embulk-core/build.gradle
+++ b/embulk-core/build.gradle
@@ -26,6 +26,7 @@ dependencies {
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-guava:2.5.3'
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-joda:2.5.3'
     compile 'com.fasterxml.jackson.module:jackson-module-guice:2.5.3'
+    compile 'org.msgpack:msgpack-core:0.7.0-p9'
     compile 'ch.qos.logback:logback-classic:1.1.3'
     compile 'org.slf4j:slf4j-api:1.7.12'
     compile 'org.jruby:jruby-complete:' + project.jrubyVersion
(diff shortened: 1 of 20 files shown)


frsyuki:
Mixi mechanism will be infrastructure to solve error handlers (#27 and #46).
This also could provide generic solution for #124.

frsyuki:
I rethink that this mechanism is too large to support error handling. And having multiple error plugin instances makes it impossible to calculate `max_error_records` globally because calculation of number of error records are done by each error plugin, while error records from both input plugin and output plugin are error records.

Next idea is to initialize one additional output plugin for error handling. A config file will have `in:`, `out:`, and `err:`. `err:` accepts any kinds of output plugins. BulkLoader.doRun takes `err:` config and calls transaction. Then Executors.process (or an additional method) opens it. Then set the opened PageOutput to thread-local storage (maybe wrapper of ExecSession). New `Exec.addRecord(Value)` method adds a record to the PageOutput. Executors.process closes the PageOutput, and passes TaskReport to Executors.ProcessStateCallback.errorCommitted right before outputCommitted call. Implementation of errorCommitted method sets errorTaskReport to ProcessState.

Issue #266, opened 2015-08-15
GuessExecutor never use 'guess_plugins' and 'exclude_guess_plugins' options in exec config


muga:
`GuessExecutor` cannot get guess_plugins from `GuessExecutorTask`. The following declaration is ignored by `GuessExecutor`.
```yaml
exec:
  guess_plugins:
  - query_string
in:
  ... ...
```
 `GuessExecutor#guessParserConfig()` executes `GuessExecutorTask#getGuessPlugins` () but the result doesn't have `guess_plugins` or `exclude_guess_plugins` options. Because `FileInputRunner` uses `Exec.session().getSessionExecConfig()` to generate `execConfig`.

frsyuki:
`guess_plugins` option has never been working, apparently.

Several ideas to fix:

* Change ExecSession to have entire `exec:` config. I think this gives too much access to `exec:` config for plugins. It is risk of abuse of `exec:` config. I don't want to make `exec:` config messy.
* Instantiate GuessExecutor for each session (maybe rename to GuessRunner) with `exec:` config and give it to plugins by adding `ExecRunner ExecSession.getGuessRunner()`.

I think the latter idea is better. Related to #242.

muga:
I agree with your 2nd idea. I thought the first idea to fix this problem. But it doesn't make sense that plugins can know `exec:` config and control the behavior of `Executor`.

frsyuki:
Fixed by #294.

PR #294, opened 2015-08-31
Added exec.guess_plugins option for guess executor

frsyuki:
New `exec: {guess_plugins: [array]}` option allows users to configure
list of guess plugins within a configuration file.

This is implemented as following: if an input plugin implements
ConfigurableGuessInputPlugin interface, GuessExecutor gives exec config
to guess method. FileInputRunner implements it and takes
guess.guess_plugins parameter.

Fixes [#266].


muga:
:+1:

PR #298, opened 2015-09-01
Added '--logback-config' option.

mikoto2000:
Supports --log PATH option in 'v0.7.4'(#291).
But I want more various logger setting.

## Proposal

1. define '--logback-config LOGBACK_CONFIG_PATH' option.
2. modify read logger setting process.
    - read default setting file after read LOGBACK_CONFIG_PATH file.

#### 1. create logback.xml

```xml
<!-- Output to './plugin.error.log', only XxxPlugin's Error message. -->
<configuration>
<appender name="ERROR" class="ch.qos.logback.core.FileAppender">
    <file>./app.error.log</file>
    <encoder>
        <charset>UTF-8</charset>
        <pattern>%d{yyyy-MM-dd'T'HH:mm:ss'Z'} - %m%n</pattern>
    </encoder>
</appender>
<logger name="org.embulk.input.xxx.XxxPlugin">
    <level value="ERROR" />
    <appender-ref ref="ERROR" />
</logger>
</configuration>
```

#### 2. set path to '--logback-config' option

```sh
./embulk run --logback-config /PATH/TO/logback.xml ./config.yml
```

Is there any problem?
or have more effective idea?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/LoggerProvider.java
+++ b/embulk-core/src/main/java/org/embulk/exec/LoggerProvider.java
@@ -55,6 +55,18 @@ public class LoggerProvider
             throw new RuntimeException(ex);
         }
 
+        String logbackConfig = systemConfig.get(String.class, "logback_config", "-");
+
+        if (!logbackConfig.equals("-")) {
+            System.setProperty("embulk.logbackConfig", logbackConfig);
+
+            try {
+                configurator.doConfigure(logbackConfig);
+            } catch (JoranException ex) {
+                throw new RuntimeException(ex);
+            }
+        }
+
         org.slf4j.Logger logger = LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
         if (logger instanceof Logger) {
             ((Logger) logger).setLevel(Level.toLevel(level.toUpperCase(), Level.DEBUG));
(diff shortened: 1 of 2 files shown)


frsyuki:
> But I want more various logger setting.

Why? In which use case default logger setting is not good enough?

mikoto2000:
I want to file output only when an error occurs.
Logback contains Embulk. I want to use it.

specifically...

1. For skip/fail record report.
2. Notify error to user.
    - Notification mechanism is already deliver from infrastructure.
    - Notification mechanism is 'error log file monitoring'.
        - Notify when modify error log file.

hmm...
Error log extracting is work of monitoring system?
Popular monitoring system is supported extract function.
-> Benefits than I thought small ...

frsyuki:
Exposing internal logging library (logback) to users seems not good idea because we need to keep backward compatibility of logback as a feature of embulk.
As you mention, for "2. Notify error to users", you can configure monitoring system to do it (fluentd + in_tail + fluent-plugin-grep + fluent-output-mail, for example). 
For "1. skip/fail record report", please try to evaluate #293. Your feedback is welcomed.

mikoto2000:
Thank you for comment.

> we need to keep backward compatibility of logback as a feature of embulk.

Okey. I think it very sound reason.
I try #293 and close this PR.

PR #299, opened 2015-09-03
Add basic multiline basic LineDecoder tests

cosmo0920:
This PR adds basic multibyte tests to LineDecoder.
It world be better more character type in test data?

### DIFF
--- a/embulk-core/src/test/java/org/embulk/spi/util/TestLineDecoder.java
+++ b/embulk-core/src/test/java/org/embulk/spi/util/TestLineDecoder.java
@@ -162,4 +162,40 @@ public class TestLineDecoder
                 bufferList(StandardCharsets.UTF_8, "て", "1", "\r\n", "す", "2", "\r", "\n", "と3"));
         assertEquals(ImmutableList.of("て1", "す2", "と3"), decoded);
     }
+
+    @Test
+    public void testDecodeBasicMS932() throws Exception
+    {
+        List<String> decoded = doDecode(
+                Charset.forName("ms932"), Newline.LF,
+                bufferList(Charset.forName("ms932"), "てすと1\nテスト2\nてすと3\n"));
+        assertEquals(ImmutableList.of("てすと1", "テスト2", "てすと3"), decoded);
+    }
+
+    @Test
+    public void testDecodeBasicMS932Tail() throws Exception
+    {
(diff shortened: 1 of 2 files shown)


embulk-core/src/test/java/org/embulk/spi/util/TestLineDecoder.java
```diff
@@ -128,5 +128,39 @@ public void testDecodeChunksCRLF() throws Exception
         assertEquals(ImmutableList.of("t1", "t2", "t3"), decoded);
     }
 
-    // TODO test multibytes
+    @Test
+    public void testDecodeBasicMultiByte() throws Exception
+    {
+        List<String> decoded = doDecode(
+                Charset.forName("utf-8"), Newline.LF,
```
frsyuki: Please use StandardCharsets.UTF_8 instead of Charset.forName

frsyuki:
UTF-16 and MS932 are very good to have.

embulk-core/src/test/java/org/embulk/spi/util/TestLineDecoder.java
cosmo0920: OK, I'll try it!

cosmo0920:
I added following test cases:

* UTF-8 test cases which contains multibyte
* MS932 test cases which contains multibyte (for Japanese edition of Windows default encoding)
* UTF-16LE test cases which contains multibyte (this test cases for PowerShell default output text encoding)

fixes:

* Add functionality to test util function to specify charset and newline style
* Use `StandardCharsets` constants if possible

frsyuki:
Good :+1:
I think this is good enough. But do you think you can reduce copy of code by using loop?

cosmo0920:
I think that there is a possibility to reduce code with loop.

But this case is more elegant to use [JUnitParams](https://github.com/Pragmatists/JUnitParams) to make parametrized test.

I'll more investigate to reduce redundant test code and try to use `JUnitParams` test library.

frsyuki:
That's interesting idea.
Or another idea is to migrate to use TestNG.
But it will be relatively large change. Let's do that using another issue thread.

cosmo0920:
:+1: 
It sounds great.
Let's investigate above idea in another issue thread!

cosmo0920:
Thanks!

PR #302, opened 2015-09-08
Purge setCompatVersion code

cosmo0920:
Because this deprecated function has no effect in JRuby 9.0.0.0 or later.

see:
https://github.com/jruby/jruby/blob/e10ec96f39d236e947d969877d138123a8af5344/core/src/main/java/org/jruby/embed/ScriptingContainer.java#L508


frsyuki:
:+1:

cosmo0920:
Thanks!

PR #303, opened 2015-09-08
Fix incorrect line number in warning messages.

yyamano:
Fix incorrect line number in warning messages.
9e70f55403e9ce3e8836cdafe8db66a43af265e1 is not the right change.
It doesn't work if files have headers. My fault.


frsyuki:
@yyamano  thanks! I created this fix instead. Does this change work?: #305

yyamano:
It works well, and better than mine:-) Thank you!

frsyuki:
:+1:

PR #311, opened 2015-09-29
Add encoder and decoder categories

hiroyuki-sato:
http://www.embulk.org/docs/

No encoder and decoder categories.


frsyuki:
:+1:

hiroyuki-sato:
Thanks!

PR #323, opened 2015-10-07
org.embulk.spi.Column#visit(): disuse "instanceof"

hishidama:


frsyuki:
I feel that having public getTypeEnum method prevents us from adding complex types (such as array and map). I don't find benefits in contrast to its potential disadvantage.

hishidama:
I think  enum-switch is faster than if-instanceof in ColumnVisitor.

dmikurube:
Sorry that this pull request has been left. Let us close this pull request.

PR #328, opened 2015-10-26
csv formatter should ignore quote character if quote_policy is NONE

### LINKED ISSUE #327: formatter-csv quotes my JSON field even quote_policy: NONE
sample.tsv

```
a[TAB]b[TAB]c[TAB]{"a":1,"b":2}
```

sample.yml

```
in:
  type: file
  path_prefix: sample.tsv
  parser:
    type: csv
    delimiter: "\t"
    quote: null
    escape: null
    columns:
      - {name: column1, type: string}
      - {name: column2, type: string}
      - {name: column3, type: string}
      - {name: column4, type: string}
out:
  type: file
  path_prefix: '/tmp/sample.'
  file_ext: 'csv'
  overwrite: true
  formatter:
    type: csv
    delimiter: "\t"
    quote_policy: NONE
    header_line: false
```

/tmp/sample.000.00.csv becomes as

```
a[TAB]b[TAB]c[TAB]{\"a\":1,\"b\":2}
```

frsyuki:
In addition to it, escape character should be the same character with
quote character by default to be consistent with csv parser plugin.

Fixes #327.


frsyuki:
Merged.

PR #329, opened 2015-10-27
[Propose] Customize StdoutOutputPlugin to use formatter

sonots:
What I expected for StdoutOutputPlugin was to output the formatted results.
So, I customized StdoutOutputPlugin to use formatter. 

If we want to see the internal states of embulk records, we can now use [embulk-filter-stdout](https://github.com/sonots/embulk-filter-stdout).

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/StdoutOutputPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/StdoutOutputPlugin.java
@@ -10,13 +10,14 @@ import org.embulk.spi.time.TimestampFormatter;
 import org.embulk.spi.Schema;
 import org.embulk.spi.Page;
 import org.embulk.spi.Exec;
-import org.embulk.spi.OutputPlugin;
-import org.embulk.spi.TransactionalPageOutput;
-import org.embulk.spi.PageReader;
-import org.embulk.spi.util.PagePrinter;
+import org.embulk.spi.Buffer;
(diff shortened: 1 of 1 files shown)


sonots:
I wanted to set the default formatter as

```
filter:
  type: csv
  quote_policy: NONE
```

as the output will be similar with the original behavior, but I could not find a good simple way to do it. Any idea?

frsyuki:
I used OutputPlugin because it's easy to use.

```
out:
  type: stdout
```

vs

```
out:
  type: stdout
  formatter:
    type: csv
```

I think stdout should be kept as-is to provide above simplicity.
How about creating another plugin like `stdout_file`?

sonots:
I want to make the default behavior to be `formmatter: csv` so that we can write as:

```
out:
  type: stdout
```

If this is possible, I think `stdout_file` is not necessary, and we can avoid the wierd name plugin > `stdout_file`.

frsyuki:
That makes sense. And I agree that. But we can't do that because of plugin API design.
FileOutputPlugin is wrapped in FileOutputRunner. So, FileOutputPlugin is not a OutputPlugin. This was necessary to make Java plugin API easy to use. But because of this, FileOutputPlugin can't modify or generate configuration for FormatterPlugin.

sonots:
Hmm

dmikurube:
Sorry that this pull request has been left. Let us close this pull request.

PR #332, opened 2015-10-29
Prevent exceptions at close() from overwriting essential exception

frsyuki:
If an exception happens both in `try` block and `finally` block,
exception by `finally` block overwrites exception by `try` block. Thus
the ssential exception information will be lost if `close()` call at
`finally` block throws an exception.

But with try-with-resource, exception happend at `close()` will be added
to suppressed exception of the exception at `try` block.


muga:
Looks good.

PR #333, opened 2015-10-29
add LifeCycleListener to log construction and destruction process of guice modules

frsyuki:
This is a step to separate org.embulk.guice module to a separated OSS module.


frsyuki:
this is done by #367.

PR #335, opened 2015-11-05
Support "mkbundle --path" option

frsyuki:
When we deploy embulk to a production environment using chef or
capistrano, gems should be installed to a shared directory
(shared/bundler) but Gemfile.lock should be isolated. --path optoin
works for that case.

This change includes bundler itself to embulk-core.jar so that
embulk_run doesn't have to install bundler. Thus initialization of
bundler becomes less magical.

Following-up of #334.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    embulk (0.7.4)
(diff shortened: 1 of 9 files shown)


sonots:
Thanks! Worked :D

PR #339, opened 2015-11-14
Change internal buffer size of GzipFileDecoderPlugin for a guess executor

muga:
When a guess executor tries to read a gzipped file, it sometimes cannot read an enough sized sample buffer to guess effectively. In my case, it read about 800 bytes only from the file as a sample buffer and cannot suggest column names and types by the sample buffer.

I changed the internal buffer size in GZIPInputStream to 8*1024 bytes. I think that 4096 bytes are also fine. But default 512 bytes are too small for my case.

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/GzipFileDecoderPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/GzipFileDecoderPlugin.java
@@ -43,7 +43,7 @@ public class GzipFileDecoderPlugin
                         if (!files.nextFile()) {
                             return null;
                         }
-                        return new GZIPInputStream(files);
+                        return new GZIPInputStream(files, 8*1024);
                     }
 
                     public void close() throws IOException


frsyuki:
why do you increase buffer size to increase sample data size of guess...?
Don't you need to increase `org.embulk.exec.SamplingParserPlugin.sampleSize` instead?

frsyuki:
increasing buffer size itself looks good.

muga:
In `SamplingParserPlugin.readSample` method, `fileInput` always has one sample `Buffer` element by `GuessExecutor.getFirstBuffer`. If the size of the sample `Buffer` is less than `sampleSize`, `GuessExecutor` needs to guess columns by the sample `Buffer` size only.

frsyuki:
@muga then problem is in GuessExecutor.getFirstBuffer. It should read more data. Buffer size should be performance parameter which doesn't affect any behavior excepting performance.

muga:
@frsyuki You're right. I reviewed #340. The fix looks good to me. thank you!

frsyuki:
this change itself is still necessary. I reopened and merged.

PR #340, opened 2015-11-15
GuessExecutor should read more than one buffer

frsyuki:
GuessExecutor.GuessParserPlugin.run reads sample data and using it
repeatedly. When it reads data from a decoder plugin, it needs to more
than one buffer because decoder plugin splits a buffer passed by
SamplingParserPlugin into multiple buffers.

Fixes #339.


muga:
:+1:

PR #352, opened 2015-12-22
Add CSV guess test

uu59:
Added CSV guess tests except `escape` option.
The reason is as [pending comment](https://github.com/embulk/embulk/pull/352/files#diff-3b3b0c90c6f748ef34ae306fd6c6f441R42), but I'm not sure that is intentional or bug...

### DIFF
--- a/test/guess/test_csv_guess.rb
+++ b/test/guess/test_csv_guess.rb
@@ -140,6 +140,24 @@ class CsvGuessTest < ::Test::Unit::TestCase
       ]
       assert_equal expected, actual["parser"]["columns"]
     end
+
+    def test_complex_line
+      actual = guess([
+        %Q(this is useless header),
+        %Q(and more),
+        %Q(num,str,quoted_num,time),
+        %Q(1, "value with space "" and quote in it", "123",21150312000000Z),
+        %Q(2),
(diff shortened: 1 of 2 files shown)


uu59:
@frsyuki do you have any idea for I described csv guess bahavior as above is intentional or not?

test/guess/test_csv_guess.rb
```diff
@@ -0,0 +1,153 @@
+require 'helper'
+require 'time'
+require 'embulk/guess/csv'
+
+class CsvGuessTest < ::Test::Unit::TestCase
+  class TestDelimiter < self
+    data(
+      "\t" => "\t",
+      "," => "\t",
```
frsyuki: Do you mean "," here?

frsyuki:
Looks very good. I commented one thing.
It would be great if you can add tests for corner cases such as `1, "value with space "" and quote in it "  ,"123",21150312T0000Z`

test/guess/test_csv_guess.rb
uu59: Oops. thank you finding it!

uu59:
Thanks review. I've added test for complex case.
How about `escape` option?

frsyuki:
Looks good. I fixed `escape` at #354

PR #357, opened 2015-12-25
Add system config gem_home

hito4t:
Add system config 'gem_home' to set `GEM_HOME` because `embulk_bundle.rb` is not called when `EmbulkEmbed` is directly executed.
If `GEM_HOME` is not set, installed plugins won't be registered.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/jruby/JRubyScriptingModule.java
+++ b/embulk-core/src/main/java/org/embulk/jruby/JRubyScriptingModule.java
@@ -1,9 +1,9 @@
 package org.embulk.jruby;
 
-import java.util.List;
-import java.util.ArrayList;
 import java.util.Set;
+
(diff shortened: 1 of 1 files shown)


grimrose:
:+1:

frsyuki:
JRuby's `ENV.store('GEM_HOME', gemHome)` modifies environment variable of current process. ENV is process-global variable (not injector-local variable). So, modifying ENV there seems unexpected behavior because a process can create multiple injectors.

frsyuki:
if useGlobalRubyRuntime is true, modifying ENV is probably valid.

frsyuki:
@hito4t @grimrose I created #376. Does it work for you?

hito4t:
Thank you!
It works!

grimrose:
:+1:

frsyuki:
Thanks. Merged #376.

PR #358, opened 2015-12-29
Fix bug of trimming in CsvTokenizer 

hito4t:
When `trim_if_not_quoted` is true, `AB C` will be parsed as "".


frsyuki:
oops! the comments is true but code was not.
:+1:

PR #361, opened 2016-01-09
[Propose] Add checkstyle settings in Java plugin generator

hiroyuki-sato:
Add recommend codestyle setting by default in Java generator.

* They want to unify codestyle. [Twitter](https://twitter.com/Civitaspo/status/685048411159019521)(Japanese)
* Some people want to know what kind of codestyle use.
  [Twitter](https://twitter.com/hanpen_good/status/623266724763140096)(Japanese)

This is also follow-up https://github.com/embulk/embulk-executor-mapreduce/issues/15


civitaspo:
+1 :+1:

frsyuki:
Thank you. But I think that it's too much to make test failed when checkstyle fails. I think checkstyle should be warning.
So I created #364.

PR #375, opened 2016-01-14
Avoid /0 when no input task

hito4t:
When input task count is 0 (ex. input file for file input plugin is empty), following exception will be thrown.

<pre>
Caused by: java.lang.ArithmeticException: / by zero
	at org.embulk.exec.LocalExecutorPlugin.newExecutor(LocalExecutorPlugin.java:64)
	at org.embulk.exec.LocalExecutorPlugin.transaction(LocalExecutorPlugin.java:53)
</pre>


frsyuki:
Thank you :+1:

PR #385, opened 2016-02-02
Fix embulk run config.yml -c diff.yml fail.

hiroyuki-sato:
``embulk run config.yml -c diff.yml`` command fail if ``diff.yml``  not exists.

[test code](https://gist.github.com/hiroyuki-sato/64792328449d6099953a)

This code create empty file. 
https://github.com/embulk/embulk/blob/master/lib/embulk/runner.rb#L61

So I ignore if file size == 0


frsyuki:
:+1:

hiroyuki-sato:
Thanks!

PR #388, opened 2016-02-08
Close GzipFileDecoder/GZIPInputStream and GzipFileEncoder/GZIPOutputStream

sakama:
It seems that GZIPInputStream and GZIPOutputStream is not closed at both of GzipFileDecoder and GzipFileEncoder plugins.
I fixed it.

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/GzipFileDecoderPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/GzipFileDecoderPlugin.java
@@ -38,17 +38,21 @@ public class GzipFileDecoderPlugin
         return new InputStreamFileInput(
                 task.getBufferAllocator(),
                 new InputStreamFileInput.Provider() {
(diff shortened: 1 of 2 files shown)


embulk-standards/src/main/java/org/embulk/standards/GzipFileDecoderPlugin.java
```diff
@@ -38,17 +38,21 @@ public FileInput open(TaskSource taskSource, FileInput fileInput)
...
                             return null;
                         }
-                        return new GZIPInputStream(files, 8*1024);
+                        gzis = new GZIPInputStream(files, 8*1024);
+                        return gzis;
                     }
 
                     public void close() throws IOException
                     {
                         files.close();
```
frsyuki: GZIPInputStream closes underlaying InputStream. Closing `gzis` closes `files`. So, closing `files` should be unnecessary.

sakama:
I got it. thanks!

PR #389, opened 2016-02-09
Workaround for deadlock issue in require method of rubygems

frsyuki:
Serializes access to Gem.loaded_specs to mitigate probability of
possible deadlock.


muga:
:+1:

PR #390, opened 2016-02-09
Update recipe. Use -c option instead of -o. And update ES 2.2

hiroyuki-sato:
Update tutorial. use ``-c`` option instead of ``-o``

Before: ``embulk run next-config.yml -o /path/to/next-config.yml``
After: ``embulk run config.yml -c diff.yml``

And Update ES 2.2


frsyuki:
Great! Thank you.

PR #392, opened 2016-02-12
Add JsonParser.Stream for streaming deserialization of JSON

frsyuki:
Usage at a parser plugin:

```java
FileInputInputStream in = new FileInputInputStream(fileInput);
while (in.nextFile()) {
    try {
        JsonParser.Stream stream = new JsonParser().open(in);
        Value value;
        while ((value = stream.next()) != null) {
            ...
        }
    }
    catch (IOException ex) {
        throw Throwables.propagate(ex);
    }
}
```


muga:
:+1:

PR #397, opened 2016-02-19
Add unit tests for PageReader/PageBulder

sakama:
I added unit tests for PageReader and PageBuilder

#### org.embulk.spi.PageReader 94%
page_reader

#### org.embulk.spi.PageBuilder 99%
page_builder

#### org.embulk.spi.json 35%
spi_json


frsyuki:
Good job :+1:

PR #396, opened 2016-02-18
Add JsonParserPlugin and JsonGuessPlugin

muga:

PR #400, opened 2016-02-22
Add JsonFormatterPlugin

sakama:
JsonParserPlugin and JsonGuessPlugin was merged to embulk-standards https://github.com/embulk/embulk/pull/396.

I wrote JsonFormatterPlugin.


muga:
Nice. I have one comment about the unit test. I think that it's better to mock and use LineEncoder to compare expected and actual.

sakama:
Thanks.
I had no idea to get formatted JSON string.
I'll try that you're saying.

embulk-standards/src/test/java/org/embulk/standards/TestJsonFormatterPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+                PageOutput pageOutput = plugin.open(taskSource, schema, output);
+                List<Page> pages = buildPage(schema,
+                        123.45, "embulk", 1L, true, Timestamp.ofEpochSecond(1422386629), getSampleJson(),
+                        null, null, null, null, null, null
+                );
+                for (Page page : pages) {
+                    pageOutput.add(page);
+                }
+                pageOutput.finish();
+                List<Object[]> records =  Pages.toObjects(schema, pages);
```
frsyuki: what does this code test?

sakama: Build PageBuilder and read from PageReader.
But It's better to compare against formatted JSON string.
```
assertEquals("{123.45, \"embulk\", 1, true...}", val)
```

So I'm thinking about to add MockLineEncoder as @muga  is saying [at above](https://github.com/embulk/embulk/pull/400#issuecomment-187068463)

embulk-standards/src/main/java/org/embulk/standards/JsonFormatterPlugin.java
```diff
@@ -0,0 +1,168 @@
...
+                            } else {
+                                addNull(column.getName());
+                            }
+                        }
+
+                        public void jsonColumn(Column column)
+                        {
+                            if (!pageReader.isNull(column)) {
+                                Value value = pageReader.getJson(column);
+                                addValue(column.getName(), value.toJson());
```
frsyuki: this is unexpected because json value becomes a string.
For example, if a record has two columns ("time:timestamp" and "params:json"), and values are `time=2016-01-01 00:00:00`, `params={"q":"test"}`, a generated file should be: `{"time":"2016-01-01T00:00:00Z","params":{"q":"test"}`. It shouldn't be: `{"time":"2016-01-01T00:00:00Z","params":"{\"q\":\"test\"}"`.

embulk-standards/src/test/java/org/embulk/standards/TestJsonFormatterPlugin.java
frsyuki: I don't know why you needed to test PageReader here.

sakama:
This PR is very old.
I close this one.

PR #404, opened 2016-02-26
Fix uninitialized constant MessagePack

sonots:
Using json type with jruby plugin, I got `Fix uninitialized constant MessagePack` as https://gist.github.com/sonots/617c7cf246a0c522651c. 

I am not sure a correct way to fix, but this patch fixed the issue anyway.


frsyuki:
Thanks!

PR #405, opened 2016-02-27
Fix possible busy loop at TaskDeserializer

frsyuki:
TaskDeserializer checks existance of JsonToken.END_OBJECT to break a
loop. But END_OBJECT never comes if cursor is not a object.

This configuration reproduce the problem:

    filters:
      - type: add_time
        to_column:
          name: time
        from_column:
          foo


muga:
:+1:

PR #409, opened 2016-03-04
LineDecoder skips UTF-8 BOM if necessary

frsyuki:
UTF-8 encoding allows to have BOM (0xEF 0xBB 0xBF) at the header although it does nothing at all.
BufferedReader returns 0xFE 0xFF when it detects a BOM at the head of UTF-8 file.


embulk-core/src/main/java/org/embulk/spi/util/LineDecoder.java
```diff
@@ -46,7 +49,43 @@ public LineDecoder(FileInput in, DecoderTask task)
...
+                // rollback to the marked position
+                try {
+                    reader.reset();
+                }
+                catch (IOException ex) {
+                    // unexpected
+                    throw new RuntimeException(ex);
+                }
+            }
+        }
```
muga: :+1:

PR #422, opened 2016-04-26
Doc: Liquid template not experimental

hiroyuki-sato:
http://www.embulk.org/docs/built-in.html#using-variables

> (This is experimental feature. Behavior might change or be removed in future releases).

A user confused about **experimental**.
(From [Twitter](https://twitter.com/satoshihirose/status/724889065909268480) Japanese text).

I think Liquid is not experimental feature.


hiroyuki-sato:
I asked frsyuki about this at TDTech.
He said this is still the experimental feature.
(Because he is planning to implement new configuration format).

So I close this PR.

PR #424, opened 2016-05-11
Accept time offset at getting Timestamp value

miniway:
We might need to adjust timestamp value in various cases. For example, time value in log records can be wrong by invalid machine time.Or time value only in log records but date value can exist in file path.


muga:
Good feedback. This feature was implemented on #425

PR #425, opened 2016-05-12
Added 'date' option to timestamp type field

frsyuki:
This option is useful when a timestamp string doesn't include date and
user wants to set date in a config file. For example, a CSV file
includes "02:46:29" and a config file includes "date: 2016-02-03".


muga:
LGTM.

PR #431, opened 2016-06-09
Check the empty string as a null string

kysnm:
Currently, empty string (without QuotedString) is always treated as a null regardless of the setting of the null_string. Empty string should be treated as `""` if null_string is `NULL` or `\N` or something.


sonots:
I will take over this at https://github.com/embulk/embulk/pull/432

PR #432, opened 2016-06-09
CSV Parser: Do not always convert empty string to null

sonots:
Currently, empty string is always converted into null even if I do not specify `null_string` option, or I specify `null_string: NULL`.

I prefer csv parser converts an empty string only if I specify its option as `null_string: ""`, and I believe it is the correct specification of csv parser which this document http://www.embulk.org/docs/built-in.html#csv-parser-plugin describes.


sonots:
Please let me know if I am misunderstanding the specification.

frsyuki:
CSV parser plugin converts a non-quoted empty string to null but quoted empty string is kept an empty string. CSV parser plugin assumes that's the most reasonable default behavior to deal with NULLs because it provides a clear way to distinguish NULLs from any strings clearly.
Changing meaning of null_string option is probably good. But your preference doesn't seems good enough to change this default behavior. Please open an issue for further discussion.

Issue #433, opened 2016-06-10
CSV parser always converts a non-quoted empty string to null regardless of `null_string`


sonots:
Discussion continued from https://github.com/embulk/embulk/pull/432.

Currently, there is no way not to convert a non-quoted emtpy string to null.
Even if we specify `null_stirng: "NULL"` or something else, a non-quoted emtpy string is always converted into null. We need an option to keep a non-quoted empty string as an empty string.

sonots:
> CSV parser plugin converts a non-quoted empty string to null but quoted empty string is kept an empty string. 

Keeping this default behavior is fine.

frsyuki:
> Keeping this default behavior is fine, and this behavior must be written on the document 
Sounds good. Let's take these ways.

sonots:
Thanks. And. my problem is not for the default behavior. The problem appears when we specify `null_stirng: "NULL"` or something. It still converts a non-quoted empty string to null.

frsyuki:
ok. it was unclear from the PR that changes the default behavior.
I think the easiest implementation is to change the meaning of `null_string:` option so that default behavior (`null_string: null`) makes non-quoted strings to NULL, setting empty string (`null_string: ""`) makes any empty strings to NULL regardless of quotation, and setting other strings (`null_strings: abc`) makes the string to NULL regardless of quotation.

sonots:
Agree with the specification. I will resend PR.

> ok. it was unclear from the PR that changes the default behavior.

yep, I mistakenly changed the default behavior with #432 because I did not know `CSV parser plugin converts a non-quoted empty string to null but quoted empty string is kept an empty string.` was its specification.

PR #434, opened 2016-06-10
CSV Parser: Do not always convert empty string into null (revisit)

sonots:
Fix https://github.com/embulk/embulk/issues/433

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/CsvParserPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/CsvParserPlugin.java
@@ -233,7 +233,6 @@ public class CsvParserPlugin
         final TimestampParser[] timestampParsers = Timestamps.newTimestampColumnParsers(task, task.getSchemaConfig());
         final JsonParser jsonParser = new JsonParser();
         final CsvTokenizer tokenizer = new CsvTokenizer(new LineDecoder(input, task), task);
-        final String nullStringOrNull = task.getNullString().orNull();
         final boolean allowOptionalColumns = task.getAllowOptionalColumns();
         final boolean allowExtraColumns = task.getAllowExtraColumns();
         final boolean stopOnInvalidRecord = task.getStopOnInvalidRecord();
@@ -261,7 +260,7 @@ public class CsvParserPlugin
                             public void booleanColumn(Column column)
                             {
                                 String v = nextColumn();
-                                if (v == null) {
+                                if (v == null || v.isEmpty()) {
                                     pageBuilder.setNull(column);
                                 } else {
                                     pageBuilder.setBoolean(column, TRUE_STRINGS.contains(v));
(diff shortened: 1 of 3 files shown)


embulk-standards/src/main/java/org/embulk/standards/CsvParserPlugin.java
```diff
@@ -261,7 +260,7 @@ public void run(TaskSource taskSource, final Schema schema,
                             public void booleanColumn(Column column)
                             {
                                 String v = nextColumn();
-                                if (v == null) {
+                                if (v == null || v.isEmpty()) {
```
sonots: Previously, an empty string for boolean, long, double, timestamp, json columns behaved as bellow with `null_string: "NULL"`.

1. assign null for a non-quoted empty string, and raises an error for a quoted empty string.

This PR makes empty strings for both a non-quoted string and a quoted empty string with `null_string: "NULL"` for any kinds of columns once. We keep the empty string for string columns, but we have to treat correctly for boolean, long, double, timestamp, and json columns. I think there are two possible specifications.

1. raises an error for both a non-quoted empty string and a quoted  empty string
2. assign null for both a non-quoted empty string and a quoted empty string

I chose the 2nd.

sonots: I agree 1 is technically the "correct" behavior.
I re-thought for use cases of 2, and I could find no use case for it.
I changed the behavior to 1.

embulk-standards/src/main/java/org/embulk/standards/CsvTokenizer.java
```diff
@@ -324,6 +326,28 @@ public String nextColumn()
         }
     }
 
+    public String nextColumnOrNull()
+    {
+        String v = nextColumn();
+        if (v == null) {
```
frsyuki: minor: this `v == null` check is unnecessary.

embulk-standards/src/main/java/org/embulk/standards/CsvTokenizer.java
```diff
@@ -324,6 +326,28 @@ public String nextColumn()
...
+            return null;
+        }
+        else if (v.isEmpty() && nullStringOrNull == null) {
```
frsyuki: minor: `nullStringOrNull == null` check can be moved to outer `if` block so that following `v.equals(nullStringOrNull)` check is also skipped if `nullStringOrNull` is null.

frsyuki:
@sakama could you also review this PR? This is a good improvement in my opinion.

embulk-standards/src/main/java/org/embulk/standards/CsvTokenizer.java
sonots: fixed

sonots:
@sakama ping

muga:
@sonots This is minor thing. Can you remove CsvTokenizer#wasQuotedColumn public method? It will not be used any more.

muga:
@sonots LGTM 👍

sonots:
> Can you remove CsvTokenizer#wasQuotedColumn public method? 

done

frsyuki:
thanks!

PR #437, opened 2016-07-02
CSV Parser: CSV Tokenizer should set lineNumber to 0

muga:
CSV Tokenizer should set lineNumber to 0 when CSV Parser's nextFile method is called.


frsyuki:
👍

Issue #415, opened 2016-03-28
Exception: Self-suppression not permitted


LanceNorskog:
This is the stack trace printed by a Python program that calls Embulk via a custom web service. (This is why the start of each line is not familiar.) From what I can learn on the net, either Embulk or the JDBC driver is tripping over a quirk of the Java 7 try-with-resources syntax.

Does anybody know what to do is this situation?

Thanks,

Lance Norskog

2016-03-28 08:09:08.560 +0000 [INFO] (embulk-output-executor-3): > 0.00 seconds

[2016-03-28 08:09:08,562] {EmbulkWSPlugin.py:82} INFO - 2016-03-28 08:09:08.560 +0000 [INFO] (embulk-output-executor-3): Running COPY from file tmp/fatbird_deliveries/e591af19-0982-4093-8b3b-9b0f9fe74f88

[2016-03-28 08:09:15,428] {EmbulkWSPlugin.py:82} INFO - 2016-03-28 08:09:15.410 +0000 [INFO] (embulk-output-executor-3): Loaded file tmp/fatbird_deliveries/e591af19-0982-4093-8b3b-9b0f9fe74f88 (6.85 seconds for COPY)

[2016-03-28 08:09:15,492] {EmbulkWSPlugin.py:82} INFO - 2016-03-28 08:09:15.486 +0000 [INFO] (embulk-output-executor-1): Uploading file id tmp/fatbird_deliveries/14b18961-5c3f-4bca-86c3-67d899c176f6 to S3 (16,785,349 bytes 98,243 rows)

[2016-03-28 08:09:16,008] {EmbulkWSPlugin.py:82} INFO - 2016-03-28 08:09:16.002 +0000 [INFO] (embulk-output-executor-1): SQL: SET search_path TO "fatbird"

[2016-03-28 08:09:16,008] {EmbulkWSPlugin.py:82} INFO - 2016-03-28 08:09:16.004 +0000 [INFO] (embulk-output-executor-1): > 0.00 seconds

[2016-03-28 08:09:16,008] {EmbulkWSPlugin.py:82} INFO - 2016-03-28 08:09:16.004 +0000 [INFO] (embulk-output-executor-1): Running COPY from file tmp/fatbird_deliveries/14b18961-5c3f-4bca-86c3-67d899c176f6

[...]

PR #438, opened 2016-07-07
Avoid IllegalArgumentException in try-with-resource

hata:
When using try-with-resource block and throw Exception in both the block
and close() method, `IllegalArgumentException: Self-suppression not
permitted` is thrown. LocalExecutorPlugin throws the exception when
OutputPlugin throws RuntimeException. RuntimeException is caught by
ExecutorService and the result is returned via Future object in
OutputWorkers. When OutputPlugin throws RuntimeException, Future.get()
throws ExecutionException. The exception is handled by
ScatterTransactionalPageOutput.completeWorkers(). completeWorkers()
are called by finish(), close(), commit(), and abort().
When finish()(or commit()) throws RuntimeException via completeWorkers(),
close()(and abort() via AbortTransactionResource) also throws
RuntimeException. From this, IAE is thrown.

This fix is to store a first RuntimeException as reportedException and
then check it in close() or abort() to avoid throwing the same exception
again.

I think this code may fix Issue #415 although I didn't test it using the same plugin.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/LocalExecutorPlugin.java
+++ b/embulk-core/src/main/java/org/embulk/exec/LocalExecutorPlugin.java
@@ -424,6 +424,8 @@ public class LocalExecutorPlugin
 
         private long pageCount;
 
+        private RuntimeException reportedException;
+
         public ScatterTransactionalPageOutput(ProcessState state, int taskIndex, int scatterCount)
         {
             this.state = state;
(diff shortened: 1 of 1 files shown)


hata:
FYI:
In my local environment, I could see the same failures using `./gradlew rubyTest` command. And my change only relates to handling RuntimeException. So, I guess the failure comes from another change.

hata:
FYI: I filed a new issue #439 and commented the first failure commit.

frsyuki:
Isn't the root cause `completeWorkers()` throwing the same exception twice?
`ScatterTransactionalPageOutput.close` should not throw RuntimeException when `ScatterTransactionalPageOutput.finish` (or commit or abort)` throw an exception. How do you think with #446?

hata:
@frsyuki You are correct and Your fix looks very good. I will close this request because your fix is good enough to me :)

Issue #439, opened 2016-07-07
Failure of rubyTest after removing CsvTokenizer.wasQuotedColumn()


hata:
commit ff8925337b6d8be3e142f06a08d6db917762b0e8 removed CsvTokenizer.wasQuotedColumn() causes the following two failure in rubyTest .

- Failure: test_complex_line(CsvGuessTest::TestColumns)
- Failure: test_trim_flag_when_will_be_long_if_strip_arround_space(CsvGuessTest::TestTrim)

TestCase

git clone https://github.com/embulk/embulk.git
cd embulk
git checkout ff8925337b6d8be3e142f06a08d6db917762b0e8
./gradlew clean; ./gradlew gem; ./gradlew rubyTest

You can see the two failures.

If you run the following command(The commit is just the previous failure commit), then this problem doesn't occur.

git checkout 44cd2341bdd6c8508c3f27ca645dcbde36c14ae8

You will see no failure.

hata:
I tested to revert the failure commit and reverted code could pass the tests.

git revert ff8925337b6d8be3e142f06a08d6db917762b0e8

FYI:

The following ruby code access wasQuotedColumn.

`embulk/lib/embulk/guess/csv.rb`

```
                  column = tokenizer.nextColumn
                  quoted = tokenizer.wasQuotedColumn # Use the removed method.
                  if null_string && !quoted && column == null_string
                    column = nil
                  end
```

PR #440, opened 2016-07-07
Revert "remove unused public wasQuotedColumn()"

sonots:
since guess uses `CsvTokenizer.wasQuotedColumn()`

Fix https://github.com/embulk/embulk/issues/439

cc: @muga


muga:
@sonots @hata @frsyuki Merged. Thank you for fixing my fault.

PR #444, opened 2016-07-16
Preview: Limit the size of buffer that PreviewExecutor reads

muga:
PreviewExecutor limits the # of records that it reads by Page basis. The limitation doesn't work in the case that PreviewExecutor tries to read a big JSON array in the file (or a big JSON object in the file like CloudTrail files). This PR allows PreviewExecutor to limit the size of buffer that it reads from the files.


muga:
follow up 4bb581bee299da36495edd8ddb64ee0b97d966cf

frsyuki:
Looks good 👍  Please merge.

muga:
merged. thank you for reviewing.

PR #446, opened 2016-07-19
fixes 'IllegalArgumentException: Self-suppression not permitted' when output plugin throws a runtime exception

frsyuki:
Fixes #438.


hata:
Hi Sadayuki,

Thank you very much for creating the new fix. I think your fix can fix #438 and it looks good to me. I will close #438  soon.

PR #453, opened 2016-07-26
[doc] Add null_string option description

hiroyuki-sato:
Add ``null_string`` description. 
Follow-up: https://github.com/embulk/embulk/issues/433#issuecomment-225265664


frsyuki:
👍

hiroyuki-sato:
Thanks!

PR #456, opened 2016-08-02
Allow type coalesce from timestamp to long at guess

frsyuki:
This fixes unexpected guess results of integer types.
Gussing `[1456012800, 1456014800]` is expected to be long but it becomes
string because `1456012800` is guessed as timestamp with `%Y%m%d%H`
format, `1456014800` is guessed as long, and type coalescing from timestamp
to long was not allowed. This adds coalescing from timestamp to long to
fix this issue.


muga:
LGTM:+1:

Issue #420, opened 2016-04-21
[Feature Request] Binary type support


tomykaira:
I am trying to handle msgpack encoded data with Embulk.
#121 mentions about binary type, but there are no progress.
Because I can control both input and output plugin, I consider Base64 encoding, but it is not simple solution and not applicable for all cases.
It would be great if embulk natively supports binary data type.

toyama0919:
+1

some use case, 
* Image recognition(deep learning)
* Image converter(like Imagemagick)

It is easy to do a little on Embulk.

frsyuki:
Sounds good idea 👍 
Similar to JSON type support, we'll also need toolset around binary columns such as:

* string conversion: base64/hex column encoder/decoder
* compression: gzip/lz4/snappy column encoder/decoder
* hash: md5/sha1/sha2 column encoder/decoder

PR #458, opened 2016-08-09
[WIP] Introduce binary type

tomykaira:
This approaches #420.

The largest problem is that many plugins will be not compatible when this new type is introduced.

I would like to discuss how the binary type should behave, and how to merge this into mainline. Thanks, all.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/Column.java
+++ b/embulk-core/src/main/java/org/embulk/spi/Column.java
@@ -4,6 +4,7 @@ import java.util.Objects;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.embulk.spi.type.Type;
+import org.embulk.spi.type.BinaryType;
 import org.embulk.spi.type.BooleanType;
 import org.embulk.spi.type.DoubleType;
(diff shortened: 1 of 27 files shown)


dallin:
Really looking forward to this feature being added.

tomykaira:
I cannot commit to this feature any more. I really appreciate if someone take over to be merged!

muga:
To avoid unexpected LinkageError, we should upgrade Java8 before binary type support.

dmikurube:
We're going to introduce Binary type in v0.9, which is expected to be released in a couple of months.

Thanks to this PR, the first implementation would be easy. But I'm afraid we may be going to make another PR to introduce Binary type since:

* It's conflicting with the current `master` at Git's level. (It's because we took much time to do it. Sorry about that.)
* Some interfaces have been changed in Embulk. (Ditto. Sorry.)
* We'd use another class for Binary (`java.nio.ByteBuffer` or Embulk's own class), instead of `byte[]`. (`byte[]` is not very good to use there -- it's mutable, and too primitive.)

Let me take this opportunity to appreciate @tomykaira 's effort on this. Thanks for your contribution!

dmikurube:
New commits to be made in #850.

PR #460, opened 2016-08-12
Embulk new logo

frsyuki:
New logo!!!
Do you like it?


muga:
cool!!! 👍

toru-takahashi:
I think some guys want to know why we change this logo.

frsyuki:
That's a good point. There're are three things behind the new logo.
First thing is that having characters ("embulk") always together with the orca was making it hard to fit in them a small area. Thus Embulk logo has been used in relatively small size compared to others.
Second thing is that Embulk is used by big enterprise companies more than I expected (even some banks...did you know that?). I want to keep the cute looking as the symbol of open, friendly, and active OSS community. But at the same time, the designer (@johntrabelsi, my friend and colleague) worked hard to make it more professional looking.
The other thing is combination with the [new Fluentd logo](https://github.com/fluent/fluentd/issues/924). Fluentd is another OSS I initially designed. Naturally, they share the same fundamental philosophy. So I wanted make them looked like a team when the logs are used together.

PR #462, opened 2016-08-13
Separate embulk-docs to another repo

frsyuki:
Separating documents into another repository (at https://github.com/embulk/embulk-docs).
The purpose is to make it easy to write documents separately from embulk's core development.

Discussion points:
- How to deploy JavaDoc and RDoc to the same site...?
- How to make sure that release note is on the docs site?


muga:
As talked with @frsyuki and @dmikurube, it's not easy to sync the source code and docs by separating repos. We're going to putting the source code and docs on same repo for now. If we will have a strong motivation to move the docs to another, we will re-think about it at that time.

PR #465, opened 2016-08-16
Add unit.TimestampParam class for plugins

frsyuki:
This unit class is useful to get a timestamp value as a config
parameter. It accepts intersection of ISO 8601 & RFC 3339 and UNIX
timestamp in seconds.


sonots:
Just an idea. If we think of parsing fractional seconds by ourselves, it is possible to treat nano second as I did in embulkf-filter-timestamp_format https://github.com/sonots/embulk-filter-timestamp_format/blob/master/src/main/java/org/embulk/filter/timestamp_format/TimestampParser.java#L177-L187

PR #470, opened 2016-09-16
Ignore IOException when embulk closes input stream in `reopen`

joker1007:
Because if input stream is already closed, `in.close()` may raise IOException.
In such a situation, Reopener cannot reopen even if input stream is resumable.

ex. SSL connection of embulk-input-gcs is sometimes closed by server side.


frsyuki:
Good 👍

frsyuki:
follow-up: https://github.com/embulk/embulk/pull/472

joker1007:
Thank you for merging and following PR !!

PR #474, opened 2016-09-22
let -g guess plugins run first

frsyuki:


muga:
I close this PR because this could be replaced with `exclude_guess_plugins` option. Please open it if needed in the future.

PR #479, opened 2016-10-01
Rename methods that Embulk generates to avoid name conflict with JRuby

muga:
Since the names of methods that Embulk genetes are simple like `java_class`, `java_object`, etc, there might be name conflict with code generation libraries in the future. To avoid the issues, it's better to rename such methods in advance.


frsyuki:
LGTM 👍

PR #480, opened 2016-10-02
Upgrade JRuby to 9.1.5.0

muga:
How do you think about the upgrading? We're using v9.1.2.0 now. The following is my concerns:

* v9.1.3.0 has a lot of changes from v9.1.2.0. It includes the fix of a critical bug (like thread-safe regression in autoload and require) that was reported on v9.1.2.0. And the upgrading requires https://github.com/embulk/embulk/pull/479
* It seems that v9.1.4.0 and v0.1.5.0 are minor changes from v9.1.3.0.


frsyuki:
LGTM!

PR #483, opened 2016-10-05
Add the "rules" element in RenameFilterPlugin

dmikurube:
Add the "rules" element in RenameFilterPlugin with initial renaming operators "lower_case_to_upper" and "upper_case_to_lower".

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/RenameFilterPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/RenameFilterPlugin.java
@@ -11,10 +11,7 @@ import org.embulk.spi.FilterPlugin;
 import org.embulk.spi.PageOutput;
 import org.embulk.spi.Schema;
 
-import com.google.common.collect.ImmutableMap;
-
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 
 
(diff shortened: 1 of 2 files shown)


dmikurube:
@uu59 Can you please take a look? I'm especially wondering if ConfigException is good or not here.

uu59:
Correctly used 👍

embulk-standards/src/main/java/org/embulk/standards/RenameFilterPlugin.java
```diff
@@ -53,4 +70,38 @@ public PageOutput open(TaskSource taskSource, Schema inputSchema,
...
+            }
+        } else {
+            throw new ConfigException("Non-string renaming operator is not supported");
+        }
+    }
+
+    private Schema applyUpperCaseToLower(Schema inputSchema) {
+        Schema.Builder builder = Schema.builder();
+        for (Column column : inputSchema.getColumns()) {
+            builder.add(column.getName().toLowerCase(), column.getType());
```
uu59: type: timestamp column has a `format`. Probably it should be kept after rename filter through.

uu59: @muga let me correct if I'm wrong

dmikurube: If so, existing "columns" should be fixed as well?

dmikurube:
@uu59 Chained rules need to be tested, but the combination of upper_case_to_lower and lower_case_to_upper may not work to test chaining. (Only the latter is meaningful!)

Will add that test with next operators.

embulk-standards/src/main/java/org/embulk/standards/RenameFilterPlugin.java
uu59: @dmikurube Ah sorry you are right. Please ignore my comment

dmikurube:
@uu59 Updated the PR per our discussion to accept map-formed rules.

btw, when or how do we usually decide to merge a PR in Embulk?

uu59:
@dmikurube well, I think one of Embulk maintainers (especially @muga or @frsyuki?) approval then merge this by themselves normally.

dmikurube:
Got it. Thanks!

dmikurube:
@muga @frsyuki I'm just wondering if I could merge this base PR at first, and then add more renaming rules in separate PRs. What do you think?

dmikurube:
Updated the PR per f2f discussion with @muga.

dmikurube:
@frsyuki  
Not to block this base PR due to the naming discussion, I've split the concrete renaming rules  ("convert_...") into another branch for the time being. https://github.com/dmikurube/embulk/tree/renamefilterplugin-case

What do you think of this base PR #483?

frsyuki:
I think that `columns` should run after `rules` because `columns` have more flexible control than `rules`.  Looks good excepting that 👍

PR #484, opened 2016-10-05
Fix the comment about checking column names.

dmikurube:


dmikurube:
Not sure this is right, but the name "column_options" is not used here.

muga:
Merged. thank you for fixing.

PR #485, opened 2016-10-05
Have Schema's internal list always immutable.

dmikurube:


dmikurube:
Not sure this change is correct, but I was concerned that the contents of the internal list `List<Column> columns` can be modified from outside.

dmikurube:
@muga Thanks for your comments! Updated the PR.

embulk-core/src/main/java/org/embulk/spi/Schema.java
```diff
@@ -31,14 +31,25 @@ public static Builder builder()
...
     @JsonCreator
     public Schema(List<Column> columns)
     {
-        this.columns = columns;
+        if (columns instanceof ImmutableList) {
+            @SuppressWarnings("unchecked")
+            ImmutableList<Column> columnsCasted = (ImmutableList)columns;
+            this.columns = columnsCasted;
+        } else {
+            this.columns = ImmutableList.copyOf(columns);
```
dmikurube: Done.

muga:
@dmikurube Merged. thank you.

PR #488, opened 2016-10-13
Add a new renaming rule "case".

dmikurube:


dmikurube:
@frsyuki From the discussion after #483, adding a renaming rule for lower/upper case. Can you take a look?

PR #489, opened 2016-10-20
Add csv_all_strings guess plugin

muga:
This guess plugin suggests column types within Csv files as string types. The default Csv guess plugin suggests column types unexpected by users. To execute Embulk correctly, users sometimes need to edit Embulk config manually and fix column types that users expect. But, some users don't want to edit Embulk config so much. 

This plugin is not included in the list of default guess plugins. And to use this plugin, users explicitly need to exclude default csv guess plugin from the list of default guess plugins like:

```
exec:
  exclude_guess_plugins: ['csv']
  guess_plugins: ['csv_all_strings']
in:
  type: file
  path_prefix: path/to/csvfile
```

### DIFF
--- a/lib/embulk/guess/csv.rb
+++ b/lib/embulk/guess/csv.rb
@@ -142,16 +142,20 @@ module Embulk
         schema = []
         column_names.zip(other_types).each do |name,type|
           if name && type
-            if type.is_a?(SchemaGuess::TimestampTypeMatch)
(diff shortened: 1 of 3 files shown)


frsyuki:
Looks good 👍 
Please add some documents as well.

muga:
Sure, please check it later. https://github.com/embulk/embulk/pull/493

PR #491, opened 2016-10-20
Add missing guess_plugins and exclude_guess_plugins docs

hiroyuki-sato:
Add `guess_plugins` and `exclude_guess_plugins` description.
Please fix if my English is wrong.

## Question.

About `guess_plugins` default value, Which is better(correct) ?
* `[]` [GuessExecutor.java](https://github.com/embulk/embulk/blob/master/embulk-core/src/main/java/org/embulk/exec/GuessExecutor.java#L52)
* `['gzip','bzip2','json','csv']` [StandardPluginModule.java](https://github.com/embulk/embulk/blob/master/embulk-standards/src/main/java/org/embulk/standards/StandardPluginModule.java#L52-L55)

## TODO

It's better to write about guess `csv_all_strings` #489 later. 

Is it better to add  `guess` section?

### DIFF
--- a/embulk-docs/src/built-in.rst
+++ b/embulk-docs/src/built-in.rst
@@ -606,13 +606,17 @@ The ``local`` executor plugin runs tasks using local threads. This is the only b
 Options
 ~~~~~~~~~~~~~~~~~~
 
-+------------------+----------+----------------------------------------------------------------------+--------------------------------------+
-| name             | type     | description                                                          | required?                            |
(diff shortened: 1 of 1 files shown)


muga:
Cool. I think, it's good to separate guess doc from local executor plugin doc. Here is working in progress: https://github.com/embulk/embulk/pull/493

hiroyuki-sato:
It's good idea. Thanks!.

PR #500, opened 2016-10-31
Throw ConfigException instead of AssertionError when task validation failed

muga:
When BVal validation for Task instantiation failed, ModelManager throws AssertionError. It's ok but, I think better to throw ConfigException instead because such validation error is not recoverable.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/config/ModelManager.java
+++ b/embulk-core/src/main/java/org/embulk/config/ModelManager.java
@@ -89,7 +89,12 @@ public class ModelManager
 
     public void validate(Object object)
     {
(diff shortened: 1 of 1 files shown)


frsyuki:
Looks good 👍 
Minor comment is that it seems better to move the catch statement to `ModelManager.validate` because catching Error is usually unexpected. Should be handled when it happens as soon as possible.
Please merge if you're good.

muga:
Sorry, I miss-understood that ModelManager doesn't throw AssertionError. It throws TaskValidationException correctly. But, TaskValidationException should inherit ConfigException. I will fix it later.

PR #502, opened 2016-11-02
Revert "changed default sequence_format of output-file to include dot…

frsyuki:
… at the end so that file_ext doesn't have to include dot"

This reverts commit 0c1f7ee51e4ef215b30999181ed689a763c8a5de.


hiroyuki-sato:
@dmikurube What do you think to apply this PR in 0.9?

dmikurube:
@hiroyuki-sato Not very positive. It's also a compatibility breaking as of now.

hiroyuki-sato:
@dmikurube Thanks. compatibility is important.

PR #503, opened 2016-11-02
Add embulk-test package

frsyuki:
This package is intended to be used from the test cases of Java-based
input plugins.

Support for other plugin types is desired but not implemented in this
change.

Example use of embulk-test is https://github.com/embulk/embulk-input-jdbc/pull/79


muga:
Looks good to me 👍 Merged.

PR #505, opened 2016-11-04
Add document for the new "rules" option in the built-in rename filter plugin.

dmikurube:
@muga The descriptions of all renaming rules are nested, and may be so long. I'd separate them out if it's not good.

### DIFF
--- a/embulk-docs/src/built-in.rst
+++ b/embulk-docs/src/built-in.rst
@@ -580,12 +580,13 @@ Options
 +---------+----------+----------------------------------------------------------------------+--------------------+
 | name    | type     | description                                                          | required?          |
 +=========+==========+======================================================================+====================+
-| columns | hash     | A map whose keys are existing column names. values are new names.    | ``{}`` by default  |
-+---------+----------+----------------------------------------------------------------------+--------------------+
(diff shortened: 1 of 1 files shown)


muga:
@dmikurube thank you! will check it soon.

embulk-docs/src/built-in.rst
```diff
@@ -582,6 +582,226 @@ Options
...
+    # This configuration suffixes numbers to duplicative column names. (Ex. ["column", "column", "column"] goes to ["column", "column_2", "column_3"].)
+    filters:
+      ...
+      - type: rename
+        rules:
+        - rule: unique_number_suffix
+
 
 Example
 ~~~~~~~~
```
muga: I prefer `""""""""` instead of `~~~~~~~~`. How about this?

This existing "Example" is good here as an example for combination with `columns` and `rules` options. But, I think, this is minor case. Because almost users use `rules` only.

dmikurube:
@muga Updated per discussion.

muga:
Thank you. I will merge it once the CI passed.

muga:
Merged.

PR #512, opened 2016-11-21
Make |EmbulkTestRuntime| accept the transaction time.

dmikurube:
@muga I'm not 100% confident this is the right way, but can we have a way to test with the transaction time specified?


muga:
@dmikurube sorry for the delay. Your PR should work but, currently we're working and refining new test framework, called embulk-test, on #513 and #514. Please wait for merging those PRs into master branch.

PR #513, opened 2016-11-23
embulk-test: Add RunInputBuilder

muga:
To test input plugin, we sometimes need custom out: config or exec: config. TestConfigSourceBuilder allows users to custom ConfigSource that they use for input plugin tests.


muga:
git rebase master and git push -f

PR #518, opened 2016-12-15
Remove limitation of minimum 40 bytes size limit of guessing

frsyuki:


muga:
cool:-) 

memo: this ticket is related to https://github.com/embulk/embulk/issues/519

PR #527, opened 2017-01-07
Add semicolon as one of delimiter list that CSV guess plugin suggests

muga:
Because semicolon is often used as delimiter char. 
This assumes https://github.com/embulk/embulk/pull/526.


muga:
Will take it on https://github.com/muga/embulk/pull/8

PR #530, opened 2017-01-11
Develop remove_columns filter plugin

muga:
Implemented 'remove_columns' filter plugin, which allows users to remove specified columns from schema, as one of builtin plugins. 
- Support "remove" and "keep" options. "remove" allows removing specified columns from schema. "keep" allows including specified columns only as schema. They are not multi-select.
- Also support "accept_unmatched_columns" flag, by default false. If false, the plugin throws ConfigException if users specify invalid column names as remove or keep columns. 
- Write tests based on embulk-test
- Add filters method in InputBuilder on 5a267186dae8714182a93789d3d5758bb058792c


hiroyuki-sato:
@muga 
Thank you for the great work. 
I'll appreciate if you tell me why this plugin develop.

This plugin seems useful plugin. However, this plugin similar to [embulk-filter-column](https://github.com/sonots/embulk-filter-column)
 So, I'm wondering why this plugin develop?

The built-in plugin is one of the advantages.

frsyuki:
I found that this plugin doesn't work if input schema includes duplicated columm names. Although duplicated column names itself is not recommended, it's better to handle well as a built-in plugin. This PR fixes it and also optimizes performance slightly: https://github.com/muga/embulk/pull/9

muga:
@hiroyuki-sato Thank you for asking about it. Since Embulk users often want this functionality, I think that we need to add and bundle it as a light-weight builtin plugin.

[sonots/embulk-filter-column](https://github.com/sonots/embulk-filter-column) has rich functionalities include this feature. It provides a lot of options to translate schema.
It would be useful for almost cases of schema translation. one of most popular plugins.

I think, personally, this builtin plugin could help beginners but, most of them will become less satisfied with it and then start using embulk-filter-column. Because builtin plugins provide standard feature and embulk-filter-column has rich functionalities. 

@frsyuki thank you for reviewing and fixing. I will take a look your PR and merge it.

hiroyuki-sato:
@muga 
Thank you for replying.
I understood like the following.

* embulk-filter-remove: simple & built-in. 
* embulk-filter-column: rich feature and external plugin.

muga:
Merged. please let me know if you have any questions.

PR #533, opened 2017-01-21
CsvGuessPlugin enables suggesting for a few rows

muga:
Current CsvGuessPlugin could not handle a few rows (i.e. only 2 or 3 lines in a CSV file) and suggest Embulk config appropriately. Because it doesn't use a last line string as sample lines and removes it from the sample lines.

In this PR, the change of guess_plugin.rb enables CsvGuessPlugin to use a last line string to suggest a config. If the sample data ends with newline string suggested, the string actually could be treated as a last line. 

This PR also fixes NewlineGuessPlugin and inherit guess method in TextGuessPlugin instead of guess_text method to suggest newline string. TextGuessPlugin#guess_text always appends CRLF to a sample text.


lib/embulk/guess_plugin.rb
```diff
@@ -113,7 +113,9 @@ def guess(config, sample)
         while line = decoder.poll
           sample_lines << line
         end
-        sample_lines.pop unless sample_lines.empty?  # last line can be partial
+        unless sample.end_with?(parser_task.getNewline.getString)
+          sample_lines.pop if sample_lines.empty? # last line is partial
+        end
```
frsyuki: Nice 👍

muga:
Thank you for reviewing. Fixed on b098936b38de82dbb91dea36f71718645e1aa2ff

muga:
git rebase master and git push -f

PR #534, opened 2017-01-23
CsvGuessPlugin enables suggesting for only one row

muga:
#533 allows CsvGuessPugin to suggest configs by a few rows. But, still cannot suggest by only one row. It's difficult for Embulk to suggest configs in this case because it cannot recognize that the one line is actually header or not. 

This PR enables suggesting configs for one line only. As the behavior of the CsvGuessPlugin, it recognizes data record for the line. Not header. For example, if the following line is given, 
```
1,32864,2015-01-27 19:23:49,20150127,embulk
```
Embulk will suggest:
```
skip_header_lines: 0
columns:
- {name: c0, type: long}
- {name: c1, type: long}
- {name: c2, type: timestamp, format: '%Y-%m-%d %H:%M:%S'}
- {name: c3, type: timestamp, format: '%Y%m%d'}
- {name: c4, type: string}
```


muga:
I talked about this PR by @frsyuki offline.

a. Not need to use `one_line_file` flag. Separate codes on whether `sample_lines.size` is 1 or not to make it simple to read.
b. The current code returns `{}` as guess failure if the guessed `column_types.size` <= 1. But, if size is 1, it doesn't need to return `{}`. Because it's not guess failure. just single column file. It can continue to guess config.

frsyuki:
LGTM. It would be even better if we have another test case in addition to `test_1_rows_with_trim_needed.csv`: there're more than 1 lines and column names and values have spaces.

muga:
Make sense. I will add it. Thank you.

muga:
rebase master and push -f

PR #543, opened 2017-01-27
Replace its command-line option parser with pure-Java.

dmikurube:
@muga Finally, it's (almost) ready for review while it would need some more updates yet... Can you have an initial quick look?


dmikurube:
Some fixes added.

embulk-cli/src/main/java/org/embulk/cli/EmbulkCommandLine.java
```diff
@@ -0,0 +1,268 @@
...
+                @SuppressWarnings("unchecked")
+                final ArrayList<String> newList = new ArrayList<String>((List<String>) existingValue);
+                newList.add(value);
+                this.systemConfig.put(key, Collections.unmodifiableList(newList));
+            }
+            else {
+                this.systemConfig.put(key, Arrays.asList(value));
+            }
+            return this;
+        }
```
muga: Users can store JSON data in system config potentially. `setSystemConfig` and `addSystemConfig` might not be good enough for any JSON data like Map, Integer. How do you think that `setSystemConfig` receives a system config path as an argument like config diff?

dmikurube: @muga As of the command line (existing `lib/embulk/command/embulk_run.rb`), it processes system configs given from the command line as strings or maps of strings. Even if a user wants to give an integer into `system_config`, the user needs to specify like `-Xsome_config=42`, and that `42` is processed as a string at least in `lib/embulk/command/embulk_run.rb`. (You can grep `system_config` on `lib/embulk/command/embulk_run.rb`.)

All the processes are the same (for the time being) after the new `EmbulkRun.java`. The `system_config`, which is given to Ruby `EmbulkRunner`, is converted into a  pure-Ruby object at: https://github.com/embulk/embulk/pull/543/files/25d5a0273cbe35d01aaef6bf882d021d242b3172#diff-305efcc3bb3a8a54312bb00f55e0c386R453

I wanted `EmbulkRun.java` depending less on other libraries, including Jackson. It should be working well in the same way with the existing `lib/embulk/command/embulk_run.rb`. What do you think?

dmikurube:
Thanks. Replied.

embulk-cli/src/main/java/org/embulk/cli/EmbulkCommandLine.java
muga: @dmikurube Ah understand. I didn't know that we already have options to configure system config. It's even better to keep same options as the current one. Thank you for explaining.

dmikurube:
@muga Thanks. Merging this finally...

PR #544, opened 2017-01-29
update download link to be https

cosmok:


hiroyuki-sato:
About SSL certificate

* The dl.embulk.org use CloudFlare CDN. 
* That certificate contain the DNS name `*.embulk.org`. so SSL certificate is valid. 

About dl.embulk.org

* This site is a redirect only server.
* This site redirect to https://dl.bintray.com.
* The dl.bintray.com use SSL. so You can download embulk via SSL connection.

I personally, it is not important. Because the dl.embulk.org connect real server on Heroku via http.
(As I remember)

But this site can change to https. It is better.

PR #545, opened 2017-02-12
TimestampFormatter uses JRuby's RubyDateFormatter instead of deprecated RubyDateFormat

muga:
Since RubyDateFormat's been deprecated, it's better to replace with RubyDateFormatter.


muga:
Close this PR. Because this doesn't match at v09.

PR #552, opened 2017-02-28
Fix typos in 0.7.8 release note

yoshihara:
I found 2 typos in [0.7.8 release note](http://www.embulk.org/docs/release/release-0.7.8.html) so fixed them.


muga:
Thank you!

PR #554, opened 2017-03-03
schema_guess.rb: Change a condition to guess double type

muga:
The current schema_guess suggests 1.01, 1.0003, 50.05 like float values string type values. This PR enables suggesting such values as double type values by using regular expression. We could use exponent to represent float values but, that is out of scope of this PR.


muga:
It enables suggesting float values with exponent as double type.

lib/embulk/guess/schema_guess.rb
```diff
@@ -62,11 +62,8 @@ def guess_type(str)
           return "long"
         end
 
-        if str.include?('.')
-          a, b = str.split(".", 2)
-          if a.to_i.to_s == a && b.to_i.to_s == b
-            return "double"
-          end
+        if str =~ /^[-+]?([1-9]\d*|0)(\.\d+)([eE][-+]?[0-9]+)?$/
```
dmikurube: 1. Can you leave a comment, the core intention of this regular expression?
2. Do you have any reason why it has both `\d` and `[0-9]`?
3. Can we have a help from some external libraries, or a kind of Ruby's or Java's standard parsers?
4. Do we have any explicit spec what strings are guessed as `double`? (I don't think so because it's just a guess, though.)

I was not sure what behavior is expected, for example:
* `0.0`
* `00.1`
* ` 12.0000`
* `1e+2`
* `1e-2`

dmikurube: Thought it may be more confident to test with `parseDouble` and `parseInt` -- if `parseDouble` succeeds and `parseInt` fails, it should be `double`, for example.

dmikurube: `parseDouble` accepts `(+/-)NaN` and `(+/-)Infinity`, though. We need to take care anyway... ;)

dmikurube: Or, Guava's `tryParse` may work fine.
https://github.com/google/guava/blob/4906fbbaeb22559563be02b0e765fda15426baff/guava/src/com/google/common/primitives/Doubles.java#L601-L633

muga: 1. This regex is to make better suggestion for double type instead of current one. The current behavior has a problem, that it could not suggest float values includes 0 after decimal point as double type like 1.01, 5.001. It's ok because this is just suggestion. But, if we could improve it, it would be better. I could use a strict spec for floating point like IEEE 754 for my regex but, since it might take a time, I skipped the implementation for now. Actually finite numbers, infinities, NaNs defined in IEEE 754 are not supported in my regex but, should be supported in the future. I will put this comments on the source code later.

2. Thank you. I will fix bad codes later.

3. I could not find better libraries as the result of my survey. Float in Ruby and parseDouble in Java were candidates but, they can parse/convert "001", "002" like string as float/double types. Since I think that it's better to suggest "001" and "002" as string type in Embulk, I don't use them and use the regex. 

4. Yea, actually we don't have spec but, IEEE 754 might be one candidate.

muga:
@dmikurube Thank you for your comments! I will fix it according to your comments.

lib/embulk/guess/schema_guess.rb
dmikurube: Got it. Then, it might be nice to have an explicit comment describing:
* It can be improved more later.
* It is intentionally rejecting "001" as a number.

I thought it's okay to "001" can be a number to be honest, but it should be determined by those who know the actual usage more.

To build a regex for that purpose, taking a look at [Guava's](https://github.com/google/guava/blob/4906fbbaeb22559563be02b0e765fda15426baff/guava/src/com/google/common/primitives/Doubles.java#L591-L599) may help instead of building from scratch. Dropping unnecessary parts from that may be easier.

Guava's license is Apache 2.0.

muga: > I thought it's okay to "001" can be a number to be honest, but it should be determined by those who know the actual usage more.

Agree.

> To build a regex for that purpose, taking a look at Guava's may help instead of building from scratch. Dropping unnecessary parts from that may be easier.

Thank you, that's much help for me.

muga: Fixed https://github.com/embulk/embulk/pull/554/commits/9a2548a9e9d85db77407dbdf3d3482488e653746 How about this?

muga:
@dmikurube Thank you for reviewing. The reason why we accept `NaN` and `Infinity` is because they are defined in IEEE 754 as special values. On the other hand, since it seems that `001` and `00.1` like values are not defined as float values, I don't accept them as float for now. Since this is the logic for guessing schema, users can manually fix the guessed schema if they don't want to use it. I think that we should suggest `001` and `00.1` as double type if some users want to do that.

PR #556, opened 2017-03-10
embulk/guess/csv.rb: it should fail if CSV parser cannot parse sample_lines

muga:
I got NoMethodError during CsvGuessPlugin is running. The cause of the issue is that CSV parser cannot parse the file. My CSV file is invalid as CSV standard format. But, it's not good to throw NoMethodError. Better that CsvGuessPlugin fails if the file cannot be parsed. 

```
org.jruby.exceptions.RaiseException: (NoMethodError) undefined method `count' for nil:NilClass
       at RUBY.guess_string_header_line(uri:classloader:/embulk/guess/csv.rb:336) ~[na:na]
       at uri_3a_classloader_3a_.embulk.guess.csv.guess_lines(uri:classloader:/embulk/guess/csv.rb:140) ~[na:na]
       at uri_3a_classloader_3a_.embulk.guess_plugin.guess(uri:classloader:/embulk/guess_plugin.rb:121) ~[na:na]
       at uri_3a_classloader_3a_.embulk.guess_plugin.guess(uri:classloader:/embulk/guess_plugin.rb:24) ~[na:na]
```


muga:
Added test to check the behavior.

lib/embulk/guess/csv.rb
```diff
@@ -104,6 +104,11 @@ def guess_lines(config, sample_lines)
 
         sample_records = split_lines(parser_guessed, true, sample_lines, delim, {})
 
+        # It should fail if CSV parser cannot parse sample_lines.
+        if sample_records.nil? || sample_records.empty?
```
dmikurube: What happens when an empty file (or a header-only file) is given as CSV?

I thought it can easily happen when users are creating multiple CSVs time-by-time.

muga: Embulk's guess behavior is different by whether the input plugin can create a Buffer object or not.

1. Even though Embulk tries to guess for an empty file, the input plugin could not create any Buffer objects. It throws NoSampleException. If several files exist, Embulk tries to read them until it creates a Buffer object.
2. On the other hand, If Embulk tries to guess for a header-only file, the input plugin could create a Buffer object includes the header line. Therefore it could suggest something.

This PR fixes a part of 2nd one. The input plugin could a Buffer object but, CsvParserPlugin could not parse the lines in the Buffer object. In this case, v0.8.18 throws `NoMethodError` as mentioned in the PR's description. This fix enables avoiding the unexpected error and returning empty as the guess result.

muga:
Merged. thank you!

PR #558, opened 2017-03-14
Un-JRuby the "example" subcommand.

dmikurube:
@muga Let's start small!

Trying to un-JRuby from the leaves -- the `example` subcommand was a good example. :)


dmikurube:
@muga Working version. Can you PTAL?

dmikurube:
Thanks! :) Merging.

PR #561, opened 2017-03-15
Add resources for developers with some design docs on "Embulk v2"

dmikurube:
@muga Adding resources for developers, with "Embulk v2" resources.

### DIFF
--- a/embulk-docs/src/developers/index.rst
+++ b/embulk-docs/src/developers/index.rst
@@ -1,0 +1,38 @@
+For core/plugin developers
+===========================
+
+Plugin API documents
+---------------------
+
+* `JavaDoc <javadoc/index.html>`_
+* `RDoc <rdoc/_index.html>`_
+
+Developer guidelines
+---------------------
+
+* Guidelines for core developers (TBA)
+* Guidelines for plugin developers (TBA)
+
+"Embulk v2"
+-----------------------
+
+The Embulk project plans a set of big updates in the Embulk core so that Embulk, its community and the eco-system keep sustainable. Backward compatibility in plugins and configurations is considered especially carefully. Other parts, for example command line options, may have some incompatibility. Join discussions if you have concerns or requests!
+
+JFYI: "Embulk v2" is just a code name. The next versions will be still ``0.?``.
+
+Summary for plugin developers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+(TBA)
+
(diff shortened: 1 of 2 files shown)


embulk-docs/src/index.rst
```diff
@@ -75,14 +75,18 @@ Documents
 
    customization
```
muga: Since the "Customization" section is for developers' topic, it's better to move it in "For developers" section, isn't it?

embulk-docs/src/developers/index.rst
```diff
@@ -0,0 +1,38 @@
...
+
+* Guidelines for core developers (TBA)
+* Guidelines for plugin developers (TBA)
+
+"Embulk v2"
+-----------------------
+
+The Embulk project plans a set of big updates in the Embulk core so that Embulk, its community and the eco-system keep sustainable. Backward compatibility in plugins and configurations is considered especially carefully. Other parts, for example command line options, may have some incompatibility. Join discussions if you have concerns or requests!
+
+JFYI: "Embulk v2" is just a code name. The next versions will be still ``0.?``.
```
muga: Agree. Next major version v0.9 is good for Embulk v2 but, we already have v0.9 planning. I think that we need to cleanup/integrate both of them to avoid any confusing. 
https://github.com/embulk/embulk/milestone/1

dmikurube: Agreed. I just started to categorize and tag the issues, and am going to look into the milestone.

embulk-docs/src/index.rst
dmikurube: Good catch. Will do!.

dmikurube:
Since v2 will be a kind of big changes, and we don't have a big set of documents for the time being, I suggest to have v2 core/plugin documents under a separate top-level section. I thought it could be nice to have other (non-v2) documents separated to core and plugins, but I'm afraid we don't have so much documents to separate...

What do you think reconstructing in parallel to adding documents?

embulk-docs/src/index.rst
dmikurube: Moved.

muga:
@dmikurube Sorry for the delay. 

> What do you think reconstructing in parallel to adding documents?

Sounds good to me. Let's merge first and then I will think about it again.

dmikurube:
@muga Thanks! Merging this, and let's improve them later.

Issue #570, opened 2017-03-24
NPE when preview large amount of data


instcode:
When preview data from command line using: `embulk preview -Ilib load-shared_products.yml`, the NPE occurs:
```
2017-03-24 11:20:58.212 +0700: Embulk v0.8.16
2017-03-24 11:20:58.966 +0700 [INFO] (0001:preview): Loaded plugin embulk/input/app_annie from a load path
2017-03-24 11:20:59.028 +0700 [INFO] (0001:preview): Logging initialized @2342ms
java.lang.NullPointerException
	at org.embulk.spi.util.Pages.toObjects(org/embulk/spi/util/Pages.java:24)
	at org.embulk.command.PreviewPrinter.printAllPages(org/embulk/command/PreviewPrinter.java:36)
	at java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:497)
	at org.jruby.javasupport.JavaMethod.invokeDirectWithExceptionHandling(org/jruby/javasupport/JavaMethod.java:453)
	at org.jruby.javasupport.JavaMethod.invokeDirect(org/jruby/javasupport/JavaMethod.java:314)
	at RUBY.preview(~/.embulk/bin/embulk!/embulk/runner.rb:48)
	at RUBY.run(~/.embulk/bin/embulk!/embulk/command/embulk_run.rb:305)
	at RUBY.<main>(~/.embulk/bin/embulk!/embulk/command/embulk_main.rb:2)
	at org.jruby.Ruby.runInterpreter(org/jruby/Ruby.java:850)
	at org.jruby.Ruby.loadFile(org/jruby/Ruby.java:2976)
	at org.jruby.RubyKernel.requireCommon(org/jruby/RubyKernel.java:963)
	at org.jruby.RubyKernel.require(org/jruby/RubyKernel.java:956)
	at org.jruby.RubyKernel$INVOKER$s$1$0$require19.call(org/jruby/RubyKernel$INVOKER$s$1$0$require19.gen)
	at RUBY.(root)(uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1)
	at ~.$_dot_embulk.bin.embulk.embulk.command.embulk_bundle.invokeOther66:require(~/$_dot_embulk/bin/embulk/embulk/command/file:~/.embulk/bin/embulk!/embulk/command/embulk_bundle.rb:51)
[...]

PR #571, opened 2017-03-25
PreviewExecutor: Fix and avoid to call SamplingPageOutput#finish twice

muga:
This PR fixes https://github.com/embulk/embulk/issues/570. 

The cause of https://github.com/embulk/embulk/issues/570 is that SamplingPageOutput#finish method is called twice (unexpectedly). Null object is inserted into `pages` object by 1st call. 2nd call uses the null object. Then, we get NPE. 

I think that a plugin implementation causes this issue but, Embulk itself should have the way to avoid this situation. In this case, PreviewResult object should be reused.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java
+++ b/embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java
@@ -26,6 +26,7 @@ import org.embulk.spi.Exec;
 import org.embulk.spi.ExecSession;
 import org.embulk.spi.ExecAction;
 import org.embulk.spi.util.Filters;
+import org.slf4j.Logger;
 
 public class PreviewExecutor
 {
@@ -144,6 +145,7 @@ public class PreviewExecutor
     private static class SamplingPageOutput
             implements PageOutput
     {
+        private final Logger log = Exec.getLogger(this.getClass());
         private final int sampleRows;
         private final Schema schema;
         private List<Page> pages;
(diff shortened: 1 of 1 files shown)


embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java
```diff
@@ -174,10 +176,15 @@ public void add(Page page)
         @Override
         public void finish()
         {
+            if (res != null) {
+                // if PreviewResult object already exists, we should reuse it and throw PreviewedNoticeError.
```
dmikurube: Does this finally fail as a whole, or the previous `PreviewResult` is reused?

dmikurube:
@muga One question left.

embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java
muga: The previous `PreviewResult` is reused in the current fix. We also can add logging "Ignore that the finish method is called twice in preview" for plugin developers. How about this?

dmikurube: Just my impression is that explicitly failing may be better if `finish()` twice is not an expected behavior. If this is a normal executor, we may want to fallback. But, this is preview.

It's just my impression, though.

muga: @dmikurube I also thought, that's consistent behavior. Ok, I will revert it and just put logging that I mentioned above.

muga: Added logging on 7ac9423d4d84a073ed2cbd6e795b2dd5707e3bc0. Plugin developers might be able to know the place where `PageOutput#finish` is called twice in the plugin and how to fix it.

muga:
git rebase master before merging.

Issue #476, opened 2016-09-26
[Feature request] guess sample modes for large files.


hiroyuki-sato:
Related: https://github.com/embulk/embulk/pull/444

For a guess, some plugins need all of the data in a file. 
(ex. [embulk-parser-jsonpath](https://github.com/hiroyuki-sato/embulk-parser-jsonpath), Excel plugins..)

If an input file bigger than 32KB. The plugin raises an exception.

Ref: (Japanese text)
[Twitter #1](https://twitter.com/kompiro/status/779890193922789376)
[Twitter #2](https://twitter.com/kompiro/status/780198572188774401)

I think guess needs some modes. For example
- line: number of lines.  (Maybe fix this issue: https://github.com/embulk/embulk/issues/471)
- buffer: Now 32KB fixed size. It needs changeable.
- file: Instead of change buffer size 

It's already TODO in Source code. But an issue has not filed yet. 
https://github.com/embulk/embulk/blob/master/embulk-core/src/main/java/org/embulk/exec/SamplingParserPlugin.java#L93

hiroyuki-sato:
`embulk preview` which uses CSV or Jsonl data sometimes raises the exception.
Because of it can't read the end of the line when the data boundary reached. 

The following cases are the same issue. 
* #471 
* [Twitter](https://twitter.com/yassan168/status/819793357819940866) (Japanese)

hiroyuki-sato:
@muga 
You can reproduce this problem with this data.
https://gist.github.com/kakoni/a1dfaafb1de76f82df938a9ba4ba400e


```
embulk preview config.yml
2017-01-14 00:19:09.553 +0900: Embulk v0.8.15
2017-01-14 00:19:11.702 +0900 [INFO] (0001:preview): Listing local files at directory '.' filtering filename by prefix 'fail.csv'
2017-01-14 00:19:11.710 +0900 [INFO] (0001:preview): Loading files [fail.csv]
[...]

Issue #471, opened 2016-09-16
CSV input  Invalid record at line 38


kakoni:
Seeing "weird" exception;
```
org.embulk.spi.DataException: Invalid record at line 38..
 at org.embulk.standards.CsvParserPlugin.run(org/embulk/standards/CsvParserPlugin.java:368)
        at org.embulk.spi.FileInputRunner.run(org/embulk/spi/FileInputRunner.java:153)
        at org.embulk.exec.PreviewExecutor$2$1.run(org/embulk/exec/PreviewExecutor.java:122)
        at org.embulk.spi.util.Filters$RecursiveControl.transaction(org/embulk/spi/util/Filters.java:96)
        at org.embulk.spi.util.Filters$RecursiveControl$1.run(org/embulk/spi/util/Filters.java:92)
        at org.embulk.filter.column.ColumnFilterPlugin.transaction(org/embulk/filter/column/ColumnFilterPlugin.java:91)
```
which started to appear with embulk 0.8.10 (and higher). Same data set works perfectly with 0.8.9.

Note there is nothing wrong with the csv itself, can be processed normally with ruby csv etc.

hiroyuki-sato:
@kakoni

Hello,

Would you show me the sample data and configuration?

kakoni:
Sample conf + data in this gist; https://gist.github.com/kakoni/a1dfaafb1de76f82df938a9ba4ba400e

Running 
`embulk preview config.yml`
will produce
`org.embulk.spi.DataException: Invalid record at line 39: ` error

However it looks like `embulk run config.yml` runs ok

hiroyuki-sato:
For developers.

This issue reproduced on my environment.

This error raised with CsvTokenizer$TooFewColumnsException on this part.
https://github.com/embulk/embulk/blob/master/embulk-standards/src/main/java/org/embulk/standards/CsvParserPlugin.java#L364-L374

Maybe sample data exceed 32kb. But the last line corrected before the end of the row?

hiroyuki-sato:
@kakoni 

Embulk 0.8.19 released!. 
You can change preview size.
Please check this 
https://github.com/embulk/embulk/pull/572#issuecomment-291858787

If it works well, please close this issue.

hiroyuki-sato:
Hello, @kakoni 

What about this issue status?
Can you close this issue?

PR #572, opened 2017-03-25
PreviewExecutor: Make sampling buffer bytes configurable

muga:
This PR enables users configuring bytes of sampling buffer that `preview` command tries to read from input data source. Since users have various files and they sometimes want to change the # of rows that `preview` shows, it's better to configure the bytes by users. In the latest version of Embulk, v0.8.18, the bytes is 32KB. For example, if an user has a TSV file that includes one record line and the record bytes is over 32KB, `preview` will fail by `NoSampleError`. 

To do that, it introduces `preview_sample_buffer_bytes` option. Users can configure sampling buffer bytes in Embulk config `exec:` section as following.

```
exec:
  preview_sample_buffer_bytes: 65536
in:
  type: file
  ... ...
```


hiroyuki-sato:
Thank you for your great work!!.

IMO, A user may want to know current sample_size. 
So It is helpful to put it into the logger. 
If a user sees `32768 bytes` message, they want to change this value for a large file.
For example,

```
2017-03-15 21:11:55.609 +0900 [INFO] (0001:preview): Sampling the first 32768 bytes for preview...
```

Ref: An user request.
[#444 comment](https://github.com/embulk/embulk/issues/476#issuecomment-286402705)

muga:
Thank you. I added the logging.

hiroyuki-sato:
Thank you!. It must be helpful.

dmikurube:
@muga What do you think about naming this more explicit? For example, `preview_sample_buffer_bytes`.

It can be confusing with existing `preview_sample_rows`. The word "size" may sound like the number of rows.

muga:
@dmikurube Agree. I will rename with `preview_sample_buffer_bytes` that you mentioned. Once I finished writing tests, I will ask you to review this feature.

hiroyuki-sato:
In short, It worked perfectly!!  👍 
Thanks!!.

By the way,  Can I use this configuration for `guess`?
(I would like to use this configuration in guess too.)

I think it doesn't need `preview_` prefix. `sample_buffer_bytes` is better. 

## Prepare test.

See detail (#471)
Test data
https://gist.github.com/kakoni/a1dfaafb1de76f82df938a9ba4ba400e


```
mkdir preview_test
cd preview_test
wget https://gist.githubusercontent.com/kakoni/a1dfaafb1de76f82df938a9ba4ba400e/raw/04d12b4862f6667403fde592b8016376e146e64a/config.yml
wget https://gist.githubusercontent.com/kakoni/a1dfaafb1de76f82df938a9ba4ba400e/raw/04d12b4862f6667403fde592b8016376e146e64a/fail.csv
```


## latest version (embulk 0.8.18)

```
2017-04-05 22:06:39.119 +0900 [INFO] (0001:preview): Listing local files at directory '.' filtering filename by prefix 'fail.csv'
2017-04-05 22:06:39.123 +0900 [INFO] (0001:preview): Loading files [fail.csv]
org.embulk.spi.DataException: Invalid record at line 39: 0000000,0000000,0000000,0000000,000,00,0,0,0,0,0,0,0,0,000000000000,00000000000000,0,0000,000000000000,00000000000000,0,0,0,0,0,00000000000,0000000000000000000000000000000000000000,0000,000000,00000000,00,00000,0000,0,00000,0000,0,00,0000000000,000000,0000,000000,0000,000000,0000,000000,0000,000000000000000000000000000000000000000000000000000000000000,0,00000,00000,0000000000,0000,0000000000,0,000000,0000,0000,0000,0000,000000,0000,00000,0000,0,00000,000000000000000000000000000000000000000000000000000,00,00,00,00,00000000000000000000000000000000000,00,000,0,000,000,000,000,000,000,000,00,00,0,00000000,00000000,00,000,0,00000,0,0000000,0000000,0,0,0,0,00000000000000,00000000000000,00000000000000,00000,0000,00000,0000,000,0000000,0000,0,0,0,0,0,0,0,0,000,0000,0000000,0000000,00000,0000000,00000000,00000,00000000,0000,0000000
	at org.embulk.standards.CsvParserPlugin.run(org/embulk/standards/CsvParserPlugin.java:368)
	at org.embulk.spi.FileInputRunner.run(org/embulk/spi/FileInputRunner.java:153)
	at org.embulk.exec.PreviewExecutor$2$1.run(org/embulk/exec/PreviewExecutor.java:122)
```


## preview_add_preview_sample_size_in_exec_config branch(44738ea54fd2fd4ae079244ef5b46bdfdc99e907)

Add entry

```yaml
exec:
  preview_sample_buffer_bytes: 1048576 # 1MB 1024 * 1024
```

```
/path/to/embulk/pkg/embulk-0.8.18.jar preview -G config.yml
2017-04-05 22:10:33.201 +0900: Embulk v0.8.18
2017-04-05 22:10:34.213 +0900 [INFO] (0001:preview): Listing local files at directory '.' filtering filename by prefix 'fail.csv'
2017-04-05 22:10:34.216 +0900 [INFO] (0001:preview): Loading files [fail.csv]
2017-04-05 22:10:34.228 +0900 [INFO] (0001:preview): Try to read 1,048,576 bytes from input source
*************************** 1 ***************************
     a (string) : a
     b (string) : b
     c (string) : c
     d (string) : d
     e (string) : e
     f (string) : f
     g (string) : g
     h (string) : h
```

muga:
@hiroyuki-sato 

First, I'm now adding preview test framework on embulk-test. Because I don't want regression by this change. 

> Can I use this configuration for guess?
> (I would like to use this configuration in guess too.)

That's good idea 👍  I will create another issue ticket to do that.

> I think it doesn't need `preview_` prefix. `sample_buffer_bytes` is better.

I see. If we can use this option for `PreviewExecutor` and `GuessExecutor`, it would be even better I think. @dmikurube how about this?

hiroyuki-sato:
@muga Thank you for your comment.

I just wanted to try this feature.
And I'm glad it worked as I expected.
I'll learn how to write test codes.

dmikurube:
@muga In terms of naming `sample_buffer_bytes`, I basically agree with removing `preview_`, but we may need to consider consistency with existing `preview_sample_rows`?

muga:
> but we may need to consider consistency with existing preview_sample_rows?

@dmikurube That's right. Thank you. I will fix it.

muga:
Add tests by https://github.com/embulk/embulk/pull/601

muga:
Rebased to use #601 for testing this feature.

muga:
@dmikurube Please take a look this PR. I updated the PR's description as well.

I didn't remove `preview_` prefix from the option. We don't have a big benefit by sharing the option between `PreviewExecutor` and `GuessExecutor`. In the case of `preview_sample_rows` option, both of executors cannot share the option. because the meaning of "sample_rows" is different between both of executors.

muga:
I'm still working on it.

muga:
I noticed that `SamplingParserPlugin.runFileInputSampling` method uses `PreviewExecutorTask` and fixed it. The method is used by both of `preview` and `guess`.

embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java
```diff
@@ -36,6 +36,10 @@
     public interface PreviewTask
             extends Task
     {
+        @Config("exec")
+        @ConfigDefault("{}")
+        public ConfigSource getExecConfig();
```
dmikurube: Any reason why we use `exec` for this while `preview_sample_rows` is out of `exec`?

Sorry, I was not very sure how `preview_sample_rows` is configured. Is it out of `exec`? In other words, the configurations is like following?

```
preview_sample_rows: 50
exec:
  preview_sample_buffer_bytes: 10000
in:
  ...
out:
  ...
```

muga: Sorry, I'm also not sure why the `PreviewTask` has `preview_sample_rows` option. We need to ask @frsyuki about the reason if needed. But, I think, it's better that the option is moved within `exec:` section (or system config) as written in TODO comment.

muga:
Merging cc: @dmikurube

PR #578, opened 2017-03-28
Remove lib/embulk/command/embulk_generate_bin.rb that is not called anywhere.

dmikurube:
@muga Removing unnecessary code from `lib/embulk/command`.


dmikurube:
@muga This file looks just unnecessary. What do you think?

muga:
@dmikurube Agree. I think that we can safely remove this.

dmikurube:
@muga Thanks! Merging the PR (removing the file).

PR #582, opened 2017-03-29
Checkstyle the entire project with modified Google's Java Style Guide

dmikurube:
Nothing has been accepted nor determined yet, but just suggesting to apply checkstyle, and to change the coding style...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,6 +39,7 @@ ext {
 }
 
 configure(subprojects.findAll { ['embulk-core', 'embulk-standards', 'embulk-test'].contains(it.name) }) {
+    apply plugin: 'checkstyle'
     apply plugin: 'findbugs'
     apply plugin: 'jacoco'
     apply plugin: 'maven'
@@ -56,11 +57,27 @@ configure(subprojects.findAll { ['embulk-core', 'embulk-standards', 'embulk-test
         systemProperties System.properties.findAll { it.key.startsWith("org.embulk") }
     }
 
+    tasks.withType(Checkstyle) {
+        reports {
+            // Not to skip up-to-date checkstyles.
+            outputs.upToDateWhen { false }
+        }
+    }
+
     tasks.withType(FindBugs) {
         reports {
             xml.enabled = false
             html.enabled = true
         }
+    }
+
+    checkstyle {
+        toolVersion '8.7'
+        configFile = new File("${rootProject.projectDir}/config/checkstyle/checkstyle.xml")
+        configProperties = [
+            'checkstyle.config.path': new File("${projectDir}/config/checkstyle")
+        ]
+        ignoreFailures = false
     }
 
     findbugs {
--- a/build.gradle
+++ b/build.gradle
@@ -73,9 +73,9 @@ configure(subprojects.findAll { ['embulk-core', 'embulk-standards', 'embulk-test
 
     checkstyle {
         toolVersion '8.7'
-        configFile = new File("${rootProject.projectDir}/config/checkstyle/checkstyle.xml")
+        configFile = file("${rootProject.projectDir}/config/checkstyle/checkstyle.xml")
         configProperties = [
-            'checkstyle.config.path': new File("${projectDir}/config/checkstyle")
+            'checkstyle.config.path': file("${projectDir}/config/checkstyle")
         ]
         ignoreFailures = false
         maxWarnings = 0  // https://github.com/gradle/gradle/issues/881
--- a/build.gradle
+++ b/build.gradle
@@ -78,6 +78,7 @@ configure(subprojects.findAll { ['embulk-core', 'embulk-standards', 'embulk-test
             'checkstyle.config.path': new File("${projectDir}/config/checkstyle")
         ]
         ignoreFailures = false
+        maxWarnings = 0  // https://github.com/gradle/gradle/issues/881
     }
 
     findbugs {
--- a/config/checkstyle/README.md
+++ b/config/checkstyle/README.md
@@ -1,0 +1,11 @@
+Checkstyle for the Embulk project
+==================================
+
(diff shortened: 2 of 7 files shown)


hiroyuki-sato:
Will you change coding style Airlift to Google style?
(I'm not sure what is difference)

Current plugin coding style based on this file
https://github.com/airlift/codestyle/blob/master/IntelliJIdea14/Airlift.xml

Ref: [Embulkのソースコードスタイルチェック方法 (Java編)](http://qiita.com/hiroysato/items/584eafb4a94a961fa571)

dmikurube:
@hiroyuki-sato As mentioned, nothing has been accepted nor determined. Just remembered some of us talked about this, and I just tried. We'd want to have checkstyle in the Embulk core for some coding styles, anyway. (The core does not have checkstyle, for the time being!)

Coding styles are totally up to each plugin. This does not have any impacts on plugins. It is for the core source code, not for plugins.

hiroyuki-sato:
@dmikurube Thank you for the comment.

IMO

* It is a good idea to add automated code style setting into core. 
* And also, it is a good idea to use same coding style (core and **template**).

These are just past story, not related to future.

A user asked the original author about coding style. 
And it was Airlift. So I requested to add code style template for a plugin #361.
(Maybe, I was also better to try it into core at the same time)

And many users didn't know much about Airlift. 
(It was(still current) minor style.) :smile:

dmikurube:
> And also, it is a good idea to use same coding style (core and template).

We don't think they always need to be the same. For example, one idea is to provide options in `embulk new` to specify coding styles. (Maintaining multiple versions are annoying, though. It's an extreme viewpoint. ;) )

hiroyuki-sato:
> For example, one idea is to provide options in embulk new to specify coding styles. 

It's a nice idea!

I want to know recommend code style. (ex: default style).
I can't select it. :sweat_smile:

dmikurube:
I believe we don't have recommendations unless the plugin is merged into embulk-standards/. Options for auto-generation would be limited, though. It's totally up to plugin developers. :)

hiroyuki-sato:
The recommendation mean that If the core use "Airlift", I will use it for a plugin.
If the core uses another coding style, I'll use it for a plugin too. 

I don't have any policy about code style. I want to use the same code style. :smile:

Ref: [Twitter](https://twitter.com/hanpen_good/status/623266724763140096) (Japanese)

muga:
This is my thoughts.
1. The current Embulk doesn't have strict coding style. That's bad thing. Strictly speaking, we started development of Embulk by using "Airlift" coding style implicitly. But, some parts are applied and others are not applied. That's a biggest issue.
2. I thought that it's good to use same style as both of Embulk and Java plugins first but, now I think that we should not force plugin developers to use the style. Also think that the checkstyle.xml should be removed from build.gradle generated by `embulk new` command. 
3. Since the change diff is small instead of "Google" if we would take "Airlift" style, I prefer "Airlift". I don't have any other reasons that I advocate it but.
4. But, I'm OK to change the style from "Airlift" to "Google". Not big issue. Rather, #1 is big problem.

hiroyuki-sato:
If the `checkstyle.xml` remove from the `gradle new` command,
Some documents may helpful for new plugin developer. 

I just want to use the same coding style.

## The Embulk coding style.

Write a description; It is not important to use the same coding style.
But If you want to use the same coding style, please refer the following. 

* For IntelliJ (URL here)
* For gradle (XML file link here), example the following.

### checkstyle example.

```
plugins {
    // java
    id "checkstyle"
}

checkstyle {
  // ... 
}
```

dmikurube:
This ticket/PR does not care about plugins anyway.

IMHO, the coding style of the core does not represent any recommendation. Although it's free for plugin developers to follow the coding style of the code, the core does not (should not) take care.

Although it's worth discussing how `embulk new` should work, it's to be done in a separate ticket if needed.

dmikurube:
It's not urgent. It's not a topic to spend much time to discuss. Let's avoid bikeshed.

Topics to discuss here in this ticket are only:

1. Should we introduce `checkstyle` for the core code base?
2. If yes for 1, which style to introduce for the core code base?

* Plugins are out of focus. It's to be discussed per plugin.
* `embulk new` is out of focus of this ticket. It's to be discussed in another ticket.

hiroyuki-sato:
Sorry for the writing much about outside of the main topic. 
I have no opinion about the main topic.
I know a little about the ​Airlift style only.

muga:
@dmikurube  Thank you for simplifying.

For 1, it's better to introduce checkstyle to perform code quality. But, I also think that it's good to postpone to do that for now, if we need to discuss for 2 more:-) 

For 2, I prefer "Airlift" because the diff is small if we would take "Airlift" style instead. But, I don't have any other reasons that I advocate it.

dmikurube:
My opinions:

1. Yeah, we should have "some".
2. I thought Google's + following modifications may fit Embulk well:
    * Indent: 2 ==> 4
    * Columns: 100 ==> 120 or more
    * Package names: allow `_`

The reasons of 2 are:

* As mentioned, the "Airlift" style is not very popular. :( To make Embulk more friendly for open-source developers, we may want to use more popular ones.
* Updating all existing indents is hard, and breaks all Git histories. Keeping 4 is reasonable.
* 100 columns are too short for 4-space indents.

----

Once we introduce some stylechecks, we anyway need to update existing code eventually. I'd suggest to utilize suppressions to update the existing code gradually. 

* Add checkstyles with all existing Java files suppressed.
* Remove good files from suppressions.
* Do not add new Java files in suppressions.
* Update existing Java files with cleaning-up and removing from suppressions when we have chances.

This process can be easily applied even for Airlift's and Google's.

dmikurube:
@muga @sakama Adding checkstyle, finally. Can you have a look?

This PR does not check any file for the time being. Next #914 will enable checkstyle for embulk-core's main files.

dmikurube:
Thanks... Merging, finally!

PR #583, opened 2017-03-29
Update recipe. Use elasticsearch and kibana5

hiroyuki-sato:
I tested on the macOS. 
Maybe It will work well on the Ubuntu. 

@sakama Please take a look.


hiroyuki-sato:
Thank you!.

PR #591, opened 2017-04-04
Add updateJRuby task document

hiroyuki-sato:
I've forgotten usage `updateJRuby` task.
I update the README. 
Could you add it to README?

By the way, JRuby 9.1.8.0 released at 6/Mar/2017
http://jruby.org/2017/03/06/jruby-9-1-8-0.html
It fixed many issues.


muga:
Merged.

PR #592, opened 2017-04-05
[WIP] Load pure-Java JAR-based plugins.

dmikurube:
@muga It's just a PoC, not to be merged yet. Let me just share the status atm.

Confirmed that a pure-Java (non-Gem) plugin can be a kind loaded.

1. Build https://github.com/dmikurube/embulk-input-dummyjar with `./gradlew pluginJar`
2. Build Embulk with this PR.
3. Write a simple YAML config like:
```
in:
  type: jar:/full/path/to/your/built/embulk-input-dummyjar-plugin-0.0.1.jar
out:
  type: stdout
```
4. `pkg/embulk-0.8.18.jar run dummyjar.yml` with Embulk built at 2.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/EmbulkService.java
+++ b/embulk-core/src/main/java/org/embulk/EmbulkService.java
@@ -9,6 +9,7 @@ import org.embulk.config.ConfigSource;
 import org.embulk.exec.SystemConfigModule;
 import org.embulk.exec.ExecModule;
 import org.embulk.exec.ExtensionServiceLoaderModule;
+import org.embulk.jar.JarPluginSourceModule;
 import org.embulk.plugin.PluginClassLoaderModule;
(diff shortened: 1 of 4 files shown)


muga:
This is great milestone 👍  When you can, please create design doc to next step and cover the following topics?
- How can we manage the order of plugin source to load plugins?
- What should Embulk users declare in Embulk config and system config?
- How do we define the "type:" format?
- How can we manage versions of the plugin? and the dependencies as well.

What to do for this PR? I think that it's OK for me to close it because the PR fulfilled the role. If you want to merge it into master, it's better to discuss about the `type:` format and rule a little bit.

dmikurube:
@muga This was really just a PoC. I confirmed it is feasible to load a JAR as a plugin, and yes, I will start designing details based on that with design docs.

Closing this PR.

Issue #596, opened 2017-04-11
Use MANIFEST instead of embulk.properties?


dmikurube:
We've implemented #560 to embed Embulk versions in `embulk.properties`, but I thought we should have used JAR manifest instead of our own original properties file.

Will update the decision, but let me just share for now... Ping me if we have plans to release new Embulk versions soon. @muga

muga:
> I thought we should have used JAR manifest instead of our own original properties file.

Sounds good to me. Let's replace with MANIFEST.MF. About next release, I can wait one week at least.

dmikurube:
Done.

PR #597, opened 2017-04-12
Use the jar manifest (META-INF/MANIFEST.MF) to provide the Embulk version at runtime [#596]

dmikurube:
@muga Per #596, removed `embulk.properties` and started to use `Implementation-Version` in the jar manifest. Can you PTAL?

(Will update other PRs depending on `embulk.properties`.)

### DIFF
--- a/embulk-core/src/main/java/org/embulk/EmbulkVersion.java
+++ b/embulk-core/src/main/java/org/embulk/EmbulkVersion.java
@@ -59,7 +59,6 @@ public final class EmbulkVersion
             }
 
             try (final JarFile selfJarFile = new JarFile(selfJarPathString)) {
-                final Manifest selfJarManifest;
                 try {
                     return selfJarFile.getManifest();
                 }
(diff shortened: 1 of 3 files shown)


embulk-core/src/main/java/org/embulk/EmbulkVersion.java
```diff
@@ -1,26 +1,109 @@
...
+                System.err.println("Embulk version unavailable due to the path of CodeSource unavailable.");
+                return null;
+            }
+            else if (selfJarPathString.isEmpty()) {
+                System.err.println("Embulk version unavailable due to the path of CodeSource empty.");
+                return null;
+            }
+
+            try (final JarFile selfJarFile = new JarFile(selfJarPathString)) {
+                final Manifest selfJarManifest;
```
muga: Is this local variable used?

dmikurube: Ah, unused. Removed it, and will be merging!

dmikurube:
Merged!

PR #598, opened 2017-04-13
Avoid saving unused string/Json values to stringReferences/valueReferenes in PageBuilder

muga:
### Problem

The current logic for saving string column values in `PageBuilder` often causes OOM. string values are stored on `stringReferences` field in `PageBuilder` class. The `stringReferences` is not cleared if `PageBuilder#addRecord` method is not called in input plugins by some reasons. Rather, unused string values continue to be stored in the `stringReferences` and the # keeps growing.

For example, when CsvParserPlugin parses and detects invalid lines from CSV files, it doesn't call `#addRecord` method to skip the invalid lines. They sometimes continue to not be able to get/parse invalid data. The `stringReferences` is growing by unused string values and then, OOM might happen.

This problem happens for Json `valueReferences` as well. 

### Solution

To avoid saving unused string and Json values in `stringReferences` and `valueReferences`, we introduce `Row` class, which is a temporal row container to save one row only. `PageBuilder` has it as a field.

Input plugins use `PageBuilder#setXXX` methods to write column values to `Page` objects. The `#setXXX` methods call `Row#setXXX` methods and save column values temporarily in the `Row` object. Once plugins call `PageBuilder#addRecord`, the method calls `Row#write(PageBuilder)`. The `#write` method saves column values in Page, in particular saves string and Json values to `stringReferences` and `valueReferences` in `PageBuilder`. Even though input plugins continue to call `PageBuilder#setXXX` methods without `#addRecord` call method, `Row` object is only updated. Not stored unused string and Json values in `stringReferences` and `valueReferences`.

Additionally I changed the type of `stringReferences` from `BiMap` to `List`. Not pretty sure but I don't think that the `BiMap` has effect as a cache. 3443651cdec6a18faccffecfc06f0220c6a6481a.


embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
```diff
@@ -170,49 +146,76 @@ public void setTimestamp(int columnIndex, Timestamp value)
...
         }
+        else {
+            row.setTimestamp(columnIndex, value);
+        }
+    }
 
-        int offset = getOffset(columnIndex);
-        bufferSlice.setLong(offset, value.getEpochSecond());
-        bufferSlice.setInt(offset + 8, value.getNano());
+    void writeNull(int columnIndex)
```
dmikurube: Any reasons why they are not `private`?

It's okay named `write*` if it is just `private`. But if the methods are somewhat visible, I'm concerned that the difference is not clear between `set*` and `write*`. Please make them `private`, or rename them so that their names represent what they do and/or comment why they are not `private`.

embulk-core/src/main/java/org/embulk/spi/Row.java
```diff
@@ -0,0 +1,318 @@
...
+
+import org.embulk.spi.time.Timestamp;
+import org.embulk.spi.type.Type;
+import org.embulk.spi.type.Types;
+import org.msgpack.value.Value;
+
+/**
+ * This class is a temporal container for storing a row only.
+ */
+class Row
```
dmikurube: Can't it be an inner private class of `PageBuilder`?

embulk-core/src/main/java/org/embulk/spi/Row.java
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    void write(PageBuilder pageBuilder)
+    {
+        for (Val v : values) {
+            v.write(pageBuilder);
+        }
+    }
+
+    interface Val
```
dmikurube: Let's use `Value` instead of `Val`. I don't think we achieve any benefit by this abbreviation.

embulk-core/src/main/java/org/embulk/spi/Row.java
```diff
@@ -0,0 +1,318 @@
+package org.embulk.spi;
+
+import org.embulk.spi.time.Timestamp;
+import org.embulk.spi.type.Type;
+import org.embulk.spi.type.Types;
+import org.msgpack.value.Value;
+
+/**
+ * This class is a temporal container for storing a row only.
```
dmikurube: IIUC reading the code, the purpose of this class `Row` is not to double-store values for the same columns, right?

If so, I couldn't get the point from this comment and the PR description. What do you think to comment:
```
/**
 * Row is a container to stage values before adding into reference lists such as |stringReferences|.
 *
 * |Row| works as a buffer against plugins that may add values incorrectly without |PageBuilder#addRecord|.
 * It accepts just one value per column while |PageBuilder| can double-store values regardless of columns.
 * Double-stored values are overwritten.
 */
```

embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
muga: Agree. I changed those modifiers with `private` since `Row` class was moved as static inner class of `PageBuilder`. d0aa12f58af67940cea6d4b39f47568473a178c6

embulk-core/src/main/java/org/embulk/spi/Row.java
muga: Thank you. Fixed d0aa12f58af67940cea6d4b39f47568473a178c6

embulk-core/src/main/java/org/embulk/spi/Row.java
muga: Fixed d0aa12f58af67940cea6d4b39f47568473a178c6 thank you.

embulk-core/src/main/java/org/embulk/spi/Row.java
muga: I also thought that but, I didn't want to confuse this with `org.msgpack.value.Value`. 

I changed this with `ColumnValue`. How about this? fd401b3a9a39d22a48e2b11e6023defdb4d377e8

muga:
@dmikurube Thank you for reviewing. I fixed them and left several comments. Please take a look again.

embulk-core/src/main/java/org/embulk/spi/Row.java
dmikurube: Looks good. :)

embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
dmikurube: They are still non-modifiered?

embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
```diff
@@ -270,6 +275,322 @@ public void close()
...
+
+        private void write(PageBuilder pageBuilder)
+        {
+            for (ColumnValue v : values) {
+                v.write(pageBuilder);
+            }
+        }
+    }
+
+    interface ColumnValue
```
dmikurube: At the same time with the above comment, nicer to specify modifiers to these `interface` and `class`.

embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
muga: Sorry.. Fixed 4ff5d43d1a23c8e24a1b1dd9e1ee69c9246fdad7

embulk-core/src/main/java/org/embulk/spi/PageBuilder.java
muga: Thank you. Fixed 4ff5d43d1a23c8e24a1b1dd9e1ee69c9246fdad7

muga:
@dmikurube Good point. Thank you. 

I added unit tests for PageBuilder.Row. Then, I will merge this into master. Please let me know if you have any concerns about it.

dmikurube:
Although I still think it looks fine, but I meant to check existing test cases are enough to confirm normal behaviors, not to check the new double-counting behavior.

muga:
My explanation was bad. I added new tests because existant tests are not enough.

PR #602, opened 2017-04-16
embulk-test: Add InputBuilder#preview

muga:
This PR introduces a testing framework to check `preview` behavior of input plugins and add it to embulk-test. embulk-test already has testing frameworks for `guess` and `run` behavior of input plugins but, it doesn't support for `preview`. 

The role of This testing framework on the PR is to compare actual input data and expected data. The actual input data is defined by users. The framework receives `in:` config from users in advance. The expected data is in Page objects that the `PreviewResult` has.

The behavior of the framework is:
1. Receives `in:` config from users (testers) in advance.
2. Executes `preview` command first and get the `PreviewResult`. 
3. Executes `run` command. The Embulk ingests Page objects from `PreviewResult`. Then, the Page objects are output to temporal files.


embulk-test/src/main/java/org/embulk/test/TestingBulkLoader.java
```diff
@@ -33,13 +44,62 @@
...
                 };
                 return ImmutableList.of(Modules.override(modules).with(ImmutableList.of(override)));
             }
         };
     }
 
+    /**
+     * This plugin is used for TestingEmbulk.InputBuilder.preview().
+     */
+    static final class PreviewResultInputPlugin
```
dmikurube: Can this class be `private`, or if used from outside, can we move this class `public` out of `TestingBulkLoader`?

muga: I moved it to out of `TestingBulkLoader`. 2f386a0c9eef56a39f4c4d84575e843babb26402

muga:
@dmikurube Thank you for reviewing. I will merge it into master once the CI passed.

Issue #600, opened 2017-04-13
Update JRuby 9.1.8.0


muga:
@hiroyuki-sato mentioned upgrading JRuby on https://github.com/embulk/embulk/pull/591

> By the way, JRuby 9.1.8.0 released at 6/Mar/2017
> http://jruby.org/2017/03/06/jruby-9-1-8-0.html
> It fixed many issues.

hiroyuki-sato:
The version 9.1.8.0 is Ruby 2.3.1 compatible. 
It is the same 9.1.5.0 which is used in Embulk 0.8.18.

It may don't break compatibility like the
Embulk 0.8.9 to 0.8.10.
http://www.embulk.org/docs/release/release-0.8.10.html
It updated Ruby 2.2.0 to 2.3.0.

```
[1] pry(main)> JRUBY_VERSION
=> "9.1.8.0"
[2] pry(main)> RUBY_VERSION
=> "2.3.1"
```

hiroyuki-sato:
Just note.

@sonots hit this issue on the embulk [0.8.6](http://www.embulk.org/docs/release/release-0.8.6.html). 
https://github.com/jruby/jruby/issues/3670

He forked and downgraded JRuby to 9.0.4.0 for a while. 

https://twitter.com/sonots/status/710759273312030720
https://twitter.com/sonots/status/710763938074333184

muga:
Thank you for sharing that. According to https://github.com/jruby/jruby/issues/3670, it seems that this issue was fixed on v9.1.0.0. Embulk v0.8.18 uses JRuby v9.1.5.0 so, v9.1.5.0 should be ok for https://github.com/jruby/jruby/issues/3670. 

I will care of this after Embulk v0.8.19.

hiroyuki-sato:
@muga thank you for the comment.

I just wanted to note. 
The update will cause another issue just in case, so we need to update it carefully. 
(ex. sonots case. Yes that already fixed.)

Almost same meaning this tweet by muga 
[Twitter](https://twitter.com/muga_nishizawa/status/852791624170053632)

dmikurube:
JRuby has already been upgraded to 9.1.13.0, and 9.1.15.0 in v0.9.0.

PR #604, opened 2017-04-19
Make environment variables immutable in JRuby/Java.

dmikurube:
@muga WIP yet, but let me share. (.bat needs to be fixed.)

### DIFF
--- a/embulk-cli/src/main/bat/selfrun.bat
+++ b/embulk-cli/src/main/bat/selfrun.bat
@@ -30,37 +30,59 @@ if "%overwrite_optimize%" == "true" (
     )
 )
 
+setlocal enabledelayedexpansion
+
+set found_bundle_option=0
+
 for %%a in (%*) do (
-    if %%a == "-b" (
-        set found_bundle_option=true
-    ) else if %%a == "--bundle" (
-        set found_bundle_option=true
-    ) else if %found_bundle_option% == "true"
-        set EMBULK_BUNDLE_PATH=%%a
-        set found_bundle_option=
+    if %%a == -b (
+        set found_bundle_option=1
+    ) else if %%a == --bundle (
+        set found_bundle_option=1
+    ) else if !found_bundle_option! == 1 (
+        set embulk_bundle_path=%%a
+        set found_bundle_option=2
     )
 )
 
+endlocal && set EMBULK_BUNDLE_PATH=%embulk_bundle_path%
+
 if not defined EMBULK_BUNDLE_PATH (
     set EMBULK_BUNDLE_PATH=
-    set BUNDLE_GEMFILE=
-    FOR /F usebackq IN (`java -cp %0 org.jruby.Main -e 'print RbConfig::CONFIG["ruby_version"]'`) DO SET rb_version=%%w
-    set GEM_HOME="%USERPROFILE%/.embulk/jruby/%rb_version%"
     set GEM_PATH=""
 ) else (
     if not exist "%EMBULK_BUNDLE_PATH%\" (
         echo Directory not found: "%EMBULK_BUNDLE_PATH%"
         exit /b 1
     )
-    call :absolute_path %EMBULK_BUNDLE_PATH%
-    set BUNDLE_GEMFILE="%absolute_path%/Gemfile"
-    if not exist "%BUNDLE_GEMFILE%" (
-        echo Gemfile not found: "%BUNDLE_GEMFILE%"
-        exit /b 1
-    )
-    set GEM_HOME=
     set GEM_PATH=
 )
+
+setlocal enabledelayedexpansion
+
+if not defined EMBULK_BUNDLE_PATH (
+    set bundle_gemfile=
+) else (
+    call :get_absolute_path %EMBULK_BUNDLE_PATH%
+    set bundle_gemfile=!absolute_path!\Gemfile
+    if not exist !bundle_gemfile! (
+        echo Gemfile not found: "!bundle_gemfile!"
+        exit /b 1
+    )
+)
+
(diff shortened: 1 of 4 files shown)


dmikurube:
Hmm, failing in tests... Will check.

embulk-cli/src/main/sh/selfrun.sh
```diff
@@ -44,11 +44,38 @@ while true; do
...
+        *)
+            shift
+            ;;
+    esac
+done
+
+if test -z ${EMBULK_BUNDLE_PATH}; then
+    unset EMBULK_BUNDLE_PATH
+    unset BUNDLE_GEMFILE
+    export GEM_HOME="$(cd && pwd)/.embulk/jruby/$(java -cp $0 org.jruby.Main -e 'print RbConfig::CONFIG["ruby_version"]')"
```
hiroyuki-sato: Let's use backquote instead of `$()`.
The`$()` is `bash` extension. 
Some OSes doesn't use `bash` as `/bin/sh` 
(As far as I remember, debian, ubuntu, xxBSD).

Maybe you can write `$HOME` instead of `$(cd && pwd)`.

```bash
GEM_HOME="$HOME/.embulk/jruby/`java -cp $0 org.jruby.Main -e 'print RbConfig::CONFIG["ruby_version"]'`"
```

dmikurube: That may be worth discussing, but `$()` is now POSIX-compliant. Recent `sh` like `ash` has no problems with that. Although the really legacy `sh` may have problems with `$()`, I believe we have no problems with recent `sh` (not `bash`). It may be still worth choosing the legacy option, though.
http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03

The reason of `cd && pwd` was that it was originally `java.lang.System.properties["user.home"] || ENV['HOME']` in JRuby, which might be to support irregular cases like Windows. I thought it may be fine to use `$HOME` since this is the `.sh` part (non-Windows), but I was not very sure about how it works with Bash on Windows or else like that.

Will fix the `export` one.

I guess `unset` for environment variables is fine in `sh`.

hiroyuki-sato: Oh. I didn't know `$()` POSIX-compliant. 
I tested some commands on FreeBSD 10.
It seems that `$()` and `unset` work well.

I'm not familiar bash on Windows too. 

```bash
$ A=`pwd`
$ echo $A
/home/vagrant
$ B=$( pwd )
$ echo $B
/home/vagrant
$ unset B
$ echo $B

$
```

hiroyuki-sato: @dmikurube FYI
bash on Windows 
* `$HOME` point to `/home/<username>`.
* It's not the same `%HOMEPATH%`
* `%HOMEPATH%` localated at `/mnt/c/Users/<username>`  in bash.

Ref: [Twitter](https://twitter.com/koturn/status/854719055986515968) (Japanese text).

dmikurube: Thanks.

Chose the safer way (with backticks) and fixed `export`. Kept `cd && pwd` and `unset` for the time being.

dmikurube:
Hmm, I need a Windows environment...

dmikurube:
Fixed the Windows bat file finally.

@muga Can you have a look?

embulk-cli/src/main/sh/selfrun.sh
```diff
@@ -89,5 +89,5 @@ else
     java_args="-XX:+AggressiveOpts -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xverify:none $java_args"
 fi
 
-exec java $java_args -jar "$0" $jruby_args "$embulk_args"
+exec java $java_args -jar "$0" $jruby_args $embulk_args
```
muga: minor: Does `"$0"` still need `"` quote character? If we could remove `"` from `"$embulk_args"`, I thought that we can remove `"` from `"$0"`.

embulk-cli/src/main/bat/selfrun.bat
```diff
@@ -129,6 +151,6 @@ if not exist "%~1" (
 set status=
 exit /b
 
-:absolute_path
-set absolute_path=%~dp1
+:get_absolute_path
+set absolute_path=%~f1
```
muga: minor: Please tell me Why you did change `%~dp1` with `%~f1`?

embulk-cli/src/main/sh/selfrun.sh
dmikurube: The reason why I removed `"` from `$0` is just because it is already quoted at https://github.com/embulk/embulk/pull/604/files#diff-73174f18d3b0fc9c849344f928ca50a0R47

```
embulk_args="$@"
```

embulk-cli/src/main/bat/selfrun.bat
dmikurube: I was just misunderstanding. `f` was what I needed and `dp` was just wrong.

dmikurube:
@muga If you have no concerns with my comments above, I'd merge this PR.

dmikurube:
This will be after #600.

dmikurube:
@muga v0.8.22 has been released. Is it good time to merge this?

muga:
@dmikurube Agree. Let's merge 👍

dmikurube:
@muga I've found that `bin/embulk` needs updating as well. Can you have another look?

muga:
@dmikurube LGTM. good catch..

dmikurube:
Thanks. Merging this toward 0.8.24...

PR #607, opened 2017-04-21
[WIP;PoC] Apply security policy by plugin's own manifest.

dmikurube:
This is just Proof-of-Concept as well as #592, not for merging. (And it depends on #592 as well.)

It's just a trial to apply a kind of Security Policy for plugin. This is manifest-based, which means that a plugin declares its required permissions by itself. It may make it easier to confirm plugins don't behave out of expectations.

Cc: @muga 

It has problems yet. The biggest one is that the main code (not plugin code) may run under `PluginClassLoader` in case of `ScatterExecutor` or like that.

### DIFF
--- a/embulk-cli/src/main/java/org/embulk/cli/Main.java
+++ b/embulk-cli/src/main/java/org/embulk/cli/Main.java
@@ -1,11 +1,38 @@
 package org.embulk.cli;
 
 import java.net.URISyntaxException;
+import java.security.AllPermission;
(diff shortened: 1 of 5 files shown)


dmikurube:
`embulk-input-dummyjar` is still its example. Changes: https://github.com/dmikurube/embulk-input-dummyjar/commit/6a9e756e506e6b4e749216a3db7df0944dc40c8d

PR #608, opened 2017-04-22
Java timestamp parser and RubyDateParser

muga:
To improve the performance of parsing timestamp values, this PR introduces Java timestamp parser. It implements `RubyDateParser`, which is compatible with Ruby's strptime parser. The parser parses timestamp values by strptime tokens pre-tokenized by strptime lexer.


embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java
```diff
@@ -62,15 +73,19 @@
...
+    private final Calendar cal;
+    private final List<StrftimeToken> compiledPattern;
 
     @Deprecated
     public TimestampParser(String format, ParserTask task)
     {
         this(task.getJRuby(), format, task.getDefaultTimeZone());
     }
 
+    @VisibleForTesting
```
dmikurube: I don't recommend this kind of usage of `@VisibleForTesting`.

If this is used by some main code, we don't need this.

If visibility is needed only for testing, I'd suggest to make a `static` creator method whose name clearly describes it is for testing. For example, `@VisibleForTesting TImestampParser createForTesting(...)`.

settings.gradle
```diff
@@ -4,4 +4,5 @@ include 'embulk-core'
 include 'embulk-standards'
 include 'embulk-cli'
 include 'embulk-test'
+include 'embulk-jruby-strptime'
```
dmikurube: Do we use this name `-strptime` while it contains `strftime` as well?

embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java
```diff
@@ -62,15 +73,19 @@
         public Optional<String> getDate();
     }
 
-    private final JRubyTimeParserHelper helper;
     private final DateTimeZone defaultTimeZone;
+    private final String format;
+    private final RubyDateParser parser;
+    private final Calendar cal;
```
dmikurube: nitpick: how about using full-spell?

embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java
```diff
@@ -91,7 +106,12 @@ public TimestampParser(ScriptingContainer jruby, String format, DateTimeZone def
 
     public TimestampParser(ScriptingContainer jruby, String format, DateTimeZone defaultTimeZone, String defaultDate)
     {
-        JRubyTimeParserHelperFactory helperFactory = (JRubyTimeParserHelperFactory) jruby.runScriptlet("Embulk::Java::TimeParserHelper::Factory.new");
+        // TODO get default current time from ExecTask.getExecTimestamp
+        Ruby runtime = jruby.getProvider().getRuntime();
```
dmikurube: Do we have plans to eliminate this JRuby instance eventually (may not be in this PR)?

embulk-jruby-strptime/src/main/java/org/jruby/lexer/StrptimeLexer.java
```diff
@@ -0,0 +1,603 @@
+/* The following code was generated by JFlex 1.4 on 4/21/17 2:55 PM */
```
dmikurube: Do we merge this generated file in the repository?

embulk-jruby-strptime/src/main/java/org/jruby/util/RubyDateParser.java
```diff
@@ -0,0 +1,977 @@
+package org.jruby.util;
```
dmikurube: Do we have these classes under `org.jruby.util`?

embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java
muga: OK, thank you. Fixed 768220f11c4ef99994769cb0f59f1664512d06ab

embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java
muga: Agree. I fixed with `createTimestampParserForTesting` 768220f11c4ef99994769cb0f59f1664512d06ab

embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java
muga: Correct. Yes, I will try to do that on another PR.

settings.gradle
muga: This doesn't include `strftime` but, I renamed several classes. `StrftimeToken` to `StrptimeToken`, `StrftimeFormat` to `StrptimeFormat`. a4b40869b17eb29dfaa01dfb72998f4a45dae188

embulk-jruby-strptime/src/main/java/org/jruby/lexer/StrptimeLexer.java
muga: We don't need to put the .java file. I will take a look how to generate this file later. This file is generated with JFlex 1.4 that JRuby uses. Since the version is old, it's not supported by several JFlex gradle plugin but.

embulk-jruby-strptime/src/main/java/org/jruby/util/RubyDateParser.java
muga: Moved them to `org.embulk.spi.time`. c4bec16b0ca8fcb26b4b86ae58c57e61cf0db6fb

muga:
This PR will be closed and not merged. We're working for this feature on https://github.com/embulk/embulk/pull/611.

muga:
Close this ticket.

PR #609, opened 2017-04-30
Add 'guess_ sample_ buffer_ bytes' option for GuessExecutor

muga:
This PR enables users to configure bytes of sampling buffers that `guess` can read. To improve the accuracy of the config suggestion by `guess` command, it's better to configure the bytes by users. In the latest version of Embulk, v0.8.18, the bytes is 32KB.

It introduces `guess_sample_buffer_bytes` option. Users can specify it in `exec:` section of Embulk config as following. Their `guess` command can configure sampling buffer bytes.

```
exec:
  guess_sample_buffer_bytes: 65536
in:
  type: file
  ... ...
```

Instead of this PR, we also working on 'preview_sample_buffer_bytes' option on https://github.com/embulk/embulk/pull/572. This assumes https://github.com/embulk/embulk/pull/572.


muga:
@dmikurube Ok, thank you for reviewing. As I merge this, I will update built-in plugin document.

PR #611, opened 2017-05-07
Java timestamp parser and RubyDateParser v2

muga:
To improve the performance of parsing timestamp values, this PR introduces Java timestamp parser. It implements RubyDateParser, which is compatible with Ruby's strptime parser. The parser parses timestamp values by strptime tokens pre-tokenized by strptime lexer.

This PR is a replacement of #608. #608 will be closed.


dmikurube:
@muga I understand this #611 is to be reviewed and merged, and #608 is not to merge. Ping me once it's ready (with the fix of `FormatBag` which we discussed).

muga:
@dmikurube I finished refactoring `FormatBag` to rename those fields declared in `FormatBag` class and add setter/getter for them. Sorry, this PR is huge change as you know but, please take a look when you have time.

dmikurube:
@muga Just a question, is this PR intended to be merged into `master` directly, or after other PR(s)?

muga:
@dmikurube This PR is expected to be merged into `master` directly.  #608 will be closed (not merged) but, just in case I keep opening it. Please let me know if you feel or have comfortable ways to review this feature.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeFormat.java
```diff
@@ -0,0 +1,44 @@
+package org.embulk.spi.time;
+
+// Ported from org.jruby.util.RubyDateFormatter.Format in JRuby
```
dmikurube: If it's a kind of copy, is licensing okay? We may want/need some notations on licensing.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeToken.java
```diff
@@ -0,0 +1,100 @@
+package org.embulk.spi.time;
+
+import static org.embulk.spi.time.StrptimeFormat.*;
+
+// Ported from org.jruby.util.RubyDateFormatter.Token in JRuby
```
dmikurube: Licensing as well.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParser.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.HashMap;
+import java.util.List;
+
+import static org.embulk.spi.time.StrptimeParser.FormatBag.has;
+import static org.jruby.RubyRational.newRationalCanonicalize;
+
+/**
+ * This class has {@code StrptimeParser} and provides methods that are calls from JRuby.
+ */
+public class RubyDateParser
```
dmikurube: IIUC, this `RubyDateParser` is not called within Embulk?

If so, and we don't need this immediately, what do you think delaying to merge this?

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParse.java
```diff
@@ -0,0 +1,283 @@
...
+        map.put("us eastern",             -18000);
+        map.put("us mountain",            -25200);
+        map.put("vladivostok",             36000);
+        map.put("w. australia",            28800);
+        map.put("w. central africa",        3600);
+        map.put("w. europe",                3600);
+        map.put("west asia",               18000);
+        map.put("west pacific",            36000);
+        map.put("yakutsk",                 32400);
+        ZONES_SOURCE = new HashMap<>(map);
```
kares: how about a plain-old `switch() case` (which is actually new for strings since Java 7) - no need for boxing or maps :)

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParser.java
```diff
@@ -0,0 +1,112 @@
...
+{
+    private final ThreadContext context;
```
kares: do not see the `RubyDateParser` being used here ... but ideally context should not be kept around
... or only if its really used for a short period and than released (no instances kept around)

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParser.java
```diff
@@ -0,0 +1,112 @@
+package org.embulk.spi.time;
+
+import org.jruby.runtime.ThreadContext;
```
muga: I added TODO comments on 1006914f22a1343a60cb76c102a0a6aaeca6dac3. We still need JRuby-dependency tentatively but, it will be removed after merging embulk-jruby-strptime into JRuby.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParser.java
muga: We still need this `RubyDateParser` because this is required for executing `test/mri/date/test_date_strptime.rb` with new `StrptimeParser`.

embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java
```diff
@@ -1,18 +1,26 @@
 package org.embulk.spi.time;
 
+import java.text.SimpleDateFormat;
+import java.text.ParseException;
```
muga: Fixed 536f36d52cbf851014b86c5fc9a98e4985ae9dd4. Thank you.

embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java
```diff
@@ -1,18 +1,26 @@
...
 import com.google.common.base.Optional;
-import org.jruby.embed.ScriptingContainer;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 import org.embulk.config.Config;
 import org.embulk.config.ConfigInject;
 import org.embulk.config.ConfigDefault;
 import org.embulk.config.ConfigException;
+import org.jruby.embed.ScriptingContainer;
+import org.embulk.spi.time.StrptimeParser.FormatBag;
```
muga: Fixed 536f36d. Thank you.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParse.java
```diff
@@ -0,0 +1,283 @@
+package org.embulk.spi.time;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class RubyDateParse
```
muga: Thank you for suggesting. I renamed this class with `TimeZoneConverter`. I also add comments about Github URL links to easy to catch up. bf1100b1d9c95265bcbfb261abfa4e4b11c08c73 and d8c4956cbd189e2fed2632eb403f0dee37a9c014

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParse.java
```diff
@@ -0,0 +1,283 @@
...
+{
```
muga: Thank you. Fixed 37f4db19e5950eccada1b917cfc5c04e099cb3be

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParse.java
```diff
@@ -0,0 +1,283 @@
...
+    // Ported zones_source in ext/date/date_parse.c
```
muga: MRI 2.3.1 uses `zones_source` in ext/date/date_parse.rb. MRI 2.3.1 is used for JRuby 9.1.x. 
https://github.com/ruby/ruby/blob/394fa89c67722d35bdda89f10c7de5c304a5efb1/ext/date/date_parse.c

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeFormat.java
muga: Since JRuby has EPL license, I believe that the porting should be ok. And I added notations on licensing according to your review comments. d8c4956cbd189e2fed2632eb403f0dee37a9c014

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeToken.java
muga: Added d8c4956cbd189e2fed2632eb403f0dee37a9c014.

muga:
@dmikurube Thank you for reviewing. I fixed and replied on your comments. Please take a look when you have time. 

> General comment in addition: wow about testing them with the same tests with #617?

Yes, I will use `test/mri/date/test_date_strptime.rb` that was added on #617 and test this parser. We need to modify `test/run-test.rb` to use the parser but, I will file the change later.

muga:
@kares Thank you for reviewing. I will check your comments soon.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeToken.java
```diff
@@ -0,0 +1,109 @@
...
+        CONVERSION2TOKEN['S'] = new StrptimeToken(FORMAT_SECONDS);
+        CONVERSION2TOKEN['s'] = new StrptimeToken(FORMAT_EPOCH);
+        CONVERSION2TOKEN['U'] = new StrptimeToken(FORMAT_WEEK_YEAR_S);
+        CONVERSION2TOKEN['u'] = new StrptimeToken(FORMAT_DAY_WEEK2);
+        CONVERSION2TOKEN['V'] = new StrptimeToken(FORMAT_WEEK_WEEKYEAR);
+        CONVERSION2TOKEN['W'] = new StrptimeToken(FORMAT_WEEK_YEAR_M);
+        CONVERSION2TOKEN['w'] = new StrptimeToken(FORMAT_DAY_WEEK);
+        CONVERSION2TOKEN['Y'] = new StrptimeToken(FORMAT_YEAR_LONG);
+        CONVERSION2TOKEN['y'] = new StrptimeToken(FORMAT_YEAR_SHORT);
+    }
```
hiroyuki-sato: @muga 
Does this parser support `%Q`.
I didn't find the following code for `%Q` in the `StrptimeToken.java`.
If this code does not need here, please ignore this comment.


```java
CONVERSION2TOKEN['Q'] = new StrptimeToken(FORMAT_MICROSEC_EPOCH);
```

muga: @hiroyuki-sato Thank you for reviewing. Since `%Q` is cared in `StrptimeParser#compilePattern` method, it works fine. But, I think that your idea is better than the current implementation. I will fix it later.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParse.java
muga: Thank you. Fixed c0d3febc272a59fca6fc526ca0081da6f03e7e4e.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParser.java
muga: Thank you for your advice. Fixed it on b9903b3690c08faace8ddfe1420a8478de4d038e. In the fix, `context` field is removed in the class and each method receives the `context` as 1st argument. If the fix is not expected, please give me an example code that you expected.

muga:
@kares I fixed `RubyDateParser` and `RubyDateParse` classes according to your comments. Please take a look when you have time.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeToken.java
muga: Fixed f9bee95cca944401c0848074a8576e81e145c556.

muga:
@hiroyuki-sato I fixed `%Q` token retrieval. Please take a look when you have time.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeParser.java
```diff
@@ -0,0 +1,883 @@
...
+import org.embulk.spi.time.lexer.StrptimeLexer;
+
+/**
+ * This is Java implementation of ext/date/date_strptime.c in Ruby v2.3.1.
+ * @see https://github.com/ruby/ruby/blob/394fa89c67722d35bdda89f10c7de5c304a5efb1/ext/date/date_strptime.c
+ *
+ * TODO
+ * This class is tentatively required for {@code TimestampParser} class.
+ * After embulk-jruby-strptime will be merged into JRuby, embulk-jruby-strptime
+ * will be removed from Embulk.
```
dmikurube: To be honest, I'm not very confident if we should remove these even once these are merged into JRuby.

It is still an option to keep them as our own "Ruby-compatible date/time parser implemented in pure-Java" in Embulk so that the Embulk core does not have any dependency on JRuby. Using JRuby's implementation means that the Embulk core's behavior may be unintentionally affected by JRuby's future updates.

Contributing to JRuby is very nice, but it's worth being careful to depend on that in our very internal data structure.

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParser.java
dmikurube: See my comment for `StrptimeParser.java`. It means that we may want to remove only the dependency here without removing our own implementation in Embulk.

kares:
@muga very nice, thank you ... this is looking JRuby 'ready' (assuming it won't break any Ruby tests/specs)

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeParser.java
muga: Agree. As I put the TODO comments, I was going to keep embulk-jruby-strptime until I upgrade JRuby that includes the RubyDateParser for Embulk. I will fix the comments.

muga: Fixed 1e54a0b7ed5eea71a0109dfe04b380bbca0e814d.

muga:
@kares Thank you for reviewing.

@dmikurube I fixed TODO comments. Please take a look?

embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeParser.java
dmikurube: @muga Hmm, I didn't mean that.

What I meant is: Even after the code is merged into JRuby, we may want to keep this in Embulk. The reason is:
> Using JRuby's implementation means that the Embulk core's behavior may be unintentionally affected by JRuby's future updates.

muga: @dmikurube Sorry for the late response.. As talked directly about it to you, I understand your concerns.

I'm OK to keep it in Embulk. I agree with your thoughts. The reason why I tried to merge it into JRuby is because I wanted to reduce the maintenance cost for us. To do that, I'm also OK to accept the risk that you concerned. If Embulk core would use JRuby's and might be affected by the future updates, we would fix or fallback our original one at that time.

How do you think that we introduce an option to switch strptime implementation between our original and JRuby's. Users can specify it in system config or somewhere. The option can help the above falling back. The default value of the option is JRuby's.

How about it?

dmikurube: @muga Personally, I'm not positive on that.

I don't think both users and ourselves get any benefit by that option. It rather confuses users. It does not reduce our maintenance costs. Why not using just our own version?

dmikurube: I'm also worried that JRuby's class name/package may change. That's kinds of their internal classes. We have no options stop such changes.

In that case, the Embulk core may not work for a kind of `LinkageError` or like that once the class name is referred even if we have such an option for users.

muga: Ok, let's keep in Embulk. `StrptimeParser` class that we contributed to JRuby is `public` modifier but, that's kinds of internal class from view point of almost all of JRuby users. Therefore, the class possibly might be refactored dramatically (name and package as well).

dmikurube: Okay.

dmikurube:
@muga Then, any more points where I should take a look?

muga:
@dmikurube Yea, I don't have anything now. Ready to merge? :-)

dmikurube:
@muga I think so. Just one point to take care is: whether we do merge this big update into v0.8.26, which already includes many updates: https://github.com/embulk/embulk/milestone/2

muga:
@dmikurube Ok, this feature will be merged for next version 0.8.27. Not include it in 0.8.26.

dmikurube:
@muga 0.8.6 has been released. Let's merge this toward next 0.8.27. :)

muga:
Thank you! I will merge this after git rebase.

dmikurube:
@muga Oops, sorry, I wrongly force-pushed 274c04f7ba8e9dcf73551f9035c6fd28a980807c to `master`...

Can you push this 3a13829217e32700be1b4e150eb8d56d74363ea7 again directly to `master` if you have that locally? I couldn't get that for the time being. If not, let's make another merge...

To avoid this happening again, I've "protected" the `master` branch on GitHib.

dmikurube:
Fortunately, 3a13829217e32700be1b4e150eb8d56d74363ea7 is just one merge of 274c04f7ba8e9dcf73551f9035c6fd28a980807c and this PR. There are not so many crisis there.

dmikurube:
e6e7d9425dd45a35fde054bc642660a5576ce130 (your last commit in this PR) is backed up just in case...

dmikurube:
https://github.com/embulk/embulk/tree/java_timestamp_parser_separateding_ruby_date_parser is e6e7d9425dd45a35fde054bc642660a5576ce130, jfyi.

dmikurube:
Finally 3a13829217e32700be1b4e150eb8d56d74363ea7 is back. Sorry for bothering you...

muga:
@dmikurube I didn't notice something happened.. but, thank you for taking time to fix it!

PR #615, opened 2017-05-11
Update Jackson v2.6.7

muga:
This PR upgrades Jackson v2.6.7 for Embulk. Latest version of Jackson is v2.8 but, since Jackson is one of fundamental libraries, we've been gradually upgrading it.


muga:

muga:
@dmikurube Thank you for reviewing. I confirmed that my plugin by jackson 2.5.3 works fine for this Embulk by jackson 2.6.7. If plugins by jackson 2.7 or 2.8 exist, they don't work but, we don't receive such issue tickets. So, I think that it should be ok.

muga:
Merged.

PR #617, opened 2017-05-13
Add test/mri/date/test_date_strptime.rb for #611

muga:
This PR ports test/mri/date/test_date_strptime.rb from JRuby. Because we need to confirm #611 's  `RubyDateParser` behavior is correct or not after #611. It's better to add this test in advance. 

I skipped `test_sz` in the test_date_strptime.rb because the spec is unstable and doesn't pass with our JRuby 9.1.5.0.


muga:

muga:
Thank you-!

Issue #618, opened 2017-05-13
Attribute exec is required but not set


hiroyuki-sato:
This behavior is incompatible previous version. 

I got `Attribute exec is required but not set`

* embulk 0.8.19 

## Reproduce steps

```
embulk example sample
embulk guess sample/seed.yml -o config.yml
embulk preview config.yml
```

## Exception

```
2017-05-13 22:10:16.490 +0900: Embulk v0.8.19
org.embulk.config.ConfigException: Attribute exec is required but not set
	at org.embulk.config.DataSourceImpl.getNested(org/embulk/config/DataSourceImpl.java:96)
	at org.embulk.config.DataSourceImpl.getNested(org/embulk/config/DataSourceImpl.java:12)
	at org.embulk.exec.PreviewExecutor.doPreview(org/embulk/exec/PreviewExecutor.java:109)
	at org.embulk.exec.PreviewExecutor.access$000(org/embulk/exec/PreviewExecutor.java:31)
	at org.embulk.exec.PreviewExecutor$1.run(org/embulk/exec/PreviewExecutor.java:83)
	at org.embulk.exec.PreviewExecutor$1.run(org/embulk/exec/PreviewExecutor.java:79)
	at org.embulk.spi.Exec.doWith(org/embulk/spi/Exec.java:25)
	at org.embulk.exec.PreviewExecutor.preview(org/embulk/exec/PreviewExecutor.java:79)
	at org.embulk.EmbulkEmbed.preview(org/embulk/EmbulkEmbed.java:169)
	at java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:498)
	at org.jruby.javasupport.JavaMethod.invokeDirectWithExceptionHandling(org/jruby/javasupport/JavaMethod.java:453)
	at org.jruby.javasupport.JavaMethod.invokeDirect(org/jruby/javasupport/JavaMethod.java:314)
	at RUBY.preview(/Users/user/.embulk/bin/embulk!/embulk/runner.rb:35)
	at RUBY.run(/Users/user/.embulk/bin/embulk!/embulk/command/embulk_run.rb:267)
	at RUBY.<main>(/Users/user/.embulk/bin/embulk!/embulk/command/embulk_main.rb:2)
	at org.jruby.Ruby.runInterpreter(org/jruby/Ruby.java:850)
	at org.jruby.Ruby.loadFile(org/jruby/Ruby.java:2976)
[...]

PR #619, opened 2017-05-13
PreviewExecutor: to fix incompatible behavior that it requires exec: config

muga:
This PR fixes incompatible behavior of `PreviewExecutor` with previous version. That unexpectedly requires `exec:` config during `preview` command is executed. 0.8.18 or older version doesn't require `exec:` config.

The root cause of this issue is that it calls`config.getNested("exec")`. If `config` doesn't have `exec:` config, it gets the `ConfigException`. It should use `task.getExecConfig()` because it will receive default value `{}` if `config` doesn't have `exec:` config.

#618


muga:
@dmikurube thank you. After CI passed, will merge this into master.

muga:
Merged.

Issue #627, opened 2017-05-15
Can't execute embulk with bundle environment


kakoni:
Embulk(0.8.19, and 0.8.20) can't execute if i try to run it with bundle environment

Following exception is presented;
```
bin/embulk mkbundle em2;bin/embulk -b em2 preview conf/foo.yml.liquid
2017-05-15 08:52:31.860 +0300: Embulk v0.8.20
undefined method `line_number' for "in:\n  type: file\n  path_prefix: ":String
    uri:classloader:/liquid/block_body.rb:99:in `block in render'
    org/jruby/RubyArray.java:1734:in `each'
    uri:classloader:/liquid/block_body.rb:75:in `render'
    uri:classloader:/liquid/template.rb:208:in `block in render'
    uri:classloader:/liquid/template.rb:242:in `with_profiling'
    uri:classloader:/liquid/template.rb:207:in `render'
    uri:classloader:/embulk/runner.rb:154:in `run_liquid'
    uri:classloader:/embulk/runner.rb:129:in `read_config'
    uri:classloader:/embulk/runner.rb:32:in `preview'
    uri:classloader:/embulk/command/embulk_run.rb:267:in `run'
    uri:classloader:/embulk/command/embulk_main.rb:2:in `<main>'
    org/jruby/RubyKernel.java:956:in `require'
    file:/Users/kakoni/dev/beans/embulk/bin/embulk!/embulk/command/embulk_bundle.rb:30:in `<main>'

Error: undefined method `line_number' for "in:\n  type: file\n  path_prefix: ":String

```

foo.yml.liquid
```
in:
  type: file
  path_prefix: file.csv
  parser:
    type: csv
    columns:
      - {name: a, type: long}
      - {name: b, type: string}
out:
  type: stdout
```

## Environment
```
ProductName:    Mac OS X
ProductVersion: 10.12.4
BuildVersion:   16E195
```

hiroyuki-sato:
@kakoni 

Thank you for reporting. 

I can reproduce your issue in my environment.
Is your problem occur if you use `-b` only? 
It this correct?

It seems that this is #623 related issue.
This issue will fix 0.8.21. 

Please execute `embulk selfupdate 0.8.18`  and downgrade embulk until fix this issue. 
I'll let you know when this issue fixed. 

Sorry for inconvenient.

hiroyuki-sato:
@dmikurube @muga reprodece steps.

## Prepare reprodude steps.

```
%  embulk mkbundle em2
[...]

PR #632, opened 2017-05-15
Downgrade liquid 4.0.0 to 3.0.6

muga:
This PR downgrades liquid 4.0.0 to 3.0.6. Liquid was upgraded on Embulk v0.8.19 but, the upgrading causes regression https://github.com/embulk/embulk/issues/627.


muga:
Thank you for reviewing. I will merge it after CI passed.

muga:
Merged.

Issue #529, opened 2017-01-10
use nanosec for ExecSession timestamp


ukisoft:
Using embulk, sometimes there is error because session are same.
So, I would like to use nanosec for ExecSession timestamp.

I found TODO. Could I know when it will be improved?

https://github.com/embulk/embulk/blob/d807e6aaae9bb02b71d9aae355b1298356ed26b5/embulk-core/src/main/java/org/embulk/spi/ExecSession.java
```
        public ExecSession build()
        {
            if (transactionTime == null) {
                transactionTime = Timestamp.ofEpochMilli(System.currentTimeMillis());  // TODO get nanoseconds for default
            }
            return new ExecSession(injector, transactionTime, Optional.fromNullable(loggerFactory));
        }
```

hiroyuki-sato:
Could you tell me more detail about the same session error?
Do you have any logs about it?

I asked about @frsyuki this issue.  
He said that nanoseconds are better than milli seconds.
But there is no method like `System.currentTimeMillis` for nano seconds.

ukisoft:
Thank you for your comment.

I use embulk from MySQL to Treasure Data with digdag.
And I'm sorry, I use embulk on my company, so I can't give you our full logs.
Instead of that, I copied some logs at end of this message.
I hope it helps you to improve embulk.

Anyway, I asked also Treasure Data support officer before about it.
Fortunately they answered last week, and they recommend session configuration of td plugin: https://github.com/treasure-data/embulk-output-td.

I will try it. Thank you.


Logs are like this. 
Session is dummy, but it seems that same session caused errors.

----- first embulk log partially -----
2016-11-07 06:45:13.817 +0900 [INFO] (0001:transaction): Create bulk_import session embulk_20161106_123456_789000000
2016-11-07 06:45:14.152 +0900 [INFO] (0001:transaction): {done:  0 / 1, running: 0}
2016-11-07 06:45:14.367 +0900 [INFO] (0001:transaction): {done:  1 / 1, running: 0}
2016-11-07 06:45:19.563 +0900 [INFO] (0001:cleanup): Deleting bulk import session 'embulk_20161106_123456_789000000'
org.embulk.exec.PartialExecutionException: com.treasuredata.client.TDClientHttpNotFoundException: [TARGET_NOT_FOUND] [404:Not Found] API request to /v3/bulk_import/show/embulk_20161106_123456_789000000 has failed: Resource not found

----- second embulk log partially -----
[...]

PR #641, opened 2017-05-19
[RFC/WIP] ExecSession.getTransactionUniqueName

hiroyuki-sato:
I roughly implement `ExecSession.getTransactionUniqueName`
How do you think this idea?

This issue may solve #529.

First I read [this comment](https://github.com/embulk/embulk/issues/529#issuecomment-274415533) and I knew it's embulk-output-td issue.
Today I read this line.
Oh!, It needs `ExecSession.getTransactionUniqueName()` method.
That's why I implement it.

It returns a random session string like `20170519_053301_000000000_mrh0tkfs45tp5`. 
(Maybe, I don't how to test this code yet)
This is temporary implementation. Because if we update Java8, we can use nanosec. 

https://github.com/treasure-data/embulk-output-td/blob/master/src/main/java/org/embulk/output/td/TdOutputPlugin.java#L576

```java
  Timestamp time = exec.getTransactionTime(); // TODO implement Exec.getTransactionUniqueName()
  return String.format("embulk_%s_%09d",
          DateTimeFormat.forPattern("yyyyMMdd_HHmmss").withZoneUTC().print(time.getEpochSecond() * 1000),
          time.getNano());
```

Ref:  [Twitter](https://twitter.com/nora96o/status/865121937248223232)

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/ExecSession.java
+++ b/embulk-core/src/main/java/org/embulk/spi/ExecSession.java
@@ -1,6 +1,7 @@
 package org.embulk.spi;
 
 import org.joda.time.DateTimeZone;
+import org.joda.time.format.DateTimeFormat;
 import org.slf4j.Logger;
 import org.slf4j.ILoggerFactory;
 import com.google.common.base.Optional;
@@ -22,6 +23,10 @@ import org.embulk.spi.time.Timestamp;
 import org.embulk.spi.time.TimestampFormatter;
 import org.embulk.spi.time.TimestampFormatter.FormatterTask;
 
+import java.math.BigInteger;
+import java.security.SecureRandom;
+import java.util.Locale;
+
 public class ExecSession
 {
     private final Injector injector;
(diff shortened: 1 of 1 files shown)


muga:
@hiroyuki-sato Thank you for investigating and suggesting. 

It seems, the root cause of the problem is that embulk-output-td's session names are duplicated so, I think that embulk-output-td should fix this issue by using random string that you mentioned or UUID or something.

hiroyuki-sato:
@muga Thank you for your comment. 

Here is my understanding. Is this correct?. If so, I'll close this issue.

* The core doesn't provide `ExecSession.getTransactionUniqueName` at this time. 
* If we update to Java8, we can use nanoseconds time. The`ExecSession.getTransactionTime` return high precision time.

muga:
@hiroyuki-sato Personally it's good to support `unique_transaction_name` in Embulk and embulk-output-td uses the unique name as bulk_import session name but, Embulk doesn't suport the parameter for now. So, I'll try to fix it in embulk-output-td as workaround.

I also think that it's good to pass the value from an external tool to Embulk like workflow or something personally. We need to discuss about the design but.

hiroyuki-sato:
@muga 
Thank you for your comment. 
I just wanted to ask your opinion for future development.

Issue #630, opened 2017-05-15
Couldn't get causes if the errors inherits Embulk::DataError or Embulk::ConfigError


civitaspo:
Hi maintainers.

When I implement JAVA plugins, we use ConfigException or DataException if the errors is not retriable. So, I've believed using ConfigError or DataError in that case is one of the best practices in JRuby plugin implementation. ref. http://qiita.com/sakama/items/4c5b08a50f6e2e313afa#embulk%E3%81%AB%E3%81%8A%E3%81%91%E3%82%8B%E4%BE%8B%E5%A4%96%E5%87%A6%E7%90%86

But, I'm worried about the errors(or inherited the errors) which cannot have causes. Errors which inherits ConfigError or DataError do not have the causes. So, I implements my own error classes as getting causes like https://github.com/medjed/embulk-input-google_spreadsheets/blob/f3cc69cebf8340180b9760f067855cf87e600419/lib/embulk/input/google_spreadsheets/error.rb .

When I implement JRuby plugins, what is the best way to handle errors?

The sample code is here: https://github.com/civitaspo/embulk-input-issue_suppress_jruby_error

```
...<snip>...

      def run
        page_builder.add(["example-value", 1, 0.1])
        page_builder.add(["example-value", 2, 0.2])
        page_builder.finish

        begin
          err
        rescue => e
          ::Embulk.logger.error "e: {class:#{e.class},msg:#{e}}"
          # error occurs here
          # org.jruby.exceptions.RaiseException: (NoMethodError) undefined method `cause' for nil:NilClass
          ::Embulk.logger.error "e.cause: {class:#{e.cause.class},msg:#{e.cause}}"
          ::Embulk.logger.error "e.cause.cause: {class:#{e.cause.cause.class},msg:#{e.cause.cause}}"
          ::Embulk.logger.error "e.cause.cause.cause: {class:#{e.cause.cause.cause.class},msg:#{e.cause.cause.cause}}"
        end

        task_report = {}
        return task_report
      end

...<snip>...

      class Error < ::Embulk::DataError
      end

      class Error1 < ::StandardError; end
      class Error2 < ::StandardError; end
      class Error3 < ::StandardError; end
      
      def err1
        raise Error1, "err"
      end
      
      def err2
        begin
          err1
        rescue Error1 => e
          raise Error2, e
        end
      end
      
      def err3
        begin
          err2
        rescue Error2 => e
          raise Error3, e
        end
      end

      def err
        begin
          err3
        rescue => e
[...]

PR #646, opened 2017-05-23
Remove unnecessary error validations

civitaspo:
fix https://github.com/embulk/embulk/issues/630

As I wrote in https://github.com/embulk/embulk/issues/630, `ConfigError`, `DataError` and `PluginLoadError` cannot have `cause`. Since the behavior is difficult for Ruby Plugin Creators, I think the error class should behave like inherited Ruby Exception class. Now I found the error class behaves like inherited Ruby Exception class even if they inherit Java Error class. See https://gist.github.com/civitaspo/415142e2d567ec5ff79827b35d50f9fb (This is just the same on issue 630.)
So, I remove the special implementations.

### DIFF
--- a/lib/embulk/error.rb
+++ b/lib/embulk/error.rb
@@ -3,25 +3,11 @@ module Embulk
   # ConfigError is not a ::StandardError but is a java.lang.RuntimeException.
   # "rescue => e" can rescues ConfigError.
   class ConfigError < Java::Config::ConfigException
-    def initialize(message=nil)
-      if message
-        super(message.to_s)
-      else
-        super()
-      end
-    end
   end
 
   # DataError is not a ::StandardError but is a java.lang.RuntimeException.
   # "rescue => e" can rescues DataError.
   class DataError < Java::SPI::DataException
-    def initialize(message=nil)
-      if message
-        super(message.to_s)
-      else
-        super()
-      end
-    end
   end
 
   class PluginLoadError < ConfigError


dmikurube:
@civitaspo Thanks for your contribution! I'm not very sure of intention of the initial implementor (@frsyuki ?), but the fix basically looks good.

I was not sure JRuby's inheritance from a Java class inherits constructors, unlike Java.

@frsyuki @muga Any thoughts?

dmikurube:
@muga Just reminding...

muga:
I will take a look.

dmikurube:
@muga I'll be merging this in a day unless you have any comments.

hiroyuki-sato:
I met and talked @civitaspo and @frsyuki in the last week. 
IIUC @frsyuki hasn't any concern. 

@civitaspo do you have any concern after talked with @frsyuki?

civitaspo:
He said that this JRuby error implementation only wanted to use JRuby's error class that inherited DataException and ConfigExcepiton to determine whether it can retry or not, and there was no more intention . So there is no big problem as long as it is implemented to wrap JRuby error classes.

In this implementation, problems may arise if DataError or ConfigError is caught as a Ruby error. However, since it inherits java's RuntimeException, it is impossible to behave exactly the same as JRuby's error class. Also, if we do our best to implement the same behavior as JRuby's error, there will be no big effect to Embulk plugin developpers, I think.

muga:
@civitaspo Sorry for late reply. Basically LGTM. Do you think that we don't need any tests to confirm this change works fine?

civitaspo:
Ok, I will add tests about this change as soon as possible. Please wait a bit to merge.

civitaspo:
Oh...wait. When writing a test, a behavior different from the behavior assumed by myself has appeared ...

muga:
@civitaspo good catch. please let me know when your fix and the test are done.

civitaspo:
@muga @dmikurube  I'm really sorry. By writing a test I noticed that this change was a terrible change. With this change, `{Data,Config}Error` could not wrap JRuby's error. This p-r should be closed.

I should have investigated the following cases.

1. Throw Java Exception wrapped with Java Exception as `{Data,Config}Error`
1. Throw Java Exception wrapped with Ruby Exception as `{Data,Config}Error`
1. Throw Ruby Exception wrapped with Java Exception as `{Data,Config}Error`
1. Throw Ruby Exception wrapped with Ruby Exception as `{Data,Config}Error`
1. Throw String as `{Data,Config}Error`
1. Throw nil as `{Data,Config}Error`

However, the case (3, 4) which wraps Ruby Exception is missing from the investigation (although I do not remember where it leaked). When I tried the test, it turned out that it does not function properly.

The reason is simple, because Ruby Exception does not inherit `java.lang.Throwable`. I misunderstood that it worked well in the survey, and thought that Ruby Exception would be cast in `java.lang.Throwable`. Below is the code which can easily reproduce the problem.

```error.rb
require 'java'

raise ::Java::JavaLang::RuntimeException.new(::StandardError.new("error!"))
# => NameError: no constructor for arguments (org.jruby.RubyException) on Java::JavaLang::RuntimeException
#      available overloads:
#        (java.lang.String)
#        (java.lang.Throwable)
#      <main> at error.rb:11
```

I sent something terrible p-r and I'm very sorry. Since this problem was originally sent p-r without writing the test writing place, we will send p-r with tests from the next time onwards. Especially JRuby and Java's integration part is carefully sent p-r because I think that it is a part where all members can not grasp the behavior.

I am very sorry ...

dmikurube:
@civitaspo Ah, thanks for your investigation. Got it. That was the reason why `ConfigError` and `DataError` were converting with `to_s`.

But AFAIU, the original issue #630 is still active. We may still want to solve that in a different manner...

civitaspo:
@dmikurube thx for the reply. 

I think that the remaining way is to hack by using `org.jruby.exceptions.RaiseException`, but it can work...?  hmm...

```error.rb
require 'java'

raise ::Java::JavaLang::RuntimeException.new(::Java::OrgJrubyExceptions::RaiseException.new(::StandardError.new("error!")))
# => Unhandled Java exception: java.lang.RuntimeException: org.jruby.exceptions.RaiseException: (StandardError) error!
#    java.lang.RuntimeException: org.jruby.exceptions.RaiseException: (StandardError) error!
#             newInstance0 at sun/reflect/NativeConstructorAccessorImpl.java:-2
#              newInstance at sun/reflect/NativeConstructorAccessorImpl.java:62
#              newInstance at sun/reflect/DelegatingConstructorAccessorImpl.java:45
#              newInstance at java/lang/reflect/Constructor.java:423
#        newInstanceDirect at org/jruby/javasupport/JavaConstructor.java:278
#                     call at org/jruby/java/invokers/ConstructorInvoker.java:85
#                     call at org/jruby/java/invokers/ConstructorInvoker.java:174
#             cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:348
#                callBlock at org/jruby/runtime/callsite/CachingCallSite.java:173
#                     call at org/jruby/runtime/callsite/CachingCallSite.java:177
#                     call at org/jruby/java/proxies/ConcreteJavaProxy.java:56
#             cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:348
#                callBlock at org/jruby/runtime/callsite/CachingCallSite.java:173
#                     call at org/jruby/runtime/callsite/CachingCallSite.java:177
#              newInstance at org/jruby/RubyClass.java:1001
#                     call at org/jruby/RubyClass$INVOKER$i$newInstance.gen:-1
#                     call at org/jruby/internal/runtime/methods/DynamicMethod.java:200
#                     call at org/jruby/java/proxies/ConcreteJavaProxy.java:158
#             cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:338
#                     call at org/jruby/runtime/callsite/CachingCallSite.java:163
#         invokeOther5:new at error.rb:3
#                   <main> at error.rb:3
#      invokeWithArguments at java/lang/invoke/MethodHandle.java:627
#                     load at org/jruby/ir/Compiler.java:95
#                runScript at org/jruby/Ruby.java:827
#              runNormally at org/jruby/Ruby.java:746
#              runNormally at org/jruby/Ruby.java:764
#              runFromMain at org/jruby/Ruby.java:577
#            doRunFromMain at org/jruby/Main.java:417
#              internalRun at org/jruby/Main.java:305
#                      run at org/jruby/Main.java:232
#                     main at org/jruby/Main.java:204
#    
#    Caused by:
#    org.jruby.exceptions.RaiseException: (StandardError) error!
#      <main> at error.rb:3
```

dmikurube:
Got it. It's not very trivial... Let's keep #630 open and keep tuned. Your ideas are always welcome!

PR #651, opened 2017-05-28
Add clean illegal characters mode to json parser.

smdmts:
A JSON which including broken-encoded character is throwing an exception from Jackson because that is supported pure JSON specification in below reasons.
- https://github.com/FasterXML/jackson-core/issues/222#issuecomment-146697502

But ideally, I want to force loading the broken-encoded JSON files.
Therefore, I add `clean_illegal_char` mode to the JSON parser and that works cleansing to illegal char inside of JSON characters.

Please, check this PRs.

- Stack Trace.
```
Caused by: com.fasterxml.jackson.core.JsonParseException: Invalid UTF-8 middle byte 0x5c
 at [Source: org.embulk.spi.util.FileInputInputStream@4f26425b; line: 17821, column: 69]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1581)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:533)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._reportInvalidOther(UTF8StreamJsonParser.java:3473)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._reportInvalidOther(UTF8StreamJsonParser.java:3480)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._decodeUtf8_2(UTF8StreamJsonParser.java:3254)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._finishString2(UTF8StreamJsonParser.java:2462)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._finishAndReturnString(UTF8StreamJsonParser.java:2414)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser.getText(UTF8StreamJsonParser.java:285)
	at org.embulk.spi.json.JsonParser$AbstractParseContext.jsonTokenToValue(JsonParser.java:188)
	at org.embulk.spi.json.JsonParser$AbstractParseContext.jsonTokenToValue(JsonParser.java:220)
	at org.embulk.spi.json.JsonParser$AbstractParseContext.next(JsonParser.java:152)
```


muga:
@smdmts Thank you for sending the PR. As you know that JSON specification supports only UTF-8, UTF-16 and UTF-32 as valid encodings, it would be useful if such non-standard JSON data could be ingested by Embulk and the JSON parser. 

I'm OK to add such conversion to JSON parser plugin in embulk-standards but, I ideally think that it's good to handle such conversion by decoder plugin. How about it? @dmikurube

dmikurube:
I agree with @muga. Adding such cleansing methods in every possible parser doesn't sound very reasonable to me.

Creating a new decoder plugin to cleanup the illegal characters, and apply the decoder before this JSON parser might be the cleanest way. What do you think, @smdmts?

smdmts:
Hi @muga, @dmikurube 
Thanks for prompt replying. I agreed that cleansing methods responsibility is better to move decoder plugin. But I have the one point of the issue what JSON's backslash(\\) escape handling.

- https://stackoverflow.com/questions/19176024/how-to-escape-special-characters-in-building-a-json-string

So, I have an idea that divides decoder plugin as the remove-nonstandard-utf8 and decoder plugin.

### add-plugins
- RemoveNonStanradUTF8BytesDecorderPlugin.java
  - removing a dust of including any bytes which breaking UTF-8.

### modify-plugins
- JsonParserPlugin.java
  - additional implementing remove unfit-backslash mode.

How do you think this Idea?
Thankas,

hiroyuki-sato:
@smdmts 

Maybe you know, 
you can develop decoder as a plugin.
So you don't need add decoder into core. 

```
embulk new java-decoder remove_nonstandard_utf8 
```

dmikurube:
@hiroyuki-sato Thanks for your follow-up. Yes, `RemoveNonStanradUTF8BytesDecorderPlugin.java` are to be implemented as a third-party plugin, not in Embulk's standard plugins.

@smdmts What does the "unfit backslash" mean?

smdmts:
@dmikurube 
The backslash such as (\\a) is *not* allowed JSON specs and if including that JACKSON is throwing an exception.
e.f.  \\t (tab) , \\n (new line) is allowed. 

Therefore I want to have unfit backslash cleansing mode.

dmikurube:
@smdmts Thanks! Got it. It sounds reasonable to implement that part in the standard's JSON parser.

How about making this config as follows?
`invalid_string_escapes: PASSTHROUGH | SKIP | UNESCAPE` (default: PASSTHROUGH)

- `PASSTHROUGH` to do nothing. (Default)
- `SKIP' to remove these invalid escapes. (Wanted in this PR)
- `UNESCAPE` to convert invalid escapes (like `\a`) just to remove `\` (like just `a`). (Not necessary in this PR. Just for future extension.)

dmikurube:
Just a memo: https://tools.ietf.org/html/rfc7159#section-7

smdmts:
@dmikurube Thank you for your reviewing. Would you mind tell me more details and that behavior is in the below?

- SKIP   `\a` convertTo `` (empty value)
- UNESCAPE `\a` convertTo `a`

dmikurube:
@smdmts Thanks for taking care of this. I think:
* `SKIP`: Yes, that's what I meant. Maybe nice to refer the RFC for characters to be passed.
* `UNESCAPE`: It is not necessary for the time being, but your understanding is correct. I showed this to represent just `true`/`false` may not be sufficient.

dmikurube:
Ah, you've already implemented. Thanks! Taking a look...

embulk-standards/src/test/java/org/embulk/standards/TestJsonParserPlugin.java
```diff
@@ -1,27 +1,27 @@
...
 import com.google.common.collect.ImmutableList;
+import com.google.common.io.CharSource;
 import org.embulk.EmbulkTestRuntime;
 import org.embulk.config.ConfigSource;
 import org.embulk.config.TaskSource;
-import org.embulk.spi.DataException;
-import org.embulk.spi.FileInput;
-import org.embulk.spi.ParserPlugin;
-import org.embulk.spi.Schema;
+import org.embulk.spi.*;
```
dmikurube: Please avoid wildcard imports.

embulk-standards/src/test/java/org/embulk/standards/TestJsonParserPlugin.java
```diff
@@ -1,27 +1,27 @@
...
+import org.joda.time.DateTimeZone;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.msgpack.value.Value;
 
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
```
dmikurube: ditto.

embulk-standards/src/test/java/org/embulk/standards/TestJsonParserPlugin.java
```diff
@@ -1,27 +1,27 @@
...
+import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
 
+import static org.embulk.standards.JsonParserPlugin.InvalidEscapeStringPolicy.*;
```
dmikurube: ditto.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
         return new PageBuilder(Exec.getBufferAllocator(), schema, output);
     }
 
-    private JsonParser.Stream newJsonStream(FileInputInputStream in)
-            throws IOException
-    {
-        return new JsonParser().open(in);
+    private JsonParser.Stream newJsonStream(FileInputInputStream in , PluginTask task)
```
dmikurube: style nitpick: no space before comma

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
...
+                return new JsonParser().open(in);
+        }
+    }
+
+    Function<String, CharSource> invalidEscapeStringFunction(final InvalidEscapeStringPolicy policy) {
+        return new Function<String, CharSource>() {
+            final Pattern p = Pattern.compile("\\p{XDigit}+");
+            @Override
+            public CharSource apply(@Nullable String input) {
+                assert input != null;
```
dmikurube: We often used `com.google.common.base.Preconditions` instead of `assert`. It's fine, though.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
...
+                        invalidEscapeStringFunction(policy));
+                return new JsonParser().open(new ReaderInputStream(CharSource.concat(lines).openStream()));
+            case PASSTHROUGH:
+            default:
+                return new JsonParser().open(in);
+        }
+    }
+
+    Function<String, CharSource> invalidEscapeStringFunction(final InvalidEscapeStringPolicy policy) {
+        return new Function<String, CharSource>() {
```
dmikurube: We'd like to go Java 8 soon... Sorry about that. ;)

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
...
+            final Pattern p = Pattern.compile("\\p{XDigit}+");
```
dmikurube: Let's use a more descriptive variable name -- such as `digitsPattern` or like that.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
...
+            @Override
+            public CharSource apply(@Nullable String input) {
+                assert input != null;
+                if (policy == InvalidEscapeStringPolicy.PASSTHROUGH) {
+                    return CharSource.wrap(input);
+                }
+                StringBuilder s = new StringBuilder();
```
dmikurube: Same as `p`: how about `builder` or `stringBuidler`?

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
...
+                char[] charArray = input.toCharArray();
+                for (int i=0 ; i < charArray.length ; i++) {
```
dmikurube: style nitpick: we've usually written this like `for (int i = 0; i < charArray.length; i++) {`

`i` is fine for this kind of smaller scopes.

hiroyuki-sato:
@dmikurube 
All other configurations use lowercase. 
So I think it is better to use lowercase these settings for consistency.
Like`skip`, `passthrough` and `unescape`.
@smdmts is implementing it with uppercase.

How do you think?

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
...
+                                            i++;
+                                            break;
+                                        case UNESCAPE:
+                                            break;
+                                    }
+                                    break;
+                            }
+                        }
+                    } else {
+                        s.append(c);
```
dmikurube: Do we keep `\\` at the end of line?

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
...
+                                case 'n':
+                                case 'r':
+                                case 't':
+                                case '"':
+                                case '\\':
+                                case '/':
+                                    s.append(c);
+                                    break;
+                                case 'u': // hexstring such as \u0001
+                                    if (charArray.length > i + 5) {
```
dmikurube: How do you think it should work for: `\\u12<EOL>` and `\\u12xY`?

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -63,17 +96,16 @@ public void run(TaskSource taskSource, Schema schema, FileInput input, PageOutpu
         final Column column = schema.getColumn(0); // record column
 
         try (PageBuilder pageBuilder = newPageBuilder(schema, output);
-                FileInputInputStream in = new FileInputInputStream(input)) {
+             FileInputInputStream in = new FileInputInputStream(input)) {
```
dmikurube: Let's get this back -- changes unrelated to this topic.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -99,12 +131,79 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output)
...
+                        s.append(c);
+                    }
+                }
+                return CharSource.wrap(s.toString());
+            }
+        };
     }
 
+
+
```
dmikurube: Let's get these newlines back -- changes unrelated to this topic.

embulk-standards/src/test/java/org/embulk/standards/TestJsonParserPlugin.java
```diff
@@ -60,7 +71,7 @@ public void readNormalJson()
                         "\"_c1\":-10,\n" +
                         "\"_c2\":\"エンバルク\",\n" +
                         "\"_c3\":[\"e0\",\"e1\"]\n" +
-                "}",
+                        "}",
```
dmikurube: Let's get this back -- changes unrelated to this topic.

dmikurube:
@smdmts Left some comments. Most of them are just style nitpicking comments.

@hiroyuki-sato Ah, it might be an implicit consensus. We've conventionally used uppercases for kinds of CONSTANTS. For example, `CRLF` in CSV parser and `MINIMAL` for CSV formatter.

hiroyuki-sato:
@dmikurube You are right. Thanks.

embulk-standards/src/test/java/org/embulk/standards/TestJsonParserPlugin.java
smdmts: Sorry, I applied Airlift-style in my IDE.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
smdmts: No, if `\\` has at the end of line , it will removed at SKIP or UNESCAPE. I will check test and fix it.

```
{"a":"b}
\\
```

convert to:

```
{"a":"b}
```

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
smdmts: I think it could not be handling for remove `\\u12xY`of  only `\\u12`. Because next character may be valid string and if removed that it break some context in the json.
So, my implementation is removed only backslash like `u12xY` or `u12<EOL>`.

smdmts:
Umm..., AppVeyor was failed none-changed code by OOM.

dmikurube:
Hmm, it sometimes fails by OutOfMemory, but the failures are not OOM. We may need some investigation...

smdmts:
@dmikurube Thanks a lot. I fixed your comment on my PR.

dmikurube:
Ah, no, it's OOM. I'm triggering retry.

dmikurube:
AppVeyor passed. :)

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
dmikurube: Got it. Sounds reasonable both for `SKIP` and `UNESCAPE`.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
dmikurube: @smdmts Thanks. It's perfect for `UNESCAPE`, but it may be worth considering `SKIP`.

Only from `SKIP`'s simple definition like "remove these invalid escapes", behavior for `\\u` may not be trivial. For example, it can be still natural expectation for some users that `\\u12` is removed from `\\u12xY` or `\\u12<EOL>`.

Reasonable expectations for `\\u...` on `SKIP` may be:
1. Only `\\u` is always removed.
    * `\\u1234` ==> `1234` / `\\u12xY` ==> `12xY`
2. Whole `\\uXXXX` is removed if valid. Only `\\` is removed if invalid. (This PR)
    * `\\u1234` ==> (empty) / `\\u12xY` ==> `u12xY`
3. Whole `\\uXXXX` is removed if valid. Only `\\u` is removed if invalid. 
    * `\\u1234` ==> (empty) / `\\u12xY` ==> `12xY`
4. Whole `\\uX...` is removed even if the length of its valid part is shorter than 4.
    * `\\u1234` ==> (empty) / `\\u12xY` ==> `xY`

I thought 2 is not very consistent with `SKIP`'s definition to be honest. What do you think about 1 or 4?

smdmts: Sorry, my thought did not reach 1 or 4 and it's better than 2.
In my opinion of SKIP mode would be **1**. 

Because In 4 mode, if 12xY of "12" is a valid part of a sentence, it will breaking removing "12". 
How do you think this idea?

smdmts:
I implemented SKIP mode with the non-standard character with `\u` is below of *1*. https://github.com/embulk/embulk/pull/651#discussion_r119328863.


1. Only \\u is always removed.
\\u1234 ==> 1234 / \\u12xY ==> 12xY


- test code
https://github.com/embulk/embulk/pull/651/files#diff-b58b1c39f56eca489cc109b6b6c6947fR253

dmikurube:
@smdmts Ahh, an urgently merged PR https://github.com/embulk/embulk/pull/661 was conflicting with this PR... Sorry about that.

Is 0afac30 still conflicting?

dmikurube:
@smdmts Sorry, it was AppVeyor just failing. I've rebuilt that, and it's now succeeded. I'd merge this if @muga feels okay.

Any thoughts? > @muga

smdmts:
@dmikurube  I merged #661 at https://github.com/embulk/embulk/commit/0afac300d9050019b371819a834387ce4f530bb4 and it’s fine. Thanks,

dmikurube:
@muga I'll be merging this in a day unless you have any comments.

muga:
@dmikurube @smdmts Baiscally LGTM. sorry for the delay.

dmikurube:
Merged!

PR #659, opened 2017-05-31
embulk/guess/json.rb: JsonGuessPlugin ignores JsonParseException

muga:
This PR changes `JsonGuessPlugin` to ignore `JsonParseException` if it can parse even one JSON data. 

### Problem

`JsonGuessPlugin` guess sometimes fails because it gets `JsonParseException` when it cannot parse the last line of sampling buffer correctly. 

### Fix

To fix this issue, I changes the guess plugin to ignore `JsonParseException` if even one JSON data can be parsed. I first wanted to use `LineGuessPlugin` but, `JsonGuessPlugin` cannot extend it because `JsonGuessPlugin` should guess multi-line JSON.

### DIFF
--- a/lib/embulk/guess/json.rb
+++ b/lib/embulk/guess/json.rb
@@ -22,7 +22,8 @@ module Embulk
             # JSON object type check (isMapValue) is required for v. Because JsonParserPlugin
             # accepts only object type. And single column CSV avoids to be guessed as Json
             # parser type.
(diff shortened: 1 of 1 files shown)


muga:
@dmikurube please take a look? If you have better algorithm, please let me know.

muga:
Thank you for reviewing. I will merge it after CI passed.

muga:
Merged.

PR #661, opened 2017-06-01
Ignore JsonParseException in JSON preview if at least one JSON is already parsed.

dmikurube:
Similar to #659, it is required for JSON preview.


dmikurube:
@muga Thanks! Merging this PR for the time being.

PR #666, opened 2017-06-05
Support complex (mapping-style) PluginType.

dmikurube:
@muga It's WIP yet (no tests!), but can I have your impressions on this change?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
+++ b/embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
@@ -15,7 +15,6 @@ import org.embulk.config.ConfigException;
 import org.embulk.config.TaskSource;
 import org.embulk.config.ConfigDiff;
 import org.embulk.config.TaskReport;
-import org.embulk.plugin.DefaultPluginType;
 import org.embulk.plugin.PluginType;
 import org.embulk.spi.Schema;
 import org.embulk.spi.Exec;
@@ -513,7 +512,7 @@ public class BulkLoader
     private ExecutorPlugin newExecutorPlugin(BulkLoaderTask task)
     {
         return Exec.newPlugin(ExecutorPlugin.class,
-                task.getExecConfig().get(PluginType.class, "type", DefaultPluginType.create("local")));
+                task.getExecConfig().get(PluginType.class, "type", PluginType.LOCAL));
     }
 
     private ExecutionResult doRun(ConfigSource config)
--- a/embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
+++ b/embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
@@ -15,6 +15,7 @@ import org.embulk.config.ConfigException;
 import org.embulk.config.TaskSource;
 import org.embulk.config.ConfigDiff;
 import org.embulk.config.TaskReport;
+import org.embulk.plugin.DefaultPluginType;
 import org.embulk.plugin.PluginType;
 import org.embulk.spi.Schema;
 import org.embulk.spi.Exec;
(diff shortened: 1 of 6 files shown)


muga:
@dmikurube Agree. I also think that this's necessary for adding new PluginSource. 

I'm OK but, before I start reviewing the implementation, I want you to confirm that we include "Maven"PluginType in this PR? If so, we need to start talking about the spec. of the type option for MavenPluginSource here. Or split PRs later?

dmikurube:
@muga Thanks. Agreed, we need to discuss the spec before merging this.

I've sent a design doc about Maven plugin source to you. Let's discuss based on that later.

dmikurube:
@muga For the time being, removed the Maven support. The PR now includes only mapping-style configurations.

muga:
@dmikurube ok, thank you for fixing. I take a look.

dmikurube:
Jfyi: made one change. `PluginType.createFromString` is now `private`.

dmikurube:
@muga Oops, sorry, confirmed this does not actually work for the example below while tests pass.

```
in:
  type:
    name: file
```

Will fix.

muga:
@dmikurube good catch 👍 

I have one more question (I should've checked your design doc before but). In this PR, who can recognize (or detect) that input plugin type is default and output plugin type is maven like following? Or the detection or searching is out of scope of this PR?

```
in:
  type: file
  ... ...
out:
  type: {source: maven, name: s3, ...}
  ... ...
```

dmikurube:
@muga Thanks for your comment. I think `PluginType` is used in `PluginSource`s to load plugins. Then, for example from #592, recognizing in `PluginSource` may be reasonable like:

https://github.com/embulk/embulk/pull/592/files#diff-b1ebb6ca84d7cd5fe5762cf0ba80ef07R68

dmikurube:
We may need `Map<String, ToString>` or like that instead of `Map<String, String>`... `ToString` is in `org.embulk.spi.util`, though.

dmikurube:
Ah, no, a type can have only one `@JsonCreator`. Will fix.
https://stackoverflow.com/questions/15931082/how-to-deserialize-a-class-with-overloaded-constructors-using-jsoncreator

muga:
Agree. `PluginType` objects are used for loading and instantiating plugins in `PluginSource`. My question is who (or how) `PluginType` objects can be created in `BulkLoader` class in the case that multiple plugin types are specified in Embulk config?

dmikurube:
Ah, got your question. `PluginType` instances are always created via Jackson's `@JsonCreator`. (There's an error there as mentioned above, though.)

Jackson + YAML maps given JSON (YAML) into `PluginType` through their `ObjectMapper` via the `@JsonCreator` method.

muga:
@dmikurube Ah got it. Thank you for explaining. I didn't notice about it. It should work 👍

dmikurube:
@muga Finally it recognizes the JSON below with the fix dd8329d. Can you have a look?

```
in:
  type:
    name: file
```

dmikurube:
Added tests that actually parse configs.

embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
```diff
@@ -512,7 +512,7 @@ public void doCleanup(ConfigSource config, ResumeState resume)
     private ExecutorPlugin newExecutorPlugin(BulkLoaderTask task)
     {
         return Exec.newPlugin(ExecutorPlugin.class,
-                task.getExecConfig().get(PluginType.class, "type", new PluginType("local")));
+                task.getExecConfig().get(PluginType.class, "type", PluginType.LOCAL));
```
muga: Minor: Up to you. I'm not sure, we don't have major reasons so much to declare new `LOCAL` static field in `PluginType` class? `DefaultPluginType.create("local")` method call is good enough here instead?

dmikurube: One reason is that I wanted to keep the creator method (`JsonCreator`) `private` until we really need that `public`.

In addition, that "local" type is substantially a constant. Making the `PluginType` value constant seemed more reasonable than making the string `"local"` constant.

dmikurube:
@muga Thanks! Will be merging after considering whether include this in releasing 0.8.24 or not.

embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
muga: Understand. thank you for explaining.

dmikurube:
0.8.24 has been released. Merging this for 0.8.25.

Issue #665, opened 2017-06-01
Plan 0.8.24


dmikurube:
* [x] Make environment variables immutable in JRuby/Java #604
* [x] Add a note on the timestamp format in embulk-doc with a link to Ruby documents #648
* [x] Fix selfrun.bat so that selfrun.sh works correctly without error messages. #670
* [x] Add clean illegal characters mode to json parser. #651
* [x] embulk-core: Fix BulkLoader to pass file input/output plugins' task source to plugins' cleanup #663

dmikurube:
#604 is causing an error message as below while it's still working:

```
Error: Could not find or load main class org.jruby.Main
```

dmikurube:
Released.

PR #670, opened 2017-06-07
Fix selfrun.bat so that selfrun.sh works correctly without error messages.

dmikurube:
Fixing: https://github.com/embulk/embulk/issues/665#issuecomment-306423707

Can you have a look? @muga


dmikurube:
Thanks! Merging.

PR #678, opened 2017-06-15
Add ConfigInputPlugin for easier testing.

dmikurube:
When quick-testing in developing the core or plugins (later than input), it is often annoying to set up input. Prepare CSV on a filesystem, confirm the path, ...

An input which does not depend on external resources may help solving the situation. This is it. An input from the Embulk config itself. :)

```
in:
  type:
    name: config
  columns:
  - {name: id, type: long}
  - {name: name, type: string}
  values:
  - - [ 12, "foo" ]
    - [ 24, "bar" ]
  - - [ 98, "hoge" ]
    - [ 21, "fuga" ]
out:
  type: stdout
```

```
2017-06-16 17:34:41.041 +0900 [INFO] (0001:transaction): Using local thread executor with max_threads=8 / output tasks 4 = input tasks 2 * 2
2017-06-16 17:34:41.053 +0900 [INFO] (0001:transaction): {done:  0 / 2, running: 0}
98,hoge
12,foo
21,fuga
24,bar
2017-06-16 17:34:41.130 +0900 [INFO] (0001:transaction): {done:  2 / 2, running: 0}
2017-06-16 17:34:41.130 +0900 [INFO] (0001:transaction): {done:  2 / 2, running: 0}
2017-06-16 17:34:41.138 +0900 [INFO] (main): Committed.
2017-06-16 17:34:41.138 +0900 [INFO] (main): Next config diff: {"in":{},"out":{}}
```

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/ConfigInputPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/ConfigInputPlugin.java
@@ -126,11 +126,13 @@ public class ConfigInputPlugin
                             final JsonNode value = rowValues.get(column.getIndex());
                             if (value == null || value.isNull()) {
                                 pageBuilder.setNull(column);
-                            } else {
+                            }
+                            else {
                                 try {
                                     pageBuilder.setTimestamp(column,
                                                              timestampParsers[column.getIndex()].parse(value.asText()));
-                                } catch (TimestampParseException ex) {
+                                }
+                                catch (TimestampParseException ex) {
                                     throw new DataException(ex);
                                 }
                             }
@@ -141,10 +143,12 @@ public class ConfigInputPlugin
                             final JsonNode value = rowValues.get(column.getIndex());
                             if (value == null || value.isNull()) {
                                 pageBuilder.setNull(column);
-                            } else {
+                            }
+                            else {
                                 try {
                                     pageBuilder.setJson(column, jsonParser.parse(value.toString()));
-                                } catch (JsonParseException ex) {
+                                }
+                                catch (JsonParseException ex) {
                                     throw new DataException(ex);
                                 }
                             }
--- a/embulk-standards/src/main/java/org/embulk/standards/ConfigInputPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/ConfigInputPlugin.java
@@ -77,12 +77,9 @@ public class ConfigInputPlugin
         try (final PageBuilder pageBuilder = new PageBuilder(Exec.getBufferAllocator(), schema, output)) {
             for (final List<JsonNode> rowValues : taskValues) {
                 schema.visitColumns(new ColumnVisitor() {
-                        private int index = 0;
-
                         public void booleanColumn(Column column)
                         {
-                            final JsonNode value = rowValues.get(index);
-                            ++index;
+                            final JsonNode value = rowValues.get(column.getIndex());
                             if (value == null || value.isNull()) {
                                 pageBuilder.setNull(column);
                             }
@@ -93,8 +90,7 @@ public class ConfigInputPlugin
 
                         public void longColumn(Column column)
                         {
-                            final JsonNode value = rowValues.get(index);
-                            ++index;
+                            final JsonNode value = rowValues.get(column.getIndex());
                             if (value == null || value.isNull()) {
                                 pageBuilder.setNull(column);
                             }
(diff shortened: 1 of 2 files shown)


dmikurube:
@muga Can I have your thoughts? I often wanted this kind of preset plugins...

civitaspo:
[Just Share]
This plugin has the similar function > https://github.com/toyama0919/embulk-input-inline

dmikurube:
Ah, thanks for sharing! I was not sure about that plugin.

Let me share the reason why I wrote this as a preset standard plugin: it's because I often wanted this kind of plugins during core development. Installing external plugins was not very convenient when testing the core itself...

civitaspo:
I see! I thought that core committers might have a policy that does not increase plugins having similar functions like the embulk-parser-jsonl, so I shared.
Actually I also thought that it was better for Embulk Core to have the function of this plugin. So this pull request is greatly appreciated.

muga:
@dmikurube @civitaspo This feature would be helpful for Embulk development. In particular, it would be great if Embulk core could be separately tested without any plugins installation. Plugins installation and distribution ways might be changed. 

> a policy that does not increase plugins having similar functions like embulk-parser-jsonl, so I shared.

Make sense but, I'm still not sure what type of a policy or document we need to have as Embulk public site? cc: @hiroyuki-sato

hiroyuki-sato:
> In particular, it would be great if Embulk core could be separately tested without any plugins installation.

I think so too. 

It would be great if this plugin developer can also use this extension in the near future. 
The Embulk community is a small community. It is better to cooperate with each other.
Many plugin developers are looking for how to test my plugin. This document(How to test your plugin) may help all of the plugin developers. 
(I'll create another ticket about this).

dmikurube:
@muga Yeah, I think it's okay to start reviewing, while we may want to implement another config-style in maps (not arrays) like embulk-input-inline.

Users need to change their configurations anyways, though, because one difference from embulk-input-inline is that this ConfigInputPlugin takes control of  task separation.

@hiroyuki-sato Thanks! Yeah, we need documents for this.

muga:
@hiroyuki-sato thank you. What do you think about @civitaspo's first thought?

@dmikurube Ok, will take a look.

embulk-standards/src/main/java/org/embulk/standards/ConfigInputPlugin.java
```diff
@@ -0,0 +1,174 @@
...
+        final JsonParser jsonParser = new JsonParser();
+
+        try (final PageBuilder pageBuilder = new PageBuilder(Exec.getBufferAllocator(), schema, output)) {
+            for (final List<JsonNode> rowValues : taskValues) {
+                schema.visitColumns(new ColumnVisitor() {
+                        private int index = 0;
+
+                        public void booleanColumn(Column column)
+                        {
+                            final JsonNode value = rowValues.get(index);
```
dmikurube: Ah, exactly. Thanks for the good catch. Fixed them.

muga:
Started rebuilding the failed CI jobs.

hiroyuki-sato:
@muga 

> Actually I also thought that it was better for Embulk Core to have the function of this plugin. So this pull request is greatly appreciated.

Agree!. And I hope plugin developer can use this plugin for their plugin test in the near future. 

We shouldn't develop another plugin which has similar functions as much as we can. 
If the embulk-input-inline has useful feature, I hope to merge that in this plugin and maintain it with him(Mr. toyama0919)

dmikurube:
@hiroyuki-sato Thanks for adding your comment. (Cc: @muga)

AFAIU, major differences from `embulk-input-inline` are:

1. `embulk-input-inline` is written in Ruby while `ConfigInputPlugin` is in Java.
2. `embulk-input-inline` accepts data in mapping while `ConfigInputPlugin` accepts in an array.
3. `embulk-input-inline` always sends data in one task while `ConfigInputPlugin` controls tasks.

If you're fine, I'd merge this PR soon.

2 is worth importing from `embulk-input-inline`. At first I thought to implement 2 in this PR, but it'd be okay to add later.

dmikurube:
(In addition, I actually thought `inline` was really a better name. I couldn't rename `ConfigInputPlugin` to `inline` since we should avoid naming conflicts... ;) )

muga:
Hm. As we should solve to avoid name conflicts later, if you strongly want to use the name, we could negotiate and ask him to transfer it but..

dmikurube:
@muga I didn't mean I seriously want to rename... No problems. :)

dmikurube:
Thanks for taking a look. Merging this.

PR #684, opened 2017-06-19
Add MavenPluginSource to load JAR-based plugins into Embulk.

dmikurube:
@muga While it's yet WIP, can you have a quick look at the current implementation?

* `MavenPluginSource`: Not yet done. Almost just a copy from #592.
* `MavenArtifactFinder`: New, almost done and tested. It looks for local Maven artifact in the local file system.
* `TestMavenArtifactFinder`: Tests for `MavenArtifactFinder`. It looks for a dummy Maven artifact in a dummy Maven repository at `embulk-core/src/test/resources/m2.test`.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
+++ b/embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
@@ -97,11 +97,6 @@ public class MavenPluginSource
             throw new PluginSourceNotMatchException(ex);
         }
 
-        if (!pluginInterface.isAssignableFrom(pluginMainClass)) {
-            throw new PluginSourceNotMatchException(
-                "Plugin class \"" + pluginMainClass.getName() + "\" is not a valid " + category + " plugin.");
-        }
-
         final Object pluginMainObject;
         try {
             // Unlike InjectedPluginSource and JRubyPluginSource,
@@ -128,6 +123,10 @@ public class MavenPluginSource
                 }
                 pluginMainObject = new FileOutputRunner(fileOutputPluginMainObject);
             } else {
+                if (!pluginInterface.isAssignableFrom(pluginMainClass)) {
+                    throw new PluginSourceNotMatchException(
+                        "Plugin class \"" + pluginMainClass.getName() + "\" is not a valid " + category + " plugin.");
+                }
                 pluginMainObject = this.injector.getInstance(pluginMainClass);
             }
         }
--- a/embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
+++ b/embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
@@ -5,7 +5,6 @@ import com.google.inject.Injector;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import org.embulk.plugin.MavenPluginType;
-import org.embulk.plugin.PluginClassLoader;
 import org.embulk.plugin.PluginClassLoaderFactory;
 import org.embulk.plugin.PluginSource;
 import org.embulk.plugin.PluginSourceNotMatchException;
--- a/embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
+++ b/embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
@@ -1,0 +1,143 @@
+package org.embulk.plugin.maven;
+
+import com.google.inject.Inject;
+import com.google.inject.Injector;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.embulk.plugin.MavenPluginType;
+import org.embulk.plugin.PluginClassLoader;
+import org.embulk.plugin.PluginClassLoaderFactory;
+import org.embulk.plugin.PluginSource;
+import org.embulk.plugin.PluginSourceNotMatchException;
+import org.embulk.plugin.PluginType;
+import org.embulk.plugin.jar.JarPluginLoader;
+import org.embulk.plugin.jar.InvalidJarPluginException;
+import org.embulk.spi.DecoderPlugin;
+import org.embulk.spi.EncoderPlugin;
+import org.embulk.spi.ExecutorPlugin;
+import org.embulk.spi.FilterPlugin;
+import org.embulk.spi.FormatterPlugin;
+import org.embulk.spi.GuessPlugin;
+import org.embulk.spi.InputPlugin;
+import org.embulk.spi.OutputPlugin;
+import org.embulk.spi.ParserPlugin;
+
+public class MavenPluginSource
+        implements PluginSource
+{
+    @Inject
+    public MavenPluginSource(Injector injector)
+    {
+        this.injector = injector;
+    }
+
+    @Override
+    public <T> T newPlugin(Class<T> pluginInterface, PluginType pluginType)
+            throws PluginSourceNotMatchException
+    {
+        final String category;
+        if (InputPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "input";
+        } else if (OutputPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "output";
+        } else if (ParserPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "parser";
+        } else if (FormatterPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "formatter";
+        } else if (DecoderPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "decoder";
+        } else if (EncoderPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "encoder";
+        } else if (FilterPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "filter";
+        } else if (GuessPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "guess";
+        } else if (ExecutorPlugin.class.isAssignableFrom(pluginInterface)) {
+            category = "executor";
+        } else {
+            // unsupported plugin category
+            throw new PluginSourceNotMatchException("Plugin interface " + pluginInterface + " is not supported.");
+        }
+
+        if (pluginType.getSourceType() != PluginSource.Type.MAVEN) {
+            throw new PluginSourceNotMatchException();
+        }
+        final MavenPluginType mavenPluginType = (MavenPluginType) pluginType;
+
+        final PluginClassLoaderFactory pluginClassLoaderFactory =
+            this.injector.getInstance(PluginClassLoaderFactory.class);
+
+        final MavenArtifactFinder mavenArtifactFinder;
+        try {
+            mavenArtifactFinder = MavenArtifactFinder.create(getLocalMavenRepository());
+        }
+        catch (MavenRepositoryNotFoundException ex) {
+            throw new PluginSourceNotMatchException(ex);
+        }
+
+        final Path jarPath;
+        try {
+            jarPath = mavenArtifactFinder.findMavenArtifactJar(
+                mavenPluginType.getGroup(),
+                "embulk-" + category + "-" + mavenPluginType.getName(),
+                mavenPluginType.getVersion());
+        }
+        catch (MavenArtifactNotFoundException ex) {
+            throw new PluginSourceNotMatchException(ex);
+        }
+
+        final Class pluginMainClass;
+        try (JarPluginLoader loader = JarPluginLoader.load(jarPath, pluginClassLoaderFactory)) {
+            pluginMainClass = loader.getPluginMainClass();
+        }
+        catch (InvalidJarPluginException ex) {
+            throw new PluginSourceNotMatchException(ex);
+        }
+
+        if (!pluginInterface.isAssignableFrom(pluginMainClass)) {
+            throw new PluginSourceNotMatchException(
+                "Invalid plugin: \"" + pluginMainClass.getName() + "\" is not " + category + ".");
+        }
+        // TODO(dmikurube): Inject for each plugin as described in plugin.properties.
+
+        final Object pluginMainObject;
+        try {
+            pluginMainObject = pluginMainClass.newInstance();
+        }
+        catch (InstantiationException ex) {
+            throw new PluginSourceNotMatchException(
+                "Invalid plugin: \"" + pluginMainClass.getName() + "\" not instantiatable.", ex);
+        }
+        catch (IllegalAccessException ex) {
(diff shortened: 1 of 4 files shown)


muga:
@dmikurube Thank you for sharing. The architecture and design are good to me. I just wanted to put my comments here.

- Agree the usage of manifest file and `Embulk-Plugin-Main-Class` as how to find plugin main class.
- Maybe next commit. I'm interested in how `MavenPluginSource` can use `Finder.getMavenPath`.

dmikurube:
@muga Thanks.

* I'm still wondering between `MANIFEST.MF` or `ServiceLoader`. For later simplicity, I'd at least add something like "plugin API version" in `MANIFEST.MF`. `Embulk-Plugin-Main-Class` may be fine for the beginning.
* `getMavenPath` is actually a method in tests. In the main code, this will be `~/.embulk/m2` or from an environment variable.

dmikurube:
Rebased. It's now based on #686.

Not buildable yet.

dmikurube:
Rebased again. It's now on #687.

Order: #686 ==> #687 ==> This (#684; WIP)

dmikurube:
This version is somewhat working.

Tried with: https://github.com/dmikurube/embulk-input-dummyjar/tree/6851dab4c1d832e7b2bb5d13fe7327969c3d5196

```
in:
  type:
    source: maven
    group: org.embulk.input.dummyjar
    name: dummyjar
    version: 0.0.2
out:
  type: stdout
```

```
$ ls -R1 ~/.embulk/m2
repository

~/.embulk/m2/repository:
org

~/.embulk/m2/repository/org:
embulk

~/.embulk/m2/repository/org/embulk:
input

~/.embulk/m2/repository/org/embulk/input:
dummyjar

~/.embulk/m2/repository/org/embulk/input/dummyjar:
embulk-input-dummyjar

~/.embulk/m2/repository/org/embulk/input/dummyjar/embulk-input-dummyjar:
0.0.2

~/.embulk/m2/repository/org/embulk/input/dummyjar/embulk-input-dummyjar/0.0.2:
embulk-input-dummyjar-0.0.2.jar
embulk-input-dummyjar-0.0.2.jar.sha1
embulk-input-dummyjar-0.0.2.pom
embulk-input-dummyjar-0.0.2.pom.sha1
```

```
$ jar tvf ~/.embulk/m2/repository/org/embulk/input/dummyjar/embulk-input-dummyjar/0.0.2/embulk-input-dummyjar-0.0.2.jar
     0 Wed Jun 21 18:58:04 JST 2017 META-INF/
   247 Wed Jun 21 18:46:16 JST 2017 META-INF/MANIFEST.MF
     0 Wed Apr 05 17:06:58 JST 2017 org/
     0 Wed Apr 05 17:06:58 JST 2017 org/embulk/
     0 Wed Apr 05 17:06:58 JST 2017 org/embulk/input/
     0 Wed Jun 21 18:58:04 JST 2017 org/embulk/input/dummyjar/
   281 Wed Jun 21 18:58:04 JST 2017 org/embulk/input/dummyjar/DummyjarInputPlugin$PluginTask.class
  4024 Wed Jun 21 18:58:04 JST 2017 org/embulk/input/dummyjar/DummyjarInputPlugin.class
   112 Wed Jun 21 18:58:04 JST 2017 plugin.properties
```

```
$ unzip -q -c ~/.embulk/m2/repository/org/embulk/input/dummyjar/embulk-input-dummyjar/0.0.2/embulk-input-dummyjar-0.0.2.jar META-INF/MANIFEST.MF
Manifest-Version: 1.0
Implementation-Title: embulk-input-dummyjar
Implementation-Version: 0.0.2
Embulk-Plugin-Spi-Version: 0
Embulk-Plugin-Main-Class: org.embulk.input.dummyjar.DummyjarInputPlugi
 n
Embulk-Plugin-Security-Policy: runtime

```

(`Embulk-Plugin-Security-Policy` and `plugin.properties` are not used.)

muga:
Great!!

dmikurube:
@muga After a small fix ebbc464, the PR is now ready to review. Can you have a look?

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
```diff
@@ -0,0 +1,168 @@
+package org.embulk.plugin.maven;
+
+import com.google.inject.Inject;
+import com.google.inject.Injector;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.embulk.plugin.MavenPluginType;
+import org.embulk.plugin.PluginClassLoader;
```
muga: minor: It is not used so, we can remove it.

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        if (!pluginInterface.isAssignableFrom(pluginMainClass)) {
+            throw new PluginSourceNotMatchException(
+                "Plugin class \"" + pluginMainClass.getName() + "\" is not a valid " + category + " plugin.");
+        }
+
+        final Object pluginMainObject;
+        try {
+            pluginMainObject = pluginMainClass.newInstance();
```
muga: Since all plugins don't have non-argument constructor, it's better to use `injector.newInstance`. And I think that it needs to wrap the instance with `FileInputRunner` or `FileOutputRunner` if the plugin is based on file based plugin. It seems that `InjectedPluginSource.registerPluginTo` method implementation can be helpful.

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    private Path getLocalMavenRepository()
+            throws PluginSourceNotMatchException
+    {
+        String env;
+        try {
+            env = System.getenv("M2_REPO");
+        }
+        catch (NullPointerException | SecurityException ex) {
```
muga: minor: since non-null value is explicitly passed to `System.getenv` so, `NullPointerException` never happens?:-)

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
dmikurube: Done.

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
dmikurube: Ah, exactly. But let me keep this not to forget in case we make it a kind of variables in the future. (E.g. from multiple environment variables, or else.)

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
dmikurube: Ah, got it. Fixing it with checking other PluginSources.

dmikurube: @muga 
1. Can I see example(s) of non-argument constructors? I'm not very sure when non-argument constructors are used.
2. I understand the "arguments" are passed through Guice, right?

muga: Good point. Yes, "arguments" needs to be passed though Guice. We might not have any document to explain that. That's no good. 
* https://github.com/embulk/embulk/blob/master/embulk-core/src/main/java/org/embulk/exec/LocalExecutorPlugin.java#L42
* https://github.com/embulk/embulk-executor-mapreduce/blob/master/embulk-executor-mapreduce/src/main/java/org/embulk/executor/mapreduce/MapReduceExecutor.java#L70

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
muga: ok, understand.

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
dmikurube: @muga Thanks. Okay, executor plugins want SystemConfig... Got it.

Btw, for `FileInputRunner` or `FileOutputRunner`, is it really a role of `newPlugin` ? Looking in `InjectedPluginSource`, `InjectedPluginSource#newPlugin` does not do that. `StandardPluginModule` do that separately. 

The same in `JRubyPluginSource`. `JRubyPluginSource#newPlugin` only loads with the given type as well (`/lib/embulk/plugin.rb`:`new_java_*`), and file-based ones are wrapped in "registration" which is called at plugin's `lib/embulk/*.rb` such as https://github.com/embulk/embulk-input-s3/blob/master/embulk-input-s3/lib/embulk/input/s3.rb.

Hmm, where is should be "register"ed...?

muga: I'm sorry to your confused. I don't think that `MavenPluginSource#newPlugin` might not need to register a plugin factory`. But, I think that it might need to create plugin instances by `PluginFactory` or similar factory. Then, it might wrap them by `FileInputRunner` or `FileOutputRunner` if needed. How do you think about them?

dmikurube: @muga Hmm, my question meant:

* `InjectedPluginSource#newPlugin` returns raw `FileInputPlugin` or `FileOutputPlugin`, and wrapping is done in another location.
* If we do that wrapping in `MavenPluginSource#newPlugin`, `MavenPluginSource#newPlugin` returns wrapped `FileInputRunner` or `FileOutputRunner`.

Does not this asynchronicity cause any problems?

dmikurube: Ah, I might be misunderstanding. May `InjectedPluginSource#newPlugin" be returning `FileInputRunner` (`InputPlugin`) which has already been converted and "register"ed?

I firstly thought the calling order is `newPlugin` => some `register`, but is it actually some `register` (with conversion to `InputPlugin`) => `newPlugin`?

muga: I was confusing.. We should've checked `JRubyPluginSource` instead of `InjectedPluginSource`. Because `MavenPluginSource` is similar to `JRubyPluginSource`. I confirmed how Embulk loads S3InputPlugin in Embulk. 

1. lib/embulk/runner.rb calls `EmbulkEmbed#run`.
2. calls `BulkLoader#run`, and `doRun`.
3. calls `Exec#newPlugin`, and `PluginManager#newPlugin`.
4. calls `JRubyPluginSource#newPlugin`
5. calls `lib/embulk/plugin.rb`, Ruby's `PluginManager#new_java_input`.
6. calls `lib/embulk/plugin_registry.rb`'s `lookup`, `search`, and `require_and_show`.
7. calls `lib/embulk/input/s3.rb`'s `JavaPlugin#register_input`. 
8. calls `lib/embulk/plugin.rb` `register_java_input`, `register_java_plugin`.
9. calls `found.last.from_java(klass)`, file_input_plugin.rb's `from_java`.
10. calls `@registries[category].register(type, adapted)` to register it.
11. After `lookup` method call, plugin.rb's `new_java_input` calls `new_java` to create `FileInputRunner`.

muga: According to the behavior of plugin loading and registering by `JRubyPluginSource`, `MavenPluginSource` also needs to 
1. load plugins
2. create plugin instances after loading. If the plugin is file basis, wrap `FileInputRunner` or `FileOutputRunner`.

dmikurube: Yeah, I now find that those `register` methods are:

* For `InjectedPluginSource` called from the initial Guice module:
    1. instantiating the plugin class,
    2. wrapping with `File*Runner` if needed, and
    3. registering for `InjectedPluginSource.PluginFactory` in  Guice so that `InjectedPluginSource#newPlugin` can find the plugin instance.
* For `JRubyPluginSource` called from `JRubyPluginSource#newPlugin`:
    1. instantiating the plugin class,
    2. wrapping with `File*Runner` if needed, and
    3. registering for `Embulk::PluginRegistry` in Ruby so that `JRubyPluginSource#newPlugin` can find the plugin instance.

In analogy with them, Maven-based plugin instances may need to be registered somewhere so that `MavenPluginSource#newPlugin` can find the instance.

But in case of `JRubyPluginSource`, that `register` is called from `newPlugin` to be used in `newPlugin`... "Registration" is done at the same time with `newPlugin`. Okay, therefore, it'd be okay to instantiate and wrap them with `File*Runner` in `MavenPluginSource#newPlugin`.

Sorry about that. I was also confused. Will be implementing that part...

dmikurube:
@muga Finally, fixed for https://github.com/embulk/embulk/pull/684#discussion_r123640201 at https://github.com/embulk/embulk/pull/684/commits/6ba77edb7b9731a659f228324f72deab5c1ab71b. Can you have another look?

embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginSource.java
```diff
@@ -0,0 +1,187 @@
...
+
+        if (!pluginInterface.isAssignableFrom(pluginMainClass)) {
+            throw new PluginSourceNotMatchException(
+                "Plugin class \"" + pluginMainClass.getName() + "\" is not a valid " + category + " plugin.");
+        }
+
+        final Object pluginMainObject;
+        try {
+            // Unlike InjectedPluginSource and JRubyPluginSource,
+            // MavenPluginSource does not have "registration" before creating an instance of the plugin class.
```
muga: One question: Why did not you register a plugin factory? I think that the registration is useful as cache. For example, `s3` output plugin is possibly instantiated as both of actual output plugin and logging plugin. The cache might enable reducing time of a plugin load (search and create).

dmikurube: @muga 
> For example, s3 output plugin is possibly instantiated as both of actual output plugin and logging plugin.

I don't think so. AFAIR, the logging plugin using the s3 plugin may contain the s3 plugin in itself, or at least, loaded separately from the original s3 plugin. Before all, plugins are loaded separately with separate ClassLoaders, right?

muga: Ah I see, right. Thank you for explaining.

dmikurube:
Thanks! Merging this finally.

PR #685, opened 2017-06-20
[Doc] Fix invalid json data

hiroyuki-sato:
Fix invalid JSON data in the built-in document. 
It's inconvenient to test the Embulk.


hiroyuki-sato:
@muga @dmikurube, Please take a look when you can.

PR #686, opened 2017-06-20
Add MavenArtifactFinder to find a plugin Maven artifact from a local Maven repository.

dmikurube:
@muga Finally, I've split this part from #684. Itself is ready to review and merge -- can you have a look?


embulk-core/src/main/java/org/embulk/plugin/maven/MavenArtifactFinder.java
```diff
@@ -0,0 +1,122 @@
...
+        final Path absolutePath;
+        try {
+            absolutePath = localMavenRepositoryPath.normalize().toAbsolutePath();
+        }
+        catch (IOError ex) {
+            throw new MavenRepositoryNotFoundException(localMavenRepositoryPath, ex);
+        }
+        catch (SecurityException ex) {
+            throw new MavenRepositoryNotFoundException(localMavenRepositoryPath, ex);
+        }
```
muga: minor: Cannot we catch handle both of `IOExrror` and `SecurityException` in a single catch statement? like `catch (IOError | SecurityException) {`

dmikurube: @muga Ah, at first I thought to change their Exception messages while they're the same at the moment.

I'm keeping this as-is for the time being, and later considering changing messages.

muga: Ok, understand.

dmikurube:
Thanks! Merging.

PR #687, opened 2017-06-20
Add JarPluginLoader to load a JAR-based plugin class.

dmikurube:
@muga Next to #686, implementing a class that actually loads a JAR-based plugin class. #684 will be after this PR.

Order: #686 ==> This (#687) ==> #684

PR #688, opened 2017-06-21
Pass Gradle's System Properties (-D) starting with "org.embulk" to testing JVM.

dmikurube:
@muga A small change in `build.gradle` only impacting tests.

It passes System Properties of Gradle (can be specified with `-D`) which start with "org.embulk" into testing JVM. It allows developers to test like:

```
./gradlew -Dorg.embulk.someTestingRuntimeParameter=foobar test
```

so that the test can switch its behavior at runtime. One typical usage is removing temporary files. Temporary files are to be removed in usual cases. But, we sometimes want to keep temporary files to analyze when we run tests locally.

https://github.com/embulk/embulk/pull/687/files#diff-de1fed31d83d9cbde4b42260d2e2a8f5R48 is an example.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -100,6 +100,9 @@ subprojects {
         tasks.withType(JavaCompile) {
             options.compilerArgs << "-Xlint:unchecked" //<< "-Xlint:deprecation"
         }
+        tasks.withType(Test) {
+            systemProperties System.properties.findAll { it.key.startsWith("org.embulk") }
+        }
         tasks.withType(FindBugs) {
             reports {
                 xml.enabled = false


muga:
@dmikurube I have one question. How often you want to change the system property? If he values of the system properties are changed every `./gradlew` execution time, this approach makes sense. If not every time, can we take an ad-hoc approach like declaring the system property and the value in build.gradle directly like https://github.com/embulk/embulk-input-jdbc/blob/master/build.gradle#L107. I'm still thinking that this approach is a little bit tricky for me.

dmikurube:
@muga Hmm, it's not often, and that is the reason why I made it like this.

For example in case of https://github.com/embulk/embulk/pull/687/files#diff-de1fed31d83d9cbde4b42260d2e2a8f5R48, we usually (in-CI and manually) run this without the system property. We sometimes want to change to debug locally.

You mean we should change `build.gradle` locally when we need such a customization? I think it's not very good because 1) changing `build.gradle` ad-hoc may impact on unexpected parts, and 2) we may merge that ad-hoc change by mistake. It is better to customize local tests without local edits.

If we write the properties directly in `build.gradle` like that JDBC example, we cannot customize at runtime. The purpose here is to customize sometimes.

muga:
@dmikurube The PR looks good to me as talked directly about it to you. Can you file a ticket about why Embulk's tests failed when all system properties were passed to the Embulk?

dmikurube:
Thanks! Filed #692, and merging this PR.

Issue #640, opened 2017-05-18
Hide Embulk's own dependencies from plugins


dmikurube:
**NOTE that nothing has been determined. Just pushing ideas here. Not ready for discussion.**

~~For licensing and to hide dependency libraries from plugins, we **may** want to package dependency JARs as-is in the Embulk JAR, not extracting dependencies as a Fat JAR.~~

Focusing on hiding dependencies from plugins in this issue.

Options? :
* http://one-jar.sourceforge.net/
  * This may solve #650 at the same time.
* Relocating dependencies by Gradle shadow plugin: http://imperceptiblethoughts.com/shadow/#relocating_packages

dmikurube:
Heard about relocating dependencies by Gradle shadow plugin. It is very very nice in terms that we don't need to use another `ClassLoader`, but we cannot use this because some plugins have been already depending on Embulk's own dependencies.

A possible plan is:
1. Use One-JAR (or else) to provide options to depend or not-depend on Embulk's own dependencies. Plugins can choose by themselves by a kind of embedded meta data.
2. Wait a while for (all) plugins to migrate as a deprecation period.
3. Drop One-JAR and switch to relocating.

dmikurube:
An experiment (out of Embulk): https://github.com/dmikurube/clean-pluggable-example

dmikurube:
It's mostly done through v0.10!

PR #690, opened 2017-06-22
Relocate recently-added dependency packages in shadowJar Embulk distribution.

dmikurube:
@muga To reduce future risks of dependency conflicts between the core and plugins like #640, this PR tries relocating embedded dependency packages. Can you have a look?

For example, `org.apache.maven*` is relocated to `org.embulk.third_party.org.apache.maven*` in embulk-0.8.XX.jar. All references to these classes are automatically relocated as well.

It's based on the relocation feature in the Gradle Shadow Plugin:
http://imperceptiblethoughts.com/shadow/#relocating_packages

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -227,6 +227,18 @@ project(':embulk-cli') {
     apply plugin: 'com.github.johnrengelman.shadow'
 
     shadowJar {
+        // 'org.apache.maven' is used in:
+        // - embulk-cli: EmbulkMigrate and EmbulkSelfUpdate (from v0.8.17)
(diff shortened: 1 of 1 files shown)


dmikurube:
Thanks!

muga:
@dmikurube I knew that you already merged this PR. I'm trying to fix #694 and noticed that these relocation might not match at our parent_first_{packages,resources}.properties. We need to consider about a better way.

dmikurube:
@muga Some problems may be possibly thought out, but I'm not sure what problems you're wondering... Can you elaborate specific problems to discuss (maybe in another issue)?

PR #696, opened 2017-06-26
Refactor: make InjectedPluginSource.PluginFactory private.

dmikurube:
@muga Just refactoring. It is unnecessarily `public`. PTAL?


dmikurube:
Thanks!

PR #701, opened 2017-06-29
Support "classifier" in Maven-based plugin configurations.

dmikurube:
@muga As chatted directly, we're supporting "classifier" in Maven configurations.


dmikurube:
Thanks! Merging.

PR #703, opened 2017-06-30
Fix typo: modifieid -> modified.

hiroyuki-sato:
@dmikurube Please take a look when you can.


dmikurube:
Found that the message was originally from the Ruby version... ;)

PR #706, opened 2017-07-05
Use JRuby's ScriptingContainer for bootstrap instead of org.jruby.Main.main

dmikurube:
@muga Can you have a look? Details are in #707.

### Context

Embulk is calling org.jruby.Main.main in org.embulk.cli.Main. It is not very good in terms of JRuby's suggestion and JRuby environment isolation -- which "JRuby instance" does `org.jruby.Main.main` use?

In the process of un-JRuby-ing, let's replace it with `ScriptingContainer` which is the standard way to call JRuby embedded from Java.

### DIFF
--- a/embulk-cli/src/main/java/org/embulk/cli/Main.java
+++ b/embulk-cli/src/main/java/org/embulk/cli/Main.java
@@ -1,37 +1,84 @@
 package org.embulk.cli;
 
-import java.net.URISyntaxException;
+import java.util.ArrayList;
+import org.jruby.RubyInstanceConfig;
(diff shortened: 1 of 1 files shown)


dmikurube:
<0.8.26 + this PR> is released as 0.8.27-TESTING1 at https://bintray.com/embulk/maven/embulk-unstable/0.8.27-TESTING1?sort=&order=#files.

dmikurube:
`-R--dev` should be supported at least: #707

dmikurube:
Thanks! Merging.

PR #708, opened 2017-07-05
Set up $LOAD_PATH (and possibly Bundler) from Java.

dmikurube:
Converting `lib/embulk/command/embulk_bundle.rb` to Java. It's to be merged after #706.

What do you think? @muga


dmikurube:
@muga Thanks. Jfyi, fixed `bin/embulk` to catch up with this change.

For testing, can we consider after all related changes are merged, or before releasing 0.8.28? That is because :

1. This part mainly manipulates the "global" JRuby instance. Testing that will be complicated.
2. These parts are going to be changed again and again before releasing 0.8.28, such as #543 and more. If we add tests with this PR, the test may need to catch up every time.

muga:
> Can you consider about adding tests for this feature?

@dmikurube Sorry to your confused. The testing that I wanted to say is not for this feature. I wanted you to suggest integration tests for avoiding regression. Because you've been changing these parts a lot. Simple test is fine for me. The test scenario is based on how Embulk is used. like 1) install plugins by bundler 2) execute Embulk 3) confirm that ingested data is correct or not. How about this?

dmikurube:
@muga Ah, got it. I do the similar processes in my local environments, but I agree that they should be finally automated, of course.

But, I think it's not the time yet. There are 3-4 more changes are expected around CLI and the bootstrap. Such tests would be added for the final state -- what do you think about this plan?

muga:
@dmikurube I see. It's good for me. LGTM.

dmikurube:
Thanks! Merging.

PR #719, opened 2017-07-09
Fix Javadoc in embulk-jruby-strptime classes to avoid Javadoc errors

muga:
This PR fixes Javadoc comments in embulk-jruby-strptime classes to avoid Javadoc errors as following. Since Embulk release task requires their Javadoc generation, we cannot release new version caused by Javadoc errors. To fix this, It's good to use href tags for external links in `@see` section. 

```
embulk/embulk-jruby-strptime/src/main/java/org/embulk/spi/time/RubyDateParser.java:33: error: unexpected text
     * @see https://github.com/jruby/jruby/blob/036ce39f0476d4bd718e23e64caff36bb50b8dbc/lib/ruby/stdlib/date/format.rb
       ^
```


muga:

muga:
Thank you for your comment. I didn't notice that it works with JDK7. Yes, I did use JDK8 to generate the Javadoc. 

We don't need to merge for v0.8.27 by JDK7 but, I will include this minor fix in v0.8.27 and release it.

dmikurube:
Yes, Javadoc requirements may be different betweek JDK 7 and 8... We may need big fixes when moving to JDK 8. ;)

PR #724, opened 2017-07-15
Fix Javadoc in PluginClassLoader to avoid Javadoc error by Oracle Java8

muga:
This PR fixes Javadoc comments in `PluginClassLoader` to avoid Javadoc error by Oracle Java8. Because Travis CI uses Oracle java8 so, it cannot generate Embulk's Javadoc documents by the following Javadoc error. I think, that might be why embulk.org is not updated.
https://github.com/embulk/embulk/compare/v0.8.25...v0.8.26

```
embulk/embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java:87: error: unexpected text
     * @see https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html#loadClass(java.lang.String,%20boolean)
       ^
embulk/embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java:88: error: unexpected text
     * @see http://hg.openjdk.java.net/jdk7u/jdk7u/jdk/file/jdk7u141-b02/src/share/classes/java/lang/ClassLoader.java
       ^
```

### DIFF
--- a/embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
+++ b/embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
@@ -49,7 +49,7 @@ public class PluginClassLoader
      *
      * Some plugins (embulk-input-jdbc, for example) are calling this method to load external JAR files.
      *
-     * @see https://github.com/embulk/embulk-input-jdbc/blob/ebfff0b249d507fc730c87e08b56e6aa492060ca/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java#L586-L595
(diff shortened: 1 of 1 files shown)


muga:

muga:
The above Javadoc errors are shown by `./gradlew site` command by Java8. (I tried to execute several commands declared in embulk-docs/push-gh-pages.sh on local.)

Since this PR fixes all Javadoc errors by `./gradlew site`, the command will succeed after merging this. But, the command also shows 20 Javadoc warnings. Do you think that it's better to fix warnings as well on this ticket?

And what do you think that Travis CI checks `./gradlew site` works or not before executing `push-gh-pages.sh` script?

muga:
Thank you for reviewing. Merging.

PR #725, opened 2017-07-18
Refactor setting the global JRuby ScriptingContainer up.

dmikurube:
@muga Just refactoring out (extracting) the creator of `ScriptingContainer`. Can you have a look?


dmikurube:
Hmm, failing in downloading Handle.zip from Sysinternals? Might be throttled?

muga:
Not sure but, Appveyor's geting stable.

dmikurube:
Hm, thanks! Merging.

PR #728, opened 2017-07-19
Cache handle.exe in AppVeyor.

dmikurube:
@muga The `Handle.zip` problem is still happening on AppVeyor. I'm trying to cache the unzipped `handle.exe` (downloaded directly from live.sysinternals.com) in AppVeyor. Can you have a look?


dmikurube:
Thanks. Merging.

PR #731, opened 2017-07-19
Replace Ruby Embulk::EmbulkRunner into Java org.embulk.EmbulkRunner.

dmikurube:
@muga Finally! We don't need `*.rb` files during the bootstrap now. (Several `runScriptlet` is still required, though.)

Can you have a look?

### DIFF
--- a/embulk-cli/src/main/java/org/embulk/cli/EmbulkRun.java
+++ b/embulk-cli/src/main/java/org/embulk/cli/EmbulkRun.java
@@ -13,6 +13,8 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import org.embulk.EmbulkEmbed;
+import org.embulk.EmbulkRunner;
 import org.embulk.EmbulkSetup;
 import org.embulk.cli.parse.EmbulkCommandLineHelpRequired;
 import org.embulk.cli.parse.EmbulkCommandLineParseException;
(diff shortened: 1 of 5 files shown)


dmikurube:
@muga Updated: some options (e.g. `-o`) were not working correctly.

dmikurube:
@muga Updated the logging from `EmbulkRunner` to use slf4j.

dmikurube:
jfyi, I'd release v0.8.28 when this PR is merged. Please ping me if you have concerns.

muga:
I started taking a look.

embulk-cli/src/main/java/org/embulk/cli/EmbulkRun.java
```diff
@@ -13,6 +13,8 @@
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import org.embulk.EmbulkEmbed;
```
dmikurube: Okay, will remove.

dmikurube:
@muga Thanks for taking a look.

The main reason why I located `EmbulkRunner` in `embulk-core` was that I just mechanically converted `lib/embulk/command/*` into `embulk-cli`, and `lib/embulk/*` into `embulk-core`. `EmbulkRunner` is a conversion from `lib/embulk/runner.rb`.

Another reason was that `EmbulkRunner` was Ruby `Embulk::Runner` which looked like a kind of "core" object throughout Embulk.

It's okay to move it to `embulk-cli` now, but I think we will anyways optimize and restructure things later gradually. Since these are big changes, I believe simple mechanical conversion works fine. Optimizing would be done in later changes seeing actual usages.

To make it clear, I've put the comment below there.
```
 * NOTE: Developers should not depend on this EmbulkRunner class. This class is created tentatively, and may be
 * re-implemented again in a different style.
```

embulk-cli/src/main/java/org/embulk/cli/EmbulkRun.java
dmikurube: Removed in 92976c6.

muga:
@dmikurube Thank you for explaining. I'm also OK to keep them in embulk-core now. Will refactor and restructure them guradually.

dmikurube:
Thanks! Merging, and releasing v0.8.28.

PR #736, opened 2017-07-21
modified the listFiles() function 

icaiyu:
Using the ArrayList instead of ImmutableList for the files so that we can sort them before set them to the task.
To load the files exactly in the correct order. The max_threads should equal 1. Using multi threads to load the files is impossible to load them in the correct order.

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
@@ -69,6 +69,8 @@ public class LocalFileInputPlugin
 
         // list files recursively
         List<String> files = listFiles(task);
+        // Sort the files before we set them to the task.
+        Collections.sort(files);
         log.info("Loading files {}", files);
         task.setFiles(files);
 
@@ -97,7 +99,7 @@ public class LocalFileInputPlugin
             }
         } else {
             List<String> files = new ArrayList<String>(task.getFiles());
-            Collections.sort(files);
+            //Need not to sort again
             configDiff.set("last_path", files.get(files.size() - 1));
         }
 
@@ -124,7 +126,8 @@ public class LocalFileInputPlugin
             directory = (d == null ? CURRENT_DIR : d);
         }
 
-        final ImmutableList.Builder<String> builder = ImmutableList.builder();
+        //Using the ArrayList to record the files instead of ImmutableList
+        final List<String> filesArray = new ArrayList<String>();
         final String lastPath = task.getLastPath().orNull();
         try {
             log.info("Listing local files at directory '{}' filtering filename by prefix '{}'", directory.equals(CURRENT_DIR) ? "." : directory.toString(), fileNamePrefix);
@@ -186,11 +189,11 @@ public class LocalFileInputPlugin
                         }
                         if (parent.equals(directory)) {
                             if (path.getFileName().toString().startsWith(fileNamePrefix)) {
-                                builder.add(path.toString());
+                                filesArray.add(path.toString());
                                 return FileVisitResult.CONTINUE;
                             }
                         } else {
-                            builder.add(path.toString());
+                            filesArray.add(path.toString());
                         }
                         return FileVisitResult.CONTINUE;
                     }
@@ -199,7 +202,7 @@ public class LocalFileInputPlugin
         } catch (IOException ex) {
             throw new RuntimeException(String.format("Failed get a list of local files at '%s'", directory), ex);
         }
-        return builder.build();
+        return filesArray;
     }
 
     @Override


dmikurube:
@icaiyu Thanks for your contribution.

But why do you think an alphabetical order is the "correct" order? It needs to be consistent per run (to resume), but it does not need to be an alphabetical order actually. AFAIU, the current ordering is at least consistent (per file system or else) even if it may not be in alphabetical order. It may be nice to have an alphabetical order just for usability, but it's not a "problem".

I think we cannot accept this PR as-is because it may break compatibility of "resumed" users. Consider that a user has used `LocalFileInput` with the current Embulk version and resumed with some files done. The config diff is generated with the current version. Imagine the user upgrades Embulk to a new version (with this PR merged), and resume with the config diff generated by the older Embulk version. The behavior can be totally "broken" -- some files may be uploaded twice, and some files may be never uploaded. It's a bigger problem than just a non-alphabetical ordering.

An acceptable option may be to have a new configuration option like `sort: alphabetical` or like that not to break existing behaviors.

What do you think?

icaiyu:
@dmikurube 
Thank you for your reply, but I am getting more confussing about the Resume function of Embulk now.
Say that I have 9 files:
sample_01.csv, sample_02.csv, sample_03.csv...sample_09.csv
I want to upload them to the server using the LocalInputPlugin
After embulk upload 4 files, what if the network is broken? what if I press the Ctrl+C the stop the upload, Could the embulk record the number the sample files which are not uploaded yet?
In this case, it should record sample_04.csv in the diff.
So that I could resume the upload from the sample_05.csv.

dmikurube:
@icaiyu Can you share that the real example what you observed? As I've mentioned above, it may not be alphabetical, but I think it'd not be a problem in resuming.

hiroyuki-sato:
@icaiyu 
I hope this comment solve your question. 

## Summaries. 

* The LocalFileInput load data in parallel.  (number of files = tasks)
  * [Slideshare](https://www.slideshare.net/frsyuki/embuk-internals/2)
* So, those data don't always load in alphabetical order. 

## How to resume?

Let's explain about `resume` behavior. 

You can download all of the sample files from [here](https://github.com/hiroyuki-sato/embulk-support/tree/master/core_736_resume_test).

### Input Data

The following example uses three files. 

```
hoge/csv/sample_01.csv
hoge/csv/sample_02.csv <-- this fie has invalid record
hoge/csv/sample_03.csv
```

Line 3 in `sample_02.csv` has invalid record.
The `aaa` isn't a valid number.

```
aaa,14824,2015-01-27 19:01:23,20150127,embulk jruby
```

### run

So `embulk run` fail. 

`embulk run -r resume.yml config.yml`

```
[INFO] Writing resume state to 'resume.yml'
[INFO] Resume state is written. Run the transaction again with -r option to resume or use "cleanup" subcommand to delete intermediate data.
java.lang.RuntimeException: org.embulk.spi.DataException: Invalid record at line 3: aaa,14824,2015-01-27 19:01:23,20150127,embulk jruby
	at org.embulk.EmbulkRunner.runInternal(EmbulkRunner.java:360)
	at org.embulk.EmbulkRunner.run(EmbulkRunner.java:173)
	at org.embulk.cli.EmbulkRun.runSubcommand(EmbulkRun.java:475)
	at org.embulk.cli.EmbulkRun.run(EmbulkRun.java:99)
	at org.embulk.cli.EmbulkBundle.checkBundleWithEmbulkVersion(EmbulkBundle.java:42)
	at org.embulk.cli.EmbulkBundle.checkBundle(EmbulkBundle.java:15)
	at org.embulk.cli.Main.main(Main.java:26)
```

### resume file.

After fail `embulk run`, Embulk generate `resume.yml` file
In this case, LocalFileInput load `sample01.csv` and `sample03.csv` correctly, 
But, `sample02.csv` doesn't load yet. 

The `null` records in `in_reports` part mean "That task not complete yet".

```
in_task:
  FileInputTaskSource:
    Files: [hoge/csv/sample_01.csv, hoge/csv/sample_02.csv, hoge/csv/sample_03.csv]
in_reports:
- {}   # Task1
- null # Task2 <-- this data not load completely, 2nd is hoge/csv/sample_02.csv
- {}   # Task3 
out_reports:
- {}
- {}
- null
- null
- {}
- {}
```

dmikurube:
@icaiyu Apart from this issue itself, we've found that the `-r` option was not working very well. It turned out that very very few people have actually tried using the resume feature... including ourselves as well.

We'll be discussing it at #740. Everything is TBD, but our future options may include removing the resume feature.

If you could share your practical and real use-cases/requirements for the resume feature, those may be very helpful for us to decide, and maybe to re-design. Thanks!

icaiyu:
@hiroyuki-sato 
The -r option doesn't generate the resume.yml.
I try the 
embulk run config.yml -o resume.yml

## My sample
sample_01.csv.gz, sample_02.csv.gz, sample_03.csv.gz, ... , sample_07.csv.gz 
In the sample_04_csv.gz it have an invalid record.
## My config.yml
                                                                                                                                        
in:                                                                                                                                     
  type: file                                                                                                                            
  path_prefix: /home/cy/data/try1/csv/sample_                                                                                           
  decoders:                                                                                                                             
  - {type: gzip}                                                                                                                        
  parser:                                                                                                                               
    charset: UTF-8                                                                                                                      
    newline: LF                                                                                                                         
    type: csv                                                                                                                           
    delimiter: ','                                                                                                                      
    quote: '"'                                                                                                                          
    escape: '"'                                                                                                                         
    null_string: 'NULL'                                                                                                                 
    trim_if_not_quoted: false                                                                                                           
    skip_header_lines: 1                                                                                                                
    allow_extra_columns: false                                                                                                          
    allow_optional_columns: false                                                                                                       
    columns:                                                                                                                            
    - {name: id, type: long}                                                                                                            
    - {name: account, type: long}                                                                                                       
    - {name: time, type: timestamp, format: '%Y-%m-%d %H:%M:%S'}                                                                        
    - {name: purchase, type: timestamp, format: '%Y%m%d'}                                                                               
    - {name: comment, type: string}                                                                                                     
out: {type: stdout} 

## My run
 embulk run config.yml -o resume.yml

## Expect result
Since the sample_04.csv.gz have an invalid record.
I expect that embulk stop at the sample_04.csv.gz
Thus in the resume.yml, it should have
last_path: sample_04.csv.gz

so that I can fix the invalid record in sample_04.csv.gz
And then I can continue at the sample_04.csv.gz

## Actual result
                                                                                                                                        
in:                                                                                                                                     
  type: file                                                                                                                            
  path_prefix: /home/cy/data/try1/csv/sample_                                                                                           
  decoders:                                                                                                                             
  - {type: gzip}                                                                                                                        
  parser:                                                                                                                               
    charset: UTF-8                                                                                                                      
    newline: LF                                                                                                                         
    type: csv                                                                                                                           
    delimiter: ','                                                                                                                      
    quote: '"'                                                                                                                          
    escape: '"'                                                                                                                         
    null_string: 'NULL'                                                                                                                 
    trim_if_not_quoted: false                                                                                                           
    skip_header_lines: 1                                                                                                                
    allow_extra_columns: false                                                                                                          
    allow_optional_columns: false                                                                                                       
    columns:                                                                                                                            
    - {name: id, type: long}                                                                                                            
    - {name: account, type: long}                                                                                                       
    - {name: time, type: timestamp, format: '%Y-%m-%d %H:%M:%S'}                                                                        
    - {name: purchase, type: timestamp, format: '%Y%m%d'}                                                                               
    - {name: comment, type: string}                                                                                                     
  last_path: /home/cy/data/try1/csv/sample_07.csv.gz                                                                                    
out: {type: stdout}

hiroyuki-sato:
@icaiyu 

* Add `stop_on_invalid_record: true` in your config.yml
* Exeucte `embulk run -r resume.yml config.yml`

Your `embulk run` will fail and generate `resume.yml`.

You have to set `stop_on_invalid_record: true`. Otherwise, embulk skip an invalid record.
https://github.com/hiroyuki-sato/embulk-support/blob/master/core_736_resume_test/config.yml#L16

Your `embulk run` generate the following line. 

```
2017-07-24 10:33:23.458 +0200 [WARN] (0015:task-0005): Skipped line 3 (java.lang.NumberFormatException: For input string: "aaa"): aaa,14
```

The resume file generated if `eumbulk run` failed.

icaiyu:
@hiroyuki-sato 
Thank you, I got the resume.yml now. What if I fix the invalid record in the sample_04.csv.gz(For example, I just delete that record manually)
Could I run the Embulk again that it only parse the sample_04.csv.gz ?

hiroyuki-sato:
@icaiyu 

@dmikurube mentioned above, 
https://github.com/embulk/embulk/pull/736#issuecomment-317349995
**we've found that the -r option was not working very well.**

Probably, `-r` option does not work as you expected. 

That's why he asked you below.

> If you could share your practical and real use-cases/requirements for the resume feature, those may be very helpful for us to decide, and maybe to re-design. Thanks!

Maybe, you can't delete `sample_04.csv.gz`. you need `sample_04.csv.gz` file even it is empty. 
But I've never tested it before.

icaiyu:
@hiroyuki-sato 
I means in the sample_04.csv there are many lines of record, only some of them are invalid, after I fix those invalid record, I hope that the embulk could continue to deal the sample_04.csv, and then the sample_05.csv, sample_06.csv ....

In my case , the order is very important, the embulk deal the file one by one
sample_01.csv     OK. upload it to the server.
sample_02.csv     OK. upload it to the server.
sample_03.csv     OK. upload it to the server.
sample_04.csv     Oh, there are some invalid records in this file! 

The embulk should stop and record the 
"last_path: sample_04.csv" in the resume.yml

Then I fix invalid data in the sample_04.csv then I run the resume.yml, the embulk should start again.

sample_04.csv     OK. upload it to the server.
sample_05.csv     OK. upload it to the server.
..
sample_09.csv     OK. upload it to the server.

It seems that the embulk doesn't work as I want.  I am trying to solve this problem in my plugin. Maybe I can write the certain resume() function in my plugin.
Anyway, Thank you for your reply, they are really helpful for me to understand how the embulk work.

hiroyuki-sato:
If you load those files one by one, I recommend you below.
It is the simplest way.  

* Put data files into queue directory.
* Create `data` load directory
* Save the following file as `load.sh`
* Execute `load.sh`
  * It moves one file from the queue to data directory.
  * Execute embulk. It loads the single file. (sample01.csv)
  * Embulk save diff.yml as sample01.csv loaded.
  * It moves one file from queue to data directory. (sample02.csv)
  * Execute embulk. It loads the single file. (sample02.csv), because embulk knows sample01.csv already loaded.
* You simply modify error file(sample04.csv) if load.sh stopped. 
  

```bash
#!/bin/bash

while [ $( ls queue | wc -l ) -ne 0 ] ; do 
  ls -1 queue/sample* | head -1 | xargs -I % mv % datas
  embulk run conf.yml -c diff.yml  >> embulk.log 2>&1 
  ret=$?
  if [ $ret -ne 0 ] ; then
    echo "Failed: $ret" >> embulk.log 
    exit 1
  fi
  echo ""
done
```



```yaml
in:
  type: file
  path_prefix: datas/sample_
  # ...
```

PR #741, opened 2017-07-24
Fix the wrongly Ruby-style "#{ENV['GEM_HOME']}" in the general usage.

dmikurube:
@muga Sorry, I had a stupid mistake. Can you have a look?


dmikurube:
Thanks!

PR #742, opened 2017-07-24
Keep the "-b" or "--bundle" option in the argument list.

dmikurube:
@muga Sorry for the last-minute update for v0.8.28. This _will_ be required to delay initializing Bundler and to delay creating a JRuby instance.


dmikurube:
Thanks! Merging.

PR #744, opened 2017-07-28
Deprecate Timestamp-related methods using JRuby ScriptingContainer.

dmikurube:
@muga Deprecating them for the future. Can you have a look?

They have never been used actually although they're passed through constructors!


dmikurube:
Thanks! Merging, and releasing v0.8.29 tomorrow.

PR #747, opened 2017-07-28
Reduce usage of the global (Singleton) JRuby instance

dmikurube:
@muga As a part of avoiding JRuby instances in the bootstrap process, this PR tries reducing usage of the global (Singleton) JRuby instance. Can you have a look?

(It is not in a hurry. It would be a kind of continued efforts.)


dmikurube:
Thanks!

PR #749, opened 2017-07-28
Delay creating the global JRuby instance to org.embulk.EmbulkRunner.

dmikurube:
@muga Yay, we're close to the last step! Moving the creation of the singleton JRuby instance to `EmbulkRunner` in this PR.

It assumes #755 is merged beforehand.

After this PR, the real last step is to move the initial JRuby configuration code to `JRubyScriptingModule`. There are many barriers, though. (E.g. we need to take care of `EmbulkEmbed` users, we need a way to store the configuration values somewhere to pass `JRubyScriptingModule`, ...)


embulk-core/src/main/java/org/embulk/EmbulkGlobalJRubyScriptingContainer.java
```diff
@@ -114,6 +122,27 @@ public static ScriptingContainer setup(
 
     private static final class UnrecognizedJRubyOptionException extends Exception {}
     private static final class NotWorkingJRubyOptionException extends Exception {}
+    private static final class UnrecognizedJRubyLoadPathException extends Exception {
+        public UnrecognizedJRubyLoadPathException()
+        {
+            super();
+        }
```
muga: minor: don't you need to declare an unused constructor?

embulk-core/src/main/java/org/embulk/EmbulkGlobalJRubyScriptingContainer.java
```diff
@@ -114,6 +122,27 @@ public static ScriptingContainer setup(
...
+
+        public UnrecognizedJRubyLoadPathException(final String message)
+        {
+            super(message);
+        }
+
+        public UnrecognizedJRubyLoadPathException(final Throwable cause)
+        {
+            super(cause);
+        }
```
muga: minor: this is also not used.

embulk-core/src/main/java/org/embulk/EmbulkGlobalJRubyScriptingContainer.java
dmikurube: Exactly. Removed.

embulk-core/src/main/java/org/embulk/EmbulkGlobalJRubyScriptingContainer.java
dmikurube: Ditto. Thanks.

dmikurube:
Merging!

PR #755, opened 2017-08-02
Refactoring: Ruby EmbulkRunner to receive Java org.embulk.EmbulkRunner directly.

dmikurube:
@muga Just refactoring.


dmikurube:
Thanks! Merging.

PR #762, opened 2017-08-21
Avoid calling the deprecated constructor in org.embulk.spi.util.PagePrinter: Fix #760

dmikurube:
@muga Can you have a look?

Cc: @hiroyuki-sato

PR #763, opened 2017-08-21
Stop using old |TimestampFormatter.FormatterTask|

dmikurube:
In addition to #762...


dmikurube:
Changed the base, and merging.

PR #765, opened 2017-08-22
Initialize the global JRuby instance in JRubyScriptingModule except for Embulk::Runner

dmikurube:
@muga I think it's now ready for review and merge... Can you have a look when you have time?

The final goal of this PR is to delay instantiating JRuby's `ScriptingContainer` to `JRubyScriptingModule`, not in the bootstrap. But, I had one concern that some Java-Embulk users might be calling JRuby `Embulk::Runner` from Java. :(

Then, this PR moves almost all the initialization of `ScriptingContainer` to `JRubyScriptingModule`, but initializing `Embulk::Runner` remains in the bootstrap with some deprecation messages to warn.

A new version would be released just with this PR. Then, after waiting for feedbacks about `Embulk::Runner`, all the instantiation would be moved to `JRubyScriptingModule`.

This large entire PR consists of smaller commits. Each commit moves single features.


dmikurube:
@muga Can you have a look? If you'd like the PR split into smaller, I'd split this.

muga:
Sorry to be late. I started taking a look!

muga:
The change looks good to me. Let's release new one and get feedback.

dmikurube:
Thanks! Sure, merging and releasing a new version.

PR #777, opened 2017-08-31
Update gradle 4.1.0 (Plugin's and core)

hiroyuki-sato:
@dmikurube 

This PR Update Gradle (plugin's and core) to version 4.1. (2/3)
Please take a look when you have a time.


hiroyuki-sato:
@dmikurube 

Thank you for reviewing.
Separate this commit to #780 and #781.

Issue #775, opened 2017-08-30
JSON's Exponential Number could not parse.


smdmts:
I'm using JSON parser and when an included number of Exponential as long, it could not parse with skipping. (e.f 12345e3 just like 12,345,000)
But JSON's specification is supported a number of exponential notation. Therefore should it parse with built-in JSON parser plugin?

```
{ "double":-1.234e-5, "string":"string","long":12345e3, "boolean":true }
```

```
Skipped an invalid record (Failed to parse '1.2345E7' as long)
```

- Reproduction codes
  https://github.com/hiroyuki-sato/embulk-support/tree/master/test_json_double

- JSON's specs 
http://www.json.org/

Thanks,

dmikurube:
@smdmts Thanks for your report! (Sorry to be late.)

As we talked separately, it may be related with #634, and possible from [`JsonParser.java`](https://github.com/embulk/embulk/blob/v0.8.31/embulk-core/src/main/java/org/embulk/spi/json/JsonParser.java). We need a deeper look...

dmikurube:
Although I know it's a kind of bugs, I'm afraid "fixing" this may break some compatibilities. Depending on the investigation in `JsonParser.java`, we may delay fixing this until v0.9.

dmikurube:
@smdmts Had a first look. (Technical internals at last.)

Just curious, in what use-case do you need to map such an exponential representation into `long`, not `double`? Exponential representation is used for floating numbers in common, so it looked a bit strange configuration to me. 

I guess it should succeed if the value is mapped into `double` instead of `long`.

----

Technical internals:

The first cause is that Jackson's `com.fasterxml.jackson.core.JsonParser` recognizes `12345e3` as a token of `VALUE_NUMBER_FLOAT`. Embulk converts `VALUE_NUMBER_FLOAT` to msgpack `FloatValue` (`1.2345e7`) in Embulk's `org.embulk.spi.json.JsonParser`.

The direct cause looks that `PageReader` (or somewhere else in Embulk) fails to map the `FloatValue` (`1.2345e7`) into Embulk's `long`. It might be fixable without breaking a kind of compatibilities, but need a hack in depth...

smdmts:
@dmikurube Thank you for the comments, I explain my situation.

Some system output large number over the LONG(cap is Long.MAX_VALUE) using BigInteger in JSON's library. 
The JSON spec shows number type can express exponential number with over the LONG value.  

[...]

Issue #634, opened 2017-05-15
Misunderstanding a String config when specifying a numeric string with a dot


dmikurube:
Known issue? (Won't fix?)

```
in:
  type: file
  path_prefix: 1234567890.
  parser:
    type: csv
    columns:
    - {name: c0, type: string}
out:
  type: stdout
```

```
2017-05-15 20:22:17.083 +0900 [INFO] (0001:transaction): Listing local files at directory '.' filtering filename by prefix '1.23456789E9'
2017-05-15 20:22:17.089 +0900 [INFO] (0001:transaction): Loading files []
2017-05-15 20:22:17.178 +0900 [INFO] (0001:transaction): Using local thread executor with max_threads=8 / tasks=0
2017-05-15 20:22:17.186 +0900 [INFO] (main): Committed.
2017-05-15 20:22:17.186 +0900 [INFO] (main): Next config diff: {"in":{},"out":{}}
```

`1234567890.` is considered as `1.23456789E9`. Quoting like `"1234567890."` fixes. It may be specification of YAML?

hiroyuki-sato:
@dmikurube 

Possibly, it's a [YamlTagesolver](https://github.com/embulk/embulk/blob/master/embulk-core/src/main/java/org/embulk/config/YamlTagResolver.java) related Issue?

dmikurube:
Thanks for your help. Hmm, possibly. Will have a look there.

hiroyuki-sato:
@dmikurube 

I roughly checked [YamlTagResolver](https://github.com/embulk/embulk/blob/master/embulk-core/src/main/java/org/embulk/config/YamlTagResolver.java#L22) regex.

Please take a look when you can.


```diff
diff --git a/embulk-core/src/main/java/org/embulk/config/YamlTagResolver.java b/embulk-core/src/main/java/org/embulk/config/YamlTagResolver.java
index 4ba00f7e..cb5708a5 100644
--- a/embulk-core/src/main/java/org/embulk/config/YamlTagResolver.java
+++ b/embulk-core/src/main/java/org/embulk/config/YamlTagResolver.java
@@ -19,7 +19,7 @@ public class YamlTagResolver
     // events without using Composer.

     public static final Pattern FLOAT_EXCEPTING_ZERO_START = Pattern
[...]

PR #783, opened 2017-09-05
Test org.embulk.spi.json.JsonParser with TestJsonParser

dmikurube:
@muga Can you have a look?

See also: #775 #634


dmikurube:
Thanks! Merging.

Issue #778, opened 2017-09-01
Environment variables can not load by Gemfile since v0.8.29.


toyama0919:
Hello.
Now I am in trouble..

Up to 0.8.28, environment variables could be loaded in liquid by the following method.
But, since 0.8.29, environmental variables can not be read by the above method.

### My Gemfile

```Gemfile

source 'https://rubygems.org/'

gem 'embulk-input-...'
gem 'embulk-filter-...'
...

gem 'dotenv'

begin
  require 'dotenv'
  Dotenv.load(File.expand_path('../../.env', __FILE__))
rescue LoadError => e
rescue => e
end

```

[dotenv](https://github.com/bkeepers/dotenv) is a library that can manage environment variables with yaml.

### My .env

```.env
DATABASE_HOST=localhost
DATABASE_USER=username
DATABASE_PASS=password
DATABASE_NAME=some_database
```

### My liquid file

```
in:
  type: mysql
  host: {{ env.DATABASE_HOST }}
  user: {{ env.DATABASE_USER }}
  password: {{ env.DATABASE_PASS }}
  database: {{ env.DATABASE_NAME }}
  query: |
    select * from users limit 1

out:
  type: stdout

```

### Run embulk
```
$ embulk run -b embulk_bundle config.yml
```

Did something change around environment variables?

It is running Gemfile on things like Digdag's --params-file.
https://docs.digdag.io/command_reference.html

dmikurube:
@toyama0919 Thanks for your report! Got it... Since v0.8.29, JRuby not for plugins (e.g. subcommands, Liquid, Bundler, ...) is separate from others to avoid creating JRuby instances in the bootstrap.

Hmm..., let me think about that. That would be non-trivial to keep the un-JRuby'ed bootstrap.

dmikurube:
Ahh, anyways, Bundler **will** run after initialization (including Liquid) from the next version. Then, that kind of environment variable hacks may not work...

dmikurube:
https://github.com/embulk/embulk/pull/765

dmikurube:
@toyama0919 After some thoughts :
- At least from the new major version (v0.9 expected), setting environment variables in Gemfile will be no longer supported. It is hard to get together with un-JRuby.
- During v0.8.*, we may have a kind of workarounds, but I have no good ideas yet so far...

[...]

PR #790, opened 2017-09-07
Add a CLI option "-E" running an external script to configure environment variables

dmikurube:
@muga Can you have a look? I guess it (indirectly) solves #778...

Will be releasing v0.8.33 after merging this PR.

### DIFF
--- a/embulk-cli/src/main/sh/selfrun.sh
+++ b/embulk-cli/src/main/sh/selfrun.sh
@@ -8,7 +8,7 @@ while true; do
         -E*)
             external_script="${1#-E}"
             echo "Running an external shell script: $external_script"
-            source $external_script
+            . $external_script
             shift
             ;;
         "-J+O")
--- a/embulk-cli/src/main/sh/selfrun.sh
+++ b/embulk-cli/src/main/sh/selfrun.sh
@@ -5,6 +5,12 @@ overwrite_optimize=""
 
 while true; do
     case "$1" in
+        -E*)
+            external_script="${1#-E}"
+            echo "Running an external shell script: $external_script"
+            source $external_script
+            shift
+            ;;
         "-J+O")
             overwrite_optimize="true"
             shift
--- a/bin/embulk
+++ b/bin/embulk
@@ -48,6 +48,11 @@ overwrite_optimize = nil
 until ARGV.empty?
   v = ARGV[0]
   case v
+  when /-E(.*)/
+    external_script = v[2..-1]
+    STDERR.puts "Running an external Ruby script: #{external_script}"
+    load external_script
+    ARGV.shift
   when "-J+O"
     overwrite_optimize = true
     ARGV.shift
--- a/embulk-cli/src/main/bat/selfrun.bat
+++ b/embulk-cli/src/main/bat/selfrun.bat
@@ -63,8 +63,7 @@ if "%status%" == "rest" (
 
 ) else if "%arg:~0,2%" == "-E" (
     if not "%arg:~2%" == "" (
-        set external_script=%arg:~2%
-        echo "Running an external batch file: %external_script%"
+        echo Running an external batch file: %arg:~2%
         call %arg:~2%
     )
 
--- a/embulk-cli/src/main/bat/selfrun.bat
+++ b/embulk-cli/src/main/bat/selfrun.bat
@@ -61,6 +61,13 @@ if "%status%" == "rest" (
 ) else if "%status%" == "read" (
     call :read_file %arg%
 
+) else if "%arg:~0,2%" == "-E" (
+    if not "%arg:~2%" == "" (
+        set external_script=%arg:~2%
+        echo "Running an external batch file: %external_script%"
+        call %arg:~2%
+    )
+
 ) else if "%arg%" == "-J+O" (
     set overwrite_optimize=true
     set status=rest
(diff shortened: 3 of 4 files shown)


embulk-cli/src/main/sh/selfrun.sh
```diff
@@ -5,6 +5,12 @@ overwrite_optimize=""
 
 while true; do
     case "$1" in
+        -E*)
+            external_script="${1#-E}"
+            echo "Running an external shell script: $external_script"
+            source $external_script
```
hiroyuki-sato: nit  `/bin/dash` doesn't work `source`. I think it is better to use `.` instead of `source`.

```
$ source /tmp/test.sh
dash: 1: source: not found
```

```
$ . /tmp/test.sh
$
```

hiroyuki-sato:
@dmikurube 

I'm wondering is it possible to set environment variables via Ruby?
`export` in shell worked, but `ENV["HOGE"]` and `dotenv` in ruby script doesn't export environment variables.

```
cat /tmp/.env.test
DOTENV=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

```bash
#!/bin/sh

export HOGE='**************** HOGEHOGE **********************'

ruby -e '
 require "dotenv"
 ENV["HOGE"] = "***************************** FUGAFUGA ***************";
 Dotenv.load "/tmp/.env.test"

 ENV.each{ |k,v| puts "RUBY -- #{k} = #{v}" }
'
```



```
embulk -E/tmp/test.sh run -I /tmp/embulk-input-hoge/lib  /tmp/test.yml

RUBY -- HOGE = ***************************** FUGAFUGA ***************
RUBY -- DOTENV = AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

2017-09-07 20:59:42.051 +0900: Embulk v0.8.31
2017-09-07 20:59:46.612 +0900 [INFO] (0001:transaction): Loaded plugin embulk/input/hoge from a load path
HOGE = **************** HOGEHOGE **********************
...
2017-09-07 20:59:46.661 +0900 [INFO] (0001:transaction): Using local thread executor with max_threads=8 / output tasks 4 = input tasks 1 * 4
2017-09-07 20:59:46.676 +0900 [INFO] (0001:transaction): {done:  0 / 1, running: 0}
example-value,1,0.1
example-value,2,0.2
2017-09-07 20:59:46.767 +0900 [INFO] (0001:transaction): {done:  1 / 1, running: 0}
2017-09-07 20:59:46.779 +0900 [INFO] (main): Committed.
2017-09-07 20:59:46.779 +0900 [INFO] (main): Next config diff: {"in":{},"out":{}}
```

test.yml

```yaml
in:
  type: hoge
out:
  type: stdout
```

embulk-input-hoge

```ruby
module Embulk
  module Input

    class Hoge < InputPlugin
      Plugin.register_input("hoge", self)

      def self.transaction(config, &control)
        # configuration code:
        task = {
#          "option1" => config.param("option1", :integer),                     # integer, required
#          "option2" => config.param("option2", :string, default: "myvalue"),  # string, optional
#          "option3" => config.param("option3", :string, default: nil),        # string, optional
        }

ENV.each{ |k,v| puts "#{k} = #{v}" }
        columns = [
          Column.new(0, "example", :string),
          Column.new(1, "column", :long),
          Column.new(2, "value", :double),
        ]

        resume(task, columns, 1, &control)
      end

      def self.resume(task, columns, count, &control)
        task_reports = yield(task, columns, count)

        next_config_diff = {}
        return next_config_diff
      end

      # TODO
      # def self.guess(config)
      #   sample_records = [
      #     {"example"=>"a", "column"=>1, "value"=>0.1},
      #     {"example"=>"a", "column"=>2, "value"=>0.2},
      #   ]
      #   columns = Guess::SchemaGuess.from_hash_records(sample_records)
      #   return {"columns" => columns}
      # end

      def init
        # initialization code:
        @option1 = task["option1"]
        @option2 = task["option2"]
        @option3 = task["option3"]
      end

      def run
        page_builder.add(["example-value", 1, 0.1])
        page_builder.add(["example-value", 2, 0.2])
        page_builder.finish

        task_report = {}
        return task_report
      end
    end

  end
end
```

embulk-cli/src/main/sh/selfrun.sh
dmikurube: I thought `source` and `.` have no differences... Can you confirm the `.` usage of yours change your environment variables on the parent shell, or not? I guess that is not changing.

dmikurube:
@hiroyuki-sato As chatted directly, they're impossible due to "environment variable"'s nature.

dmikurube:
Thanks! Merging, and releasing v0.8.32 soon...

kakoni:
Late to the party, but if we assume -E would be used for env variables, would it make more sense to do something like
```bash
env_file="${1#-E}"
echo "Running an external env: $env_file"
env $(cat $env_file | xargs)
```
with -E option.

This way one could keep Procfile/.dotenv format and no need for extra sh scripts?

dmikurube:
@kakoni Thanks for your comments. It would be feasible (on Linux/Mac), but I think we won't adopt it from now since:

* It won't work the same for Windows. It'd be more consistent to use "the more basic scripting language" on the OS.
* We see some benefits on the idea, but not very big.
* The `-E` option has already been merged and released.

Issue #799, opened 2017-10-02
Use of deprecated org.embulk.spi.util.dynamic.AbstractDynamicColumnSetter#setRubyObject


dmikurube:
`org.embulk.spi.util.dynamic.AbstractDynamicColumnSetter` has had a method `setRubyObject` to implement its `set` for Ruby called from Ruby `Embulk::PageBuilder`. As a part of un-JRuby-ing Embulk, they are deprecated since v0.8.36. They are to be removed in v0.9.

It is expected that no plugins are using the method directly. If you find any plugin using the method, comments are welcome in this issue to report the plugins.

dmikurube:
They are finally removed in v0.9.0.

PR #800, opened 2017-10-02
Deprecate org.embulk.spi.util.dynamic.AbstractDynamicColumnSetter#setRubyObject

dmikurube:
@muga As a part of un-JRuby-ing, deprecating `AbstractDynamicColumnSetter#setRubyObject`. Can you have a look?

(I think this is used almost only from Ruby's `page_builder`. As its `set` is re-implemented in Ruby, I believe it should be fine...)


muga:
I also believe that the change works fine for both of Ruby and Java plugins. I saw that 3rd-party Java plugins've used them before (I don't remember those plugin names) but, the plugins might not have used `setRubyObject` method.

And It would be good to update #799 before v0.8.36 is released.

dmikurube:
Thanks! Merging this.

@muga What do you mean by "And it would be good to update #799 before v0.8.36 is released." ?

PR #803, opened 2017-10-02
Deprecate org.embulk.spi.json.RubyValueApi

dmikurube:
@muga Next to #800, deprecating `RubyValueApi` which is static methods to convert MessagePack `Value`s and Ruby objects.

I couldn't find OSS plugins calling these static methods at least by Google.

It expects #800 as `org.embulk.spi.util.dynamic.AbstractDynamicColumnSetter#setRubyObject` is calling `RubyValueApi`.


dmikurube:
Thanks. Merging.

PR #810, opened 2017-10-10
Fix typo in FALSE_STRINGS and add their test

takuti:
`N` is doubly defined. It should be `F`.


dmikurube:
@takuti Thanks! The fix looks correct... can you just confirm, @muga?

muga:
oh, good catch.. 👍

dmikurube:
Thanks. Merging...

PR #811, opened 2017-10-10
Fix typos in Javadoc in PluginClassLoader

dmikurube:
@muga It was blocking the embulk.org documents form updates. Can you have a look?


dmikurube:
Thanks. ;)

PR #812, opened 2017-10-13
Do not use numeric prefixes (like "%6N") in default date/time format specifiers in parsing

dmikurube:
@muga Can you have a look?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/util/DynamicColumnSetterFactory.java
+++ b/embulk-core/src/main/java/org/embulk/spi/util/DynamicColumnSetterFactory.java
@@ -74,7 +74,9 @@ public class DynamicColumnSetterFactory
         if (option != null) {
             return option.getTimestampFormat();
         } else {
-            return new TimestampFormat("%Y-%m-%d %H:%M:%S.%6N");
+            // DynamicColumnSetter is used for inputs, then datetime parsing.
+            // Ruby's strptime does not accept numeric prefixes in specifiers such as "%6N".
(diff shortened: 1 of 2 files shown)


muga:
@dmikurube Good catch and thank you for fixing.

This change might break backward compatible for `TimestampFormatter`. The default format is used for both of formatter and parser.

```ruby
# checked strftime behavior.
irb(main):065:0> t = Time.now
=> 2017-10-17 17:05:04 -0700
irb(main):066:0> t.strftime('%Y-%m-%d %H:%M:%S.%N')
=> "2017-10-17 17:05:04.836793000"
irb(main):067:0> t.strftime('%Y-%m-%d %H:%M:%S.%9N')
=> "2017-10-17 17:05:04.836793000"
irb(main):068:0> t.strftime('%Y-%m-%d %H:%M:%S.%6N')
=> "2017-10-17 17:05:04.836793"
```

I think that we need to introduce each default value for formatter and parser. What do you think?

dmikurube:
@muga As they are `DynamicColumnSetterFactory` and `DynamicPageBuilder`, I thought they're used only for parsing, not for formatting. (`%6N` still exists in `TimestampFormatter` and `TimestampType`.)

Can you show an example path where these default formats are used for formatting?

dmikurube:
@muga Sorry, `getTImestampFormat` was called within `DynamicColumnSetterFactory` even for `new TimestampFormatter`, as chatted f2f. 15a175c fixes this by making two private methods `getTimestampFormatForFormatter` and `getTimestampFormatForParser`.

dmikurube:
Thanks!

PR #813, opened 2017-10-13
Remove dead code around the Ruby timestamp parser replaced

dmikurube:
@muga Removing the dead code replaced by embulk-jruby-strptime. Can you have a look?


dmikurube:
Thanks. Merging!

PR #814, opened 2017-10-16
Enable setting timestamp formats from Columns' in Ruby plugins

dmikurube:
@muga Trying to make Ruby plugins more flexible... Not yet sure this is the right design, though.

Can I have your opinions?


dmikurube:
Ah, conflicted... Rebasing, and then merging.

dmikurube:
Rebased. Will be merging after confirming CI gets happy. Thanks!

dmikurube:
Merging!

PR #816, opened 2017-10-16
Notify Embulk-announce mailing list in CLI

dmikurube:
@muga @sakama We have no other channels to notify existing users about the mailing list... :( (It's a chicken-and-egg!)

The notification would be tentative until we believe the mailing list gets well-known to existing users.

What do you think?


dmikurube:
Thanks. Merging this for v0.8.36...

PR #817, opened 2017-10-17
Fix typo of a parameter name in Javadoc in PluginClassLoader

dmikurube:
@muga Sorry, it had another typo...


dmikurube:
Thanks!

PR #818, opened 2017-10-17
Workaround against the rubygems.lasagna.io problem in embulk-docs/

dmikurube:
Uploading embulk-docs/ to embulk.org is still failing because of that old lasagna.org problem... https://travis-ci.org/embulk/embulk/jobs/288866021#L1137


muga:
Looks good 👍

PR #819, opened 2017-10-17
Locate embedded dependency gems under gems/ in Embulk .jar file

dmikurube:
Embedded dependency gems (`msgpack-java`, `liquid`, and `bundler` for the time being) are located under `gems/` and `specifications/` in the embulk-core jar file.

```
META-INF/
  ...
org/
  embulk/
    *.class
    ...
gems/
  bundler-1.10.6/
    ...
  liquid-4.0.0/
    ...
  msgpack-java-1.1.0/
     ...
specifications/
  bundler-1.10.6.gemspec
  msgpack-java-1.1.0.gemspec
  liquid-4.0.0.gemspec
```

With the location, developers and users can load embulk "like gem" in JRuby by

```
require `./embulk-core-....jar`
require `msgpack`
```

Eventually, Embulk's Ruby code would be located like this as well.

@sakama Can you have a look? (Cc: @muga)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -269,10 +269,10 @@ project(':embulk-cli') {
 }
 
 task rubyTestVanilla(type: JRubyExec, dependsOn: "classpath") {
-    jrubyArgs '-Ilib', '-Itest', '-rtest/unit', '--debug', './test/vanilla/run-test.rb'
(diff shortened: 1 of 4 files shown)


dmikurube:
Thanks!

PR #821, opened 2017-10-18
Deprecate org.embulk.spi.time.Timestamp's getRubyTime and fromRubyTime

dmikurube:
@muga I guess it'd be the final deprecation around `org.jruby`. Can you have a look?


dmikurube:
Ah..., no, there are some more deprecations for Timestamps.

dmikurube:
Thanks. Merging!

PR #822, opened 2017-10-18
Revert "Merge pull request #813 from embulk/remove-dead-code-in-old-datetime-parser"

dmikurube:
@muga Let me revert this for the time being.

I found that we got errors like below after merging #813 :

```
org.jruby.exceptions.RaiseException: (NoMethodError) undefined method `parse' for Time:Class
```
```
org.jruby.exceptions.RaiseException: (NoMethodError) undefined method `strptime' for Time:Class
```

It looks like that just replacing `require 'embulk/java/time_helper'` to `require 'time'` would fix the problem. The replacement fixed my cases at least. But, let me postpone this to the next next release (expected v0.8.37).

v0.8.36 already has many changes, then I don't want to include suspicious changes in v0.8.36.


dmikurube:
Thanks. Merging (reverting) this.

PR #825, opened 2017-10-19
Deprecate constructor of org.embulk.spi.util.PagePrinter with TimestampFormatter.FormatterTask

dmikurube:
@muga Can you have a look?

`org.embulk.spi.util.PagePrinter` has been implemented with deprecated `TimestampFormatter.FormatterTask`... The constructor is also to be deprecated.


dmikurube:
Thanks.

PR #826, opened 2017-10-19
Deprecate newFormatter with TimestampFormatter.FormatterTask and newParser with TimestampParser.ParserTask in TimestampFormat

dmikurube:
`TimestampFormat`'s `newFormatter` and `newParser` have been implemented with `TimestampFormatter.FormatterTask` and `TimestampParser.ParserTask` which have been deprecated for a long time.


dmikurube:
@muga Thanks!

But, sorry that I realized that we don't need to add the alternative methods `newFormatter(DateTimeZone)` and `newParser(DateTimeZone)`. `TimestampFormatter` and `TimestampParser` can be created directly. Adding new methods may introduce other dependencies...

I've removed them in abef66c. Can you have another quick look?

dmikurube:
Thanks! Merging.

PR #829, opened 2017-10-20
Modify notification message on Embulk-announce mailing list

dmikurube:
@muga @sakama Updating the message introduced in #816 to have more explicit notion on compatibility-breaking changes. Can you have a look?


dmikurube:
Thanks!

Issue #796, opened 2017-09-19
CSV Parser does not support full Ruby datetime format


mjalkio:
According to the documentation at http://www.embulk.org/docs/built-in.html#csv-parser-plugin "The Timestamp format refers to Ruby strftime format."

However, I am trying to parse datetimes in this format `2017-09-19 13:39:15.139396-07:00` (this is default string formatting from Python) using `%Y-%m-%d %H:%M:%S.%6N%Z`, and it does not work. I believe that `%6N` is not supported.

I believe you should add additional documentation to explain that you're not actually using the Ruby methods directly and you've built your own parser (https://github.com/embulk/embulk/blob/master/embulk-jruby-strptime/src/main/java/org/embulk/spi/time/StrptimeParser.java)

dmikurube:
@mjalkio Thanks for your feedback! Will have a deeper look. Cc: @muga

dmikurube:
Sorry to be late. Let me leave my view. TL;DR, we should just update the document like you mentioned, @mjalkio.

The original Ruby has had different format rules for formatting and parsing. `Time.strftime` accepts numeric prefix like `%6N` while `Time.strptime` does not accept. It was discussed at: https://bugs.ruby-lang.org/issues/11220 (in Japanese, though)

We're (and should be) actually following original Ruby's behavior for formatting and parsing. The description "The Timestamp format refers to Ruby strftime format." is really wrong in terms of parsing. We should just mention it follows `strptime`'s format, and it is different from `strftime`'s.

Embulk's `StrptimeParser` is trying to emulate the original Ruby `Time.strptime`. If it is not working like `Time.strptime`, it'd be recognized as a bug to be fixed.

@muga Any more thoughts do you have?

mjalkio:
Makes sense, I pushed a PR with the proposed change to the language.

PR #831, opened 2017-10-20
Update timestamp format documentation

mjalkio:
Proposed change from discussion in #796.


dmikurube:
@mjalkio Thanks for your contribution! But sorry, let us confirm how it should be described (with taking a deeper look at the actual behavior and the implementation) by ourselves at first.

dmikurube:
The document is to be fixed along with #835 and some more cleanups, fixes, and tests around timestamp formats.

Issue #830, opened 2017-10-20
Avoid using org.jruby.util.RubyDateFormat in TimestampFormatter


dmikurube:
1. Embulk shouldn't be impacted by JRuby's future changes.
2. Embulk should be able to run without JRuby.
3. `org.jruby.util.RubyDateFormat` has been deprecated in JRuby! https://github.com/jruby/jruby/blob/9.1.13.0/core/src/main/java/org/jruby/util/RubyDateFormat.java#L50

PR #835, opened 2017-10-25
Move embulk-jruby-strptime into embulk-core with refactoring, making licenses clearer, and giving reasonable visibilities

dmikurube:
@muga @sakama Before addressing #860, #846, and #830, let me refactor `embulk-jruby-strptime` to make it easier to modify. Can you have a look?


dmikurube:
@sakama @muga Thanks!

@muga Yeah, I tried to divide, but the original code has had much mutual dependencies. Once refactoring one place, other places needed to change as well... Some more refactoring changes (which could be divided) are in the waiting list, actually. ;)

PR #836, opened 2017-10-26
Split rubyTest into vanilla tests and tests with _strptime monkey patch

dmikurube:
@muga For adding more tests of `TimestampParser` even in the Ruby world, let me split `rubyTest` into two parts at first.

* `test/vanilla` is long-lived tests without any monkey patch.
* `test/monkey_strptime` is recently added MRI's test imported with monkey patching `Date._strptime`.


dmikurube:
Thanks! Merging.

PR #837, opened 2017-10-26
Add Ruby tests with TimestampParser to compare with Ruby's strptime

dmikurube:
@muga Adding tests for `TimestampParser`, not just our Ruby-compatible time parser (jruby-strptime), by comparing the result epoch time with Ruby's `strptime`.

It's now ready. Can you have a look?

Jfyi, the test has commented-out cases where `TimestampParser` cannot parse `sec` == 60 (leap second) correctly. We may want to fix it later, but I think it should be acceptable for the time being.


dmikurube:
Thanks!

Issue #834, opened 2017-10-24
Use non-mDay values as well to build Timestamp


dmikurube:
At the same time with #833, I see that only `mDay` may be used to build `Timestamp`. Other day-values (`yday`) may not be used to build `Timestamp`.

* https://github.com/embulk/embulk/blob/v0.8.36/embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java#L229-L245

@muga Can you check it? Like #833, I'd fix it through the refactoring if it's true.

muga:
Oh `yday` may not be used. I double-check CRuby and JRuby's implementation but, `yday` is not used from other properties. `rubyTest` would be broken but, we could refactor and remove unnecessary properties for Embulk timestamp.

dmikurube:
@muga Hmm, what do you mean? I mean that `"2017 271 12:34:56"` might be parsed well into `Timestamp` with `"%Y %j %H:%M:%S"` when it was parsed with JRuby, but currently may not.

I'm seeing CRuby parses that, and converts to the epoch time which is used to build `Timestamp`.

```
irb(main):001:0> require 'date'
=> true
irb(main):002:0> DateTime.strptime("2017 271 12:34:56", "%Y %j %H:%M:%S").to_time.to_i
=> 1506602096
```

PR #838, opened 2017-10-26
Add TestTimestampParser, direct tests for org.embulk.spi.time.TimestampParser

dmikurube:
@muga Can you have a look? It adds tests for `TimestampParser` in the Java world.

Some test cases are disabled because of #834 -- `%j` is not considered in `TimestampParser`.


dmikurube:
Thanks!

PR #840, opened 2017-10-26
Parse Ruby-style time zone names in addition: Fix #833

### LINKED ISSUE #833: Use Ruby's timezone names instead of Joda-Time's
While I was walking-through and trying to refactor our Ruby-compatible date/time parser, I'm afraid I found that we may not using Ruby's timezone names (`TimeZoneConverter`), but using Joda-Time's.

* https://github.com/embulk/embulk/blob/v0.8.36/embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java#L188
* https://github.com/embulk/embulk/blob/v0.8.36/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormat.java#L58-L88

@muga Can you check it? If it's true, I'd fix it throughout my refactoring.

dmikurube:
@muga @sakama Can you have a look? It fixes #833.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormat.java
+++ b/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormat.java
@@ -57,6 +57,12 @@ public class TimestampFormat
 
     public static DateTimeZone parseDateTimeZone(String s)
     {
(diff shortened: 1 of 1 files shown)


dmikurube:
Thanks!

PR #841, opened 2017-11-10
Update use cases link

ksss:
http://www.embulk.org/docs/recipe/scheduled-csv-load-to-elasticsearch-kibana4.html is 404.


dmikurube:
@ksss Sorry to be late. Let us merge this... Thanks for your contribution!

PR #842, opened 2017-11-10
Fix parsing leap seconds to get the same Timestamp with pure Ruby's

dmikurube:
@muga Fixing leap seconds in TimestampParser. Can you have a look as well as #843?


dmikurube:
Thanks! Merging.

PR #843, opened 2017-11-10
Process yday (day of the year) in TimestampParser as well: Fix #834

### LINKED ISSUE #834: Use non-mDay values as well to build Timestamp

dmikurube:
@muga Fixing #834 ... supporting yday (`%j`) in `TimestampParser`. Can you have a look?


dmikurube:
Thanks!

PR #844, opened 2017-11-10
Add path_match_pattern option.

hiroyuki-sato:
Add `path_match_pattern` to LocalFileInputPlugin.

If you set `path_match_pattern: \.csv$`, Embulk reads following files.

```
.
`-- path
    `-- to
        `-- files
            |-- sample_01.csv    -> read
            |-- sample_02.csv.gz -> skip
            |-- sample_03.csv    -> read
            |-- sample_04.csv    -> read
```

This option available in 
[embulk-input-s3](https://github.com/embulk/embulk-input-s3#s3-file-input-plugin-for-embulk).

Please take a look when you have a time. 

Some users want to use this option. 

Ref: [Twitter](https://twitter.com/syu_cream/status/928278026986520576) (Japanese text)


y-ken:
This is perfect what I need.

dmikurube:
@hiroyuki-sato I personally think that matching with globs, not with regexps, can be better for local files... https://en.wikipedia.org/wiki/Glob_(programming)

But, options compatible with embulk-input-s3 are reasonable as well... Hmm, let us consider this. Can you leave the final design on us?

hiroyuki-sato:
Thank you for replying. 
I don't have any concern about globs or regexps.
I think options compatible with embulk-input-s3 is important.

kakoni:
Hi @dmikurube!
Getting back to this old topic. Should recent stable (0.9.24) embulk versions support file globbing?
I'm trying to do something simple as; ```path_prefix: abc[0-9]``` but can't get it to work.

dmikurube:
Unfortunately, it's not implemented there yet.

PR #847, opened 2017-11-13
Name and comment StrptimeParser's %Q as milliseconds, not microseconds

dmikurube:
@muga Not affecting any behavior, though.

`%Q` looks actually "milliseconds" in Ruby's documents, and even in our implementation behavior. But, the constant name and the comments are written it's "microseconds"...
https://docs.ruby-lang.org/en/2.4.0/DateTime.html#method-i-strftime

It might come from JRuby's long-lived code:
https://github.com/jruby/jruby/blame/a41e22b8f44114fc2783798cd22c76c47876bc51/core/src/main/java/org/jruby/util/RubyDateFormatter.java#L144-L146


dmikurube:
Thanks.

PR #848, opened 2017-11-14
Fix nanoseconds set by StrptimeParser's %Q

dmikurube:
@muga Can you have a look? I guess it's the last fix on `TimestampParser`.


dmikurube:
Thanks!

PR #773, opened 2017-08-29
Remove any global JRuby instance from the bootstrap by removing Ruby Embulk::Runner when Embulk is triggered from Java.

dmikurube:
This PR would be merged several versions after #765 is released.

PR #851, opened 2017-11-21
Release v0.8.38

dmikurube:
Merged #773. It would be released as a new version because I'm afraid it "can" have large impacts although it "shouldn't" have impacts.

@muga Can you have a look?


dmikurube:
Thanks!

PR #854, opened 2017-11-27
Move new plugin templates into Java resources not as a part of Ruby code

dmikurube:

This change is:
1. To locate the files more natural location.
    * `embulk new` and `embulk migrate` are now processed with Java code, not Ruby code. It is natural to locate as a Java resource, not as a part of Ruby code (in `/lib`).
2. To minimize the size of Ruby code in the Embulk package.

----

The files go into embulk-core, not embulk-cli, as a workaround against Shadow Gradle Plugin's bug.
https://github.com/johnrengelman/shadow/issues/111

If gradle-wrapper.jar is processed directly by Shadow Plugin in embulk-cli, gradle-wrapper.jar is unzipped unexpectedly. If the jar is once contained in embulk-core's jar, gradle-wrapper.jar is not unzipped.


dmikurube:
Thanks!

PR #855, opened 2017-11-28
Move new bundle templates into Java resources not as a part of Ruby code

dmikurube:
Similar to #854. This is moving `bundle` templates used for `embulk mkbundle`.

Removing `package_data.rb` as well which has been used only for this.


dmikurube:
Thanks! Merging.

Issue #628, opened 2017-05-15
Remove embulk.gem: `gem install embulk` and `require 'embulk'`


dmikurube:
Along with [un-JRuby-ing the boot process](https://github.com/embulk/embulk/issues?utf8=%E2%9C%93&q=milestone%3Aun-jruby-bootstrap%20), we plan to terminate supporting `embulk.gem`.

If we go with this, `gem install embulk` and `require 'embulk'` from CRuby/JRuby will be unavailable.

For testing purpose, we'll have another gem like `embulk-testing.gem` (TBD) instead.

If you have critical use-cases of `embulk.gem`, please share!

dmikurube:
Gems are no longer released since v0.9.0.

sonots:
Let me write a note about embulk.gem here.

Note that `embulk.gem` is no longer released on rubygems.org for embulk >= 0.9.0, but it is still available inside of embulk.jar.

# embulk -b /path/to/Gemfile

To use embulk -b option, you must write

```
gem 'embulk'
```

on your Gemfile, and you must create Gemfile.lock with your embulk as:

```
$ embulk bundle
```

You should see something like below on your Gemfile.lock:

```
    embulk (0.9.15-java)
      bundler (>= 1.10.6)
      liquid (~> 4.0.0)
      msgpack (~> 1.1.0)
```

which describes the version of your `embulk` gem which is bundled inside your embulk (so, same version with your embulk). Then, you can run your embulk with Gemfile.lock as:

```
$ embulk run -b /path/to/Gemfile
```

# Run tests of ruby plugin with embulk.jar

Although `embulk.gem` is not released on rubygems.org, it is still available inside embulk.jar.

Download embulk with .jar extension (this is mandatory to require from jruby):

```
$ curl -o embulk.jar --create-dirs -L "http://dl.embulk.org/embulk-0.9.15.jar"
```

Find the jruby version bundled in your embulk.jar, for example, as:

```
$ chmod a+x embulk.jar
$ ./embulk.jar irb
irb> JRUBY_VERSION
"9.1.5.0"
```

Install the same versioned jruby using rbenv or something (Currently, we do not have any options to use the bundled jruby of embulk, so we have to download jruby individually).

```
$ rbenv install jruby-9.1.5.0
$ gem install bundler
```

It is mandatory to require `embulk.jar` before bundler loads listed gems on your Gemfile. To achieve it, you can use `RUBYOPT` env or `ruby -r` option.

[...]

PR #856, opened 2017-11-28
Stop building embulk-*.gem distributions, and embed gem structure in embulk-core-*.jar

dmikurube:

Sorry that the PR is big... It's not in a hurry. This PR does multiple things at the same time due to their mutual dependencies.

1. Stops building `embulk-*.gem` packages as mentioned in #628.
2. Embeds a gem-like structure in `embulk-*.jar` similar to #819, instead.
3. Moves `lib/` and `test/` Ruby files into `embulk-core/src/{main,test}/ruby`.
4. Cleans-up unnecessary Ruby-related files and statements.


dmikurube:
Thanks! Merging.

PR #857, opened 2017-11-29
Make Ruby Embulk::VERSION based on Java org.embulk.EmbulkVersion.VERSION

dmikurube:
We'd like to have a version declaration just in one place! This PR automatically generates Ruby version string (`lib/embulk/verison.rb`) from Java version strting (`org.embulk.EmbulkVersion.VERSION`).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,6 +124,16 @@ subprojects {
             }
         }
 
+        jar {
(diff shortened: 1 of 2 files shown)


dmikurube:
Thanks!

PR #858, opened 2017-11-29
[WIP] Update embulk gem's version in Gemfile of bundle directory template to 0.9.0

dmikurube:


dmikurube:
Writing no any version here may work well with the gem-embedded embulk.jar? It'd be better if it works. Would check...

dmikurube:
Just `gem 'embulk'` would work, and it would be better than `gem 'embulk', '~> 0.9.0'`. Closing this PR.

PR #863, opened 2017-12-04
Parse time zones in Ruby-style in lower priority than Joda-Time

dmikurube:

It was a mistake in #840. I thought there were no duplications between Ruby's and Joda-Time's, but duplicated actually.
https://gist.github.com/dmikurube/c62ea8ebfbb5c41c0dc21584350bb274

If the given time zone is one of "CET", "EET", "Egypt", "Iran", "MET", or "WET", Joda-Time takes care of summer time, but Ruby's does not. It means that Joda-Time considers "CET" to be in summer time (+2) if the date is in a summer time period. Ruby's just always considers "CET" as a fixed offset of an hour (+1).

I'll backport this into v0.8 series, and release v0.8.39.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormat.java
+++ b/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormat.java
@@ -57,20 +57,39 @@ public class TimestampFormat
 
     public static DateTimeZone parseDateTimeZone(String s)
     {
-        final int rubyStyleTimeOffsetInSecond = TimeZoneConverter.dateZoneToDiff(s);
-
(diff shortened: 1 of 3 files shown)


dmikurube:
@sakama @muga Previous ea66a3ae37e022fa36b9bbb984b510b09374dfce was broken.

Please see 7c3a4f08316777b83e57fe917bbf054da24c231e which is a consistent (but likely tentative) fix.

dmikurube:
Thanks! Since this is deep inside the timestamp parser, waiting for @muga's response before merging.

dmikurube:
Thanks!

PR #866, opened 2017-12-05
Upgrade embedded Bundler to 1.16.0

dmikurube:
@muga @sakama Upgrading the embedded bundler to the latest.

`Bundler.load.setup_environment` was removed in 1.12.0, so calling `Bundler::SharedHelpers.set_bundle_environment` instead.


dmikurube:
Thanks!

PR #867, opened 2017-12-07
Stop using the global JRuby runtime when executed through CLI

dmikurube:
The JRuby instance for plugins is now operated only in `JRubyScriptingModule`. We can use non-global JRuby instance.


dmikurube:
Hmm, failing... Will check.

dmikurube:
Fixed. `rubyTest` runs from JRuby CLI ( == global JRuby instance), so it needs Embulk initialized for the global JRuby instance.

dmikurube:
Thanks!

PR #868, opened 2017-12-07
Use java.time classes instead of Joda-Time in embulk-cli

dmikurube:
@muga @sakama Small change. Let's use `java.time` more, and Joda-Time less. Can you have a look?


dmikurube:
Thanks! Merging.

PR #869, opened 2017-12-07
Call Gem.paths= instead of setting environment variables GEM_HOME and GEM_PATH with cleanups

dmikurube:
We don't want to touch environment variables as far as possible -- they are global and too impactful. This PR cleans up them.

@muga @sakama Can you PTAL?


dmikurube:
Thanks!

PR #870, opened 2017-12-08
Refactor setting Gem and Bundler variables in JRubyScriptingModule

dmikurube:
Refactoring in addition to #869. (Sorry for a kind of duplicated work.)

It refactors Gem and Bundler variable settings with more log messages added. It is also preparation for #871, which starts to use embulk.jar's location to find plugin gems.

@muga @sakama Can you PTAL?


dmikurube:
Thanks! Merging.

PR #871, opened 2017-12-08
[WIP] Use embulk.jar's location to find plugin gems

dmikurube:
(`embulk gem` command should take care of embulk.jar's location as well.)


dmikurube:
A similar thing implemented in a different approach as a part of v0.10.20: https://github.com/embulk/embulk/milestone/60

PR #872, opened 2017-12-11
Stop adding the path to the jar itself in JRuby $LOAD_PATH

dmikurube:
@sakama @muga Can you PTAL?

That `$LOAD_PATH` in the non-Bundler case is not required now.


muga:
LGTM!

dmikurube:
Thanks! Merging.

PR #873, opened 2017-12-11
Move the default install directory of gem plugins to ~/.embulk/lib/gem without Ruby versions

dmikurube:
I'm thinking to move the install directory of gem plugins to `~/.embulk/lib/gem` from `~/.embulk/jruby/<ruby-version>` (e.g. `~/.embulk/jruby/2.3.0` for JRuby 9.1.13.0). AFAIU, the directory with Ruby version has made more confusions than benefits on users when we upgrade JRuby. When JRuby of Embulk is upgraded, users need to re-install their plugins.

The v0.9 jump is a good chance to remove this for the future.

@muga @sakama Can I have your opinions? I might have overlooked kinds of problems or dependencies on the Ruby version.


dmikurube:
@muga @sakama Hearing no problems/claims on this.

Users need to re-install their plugins after this, but the same would happen when we upgrade JRuby to 9.2 series anyways. Once this is merged, such re-installation would be just once, and never happen.

dmikurube:
Thanks!

PR #874, opened 2017-12-12
Indirect JRuby calls not to require JRuby classes directly

dmikurube:
It comes after #872.

As in its Javadoc, this trick enables Embulk:
* To load a JRuby runtime (jruby-complete-...jar) specified at runtime out of embulk-...jar.
* To switch versions of JRuby without releasing or repackaging.
* To distribute Embulk without JRuby runtimes embedded.

Releasing without JRuby runtimes would not be done very soon. But, specifying other JRuby runtimes would sometimes help users.

@muga @sakama Can you PTAL?


dmikurube:
@muga I'll be merging this and later PRs next Monday (JST), jfyi.

dmikurube:
@muga Merging this. Please take a look later.

muga:
LGTM 👍 Sorry to be late.

PR #875, opened 2017-12-12
Simplify Ruby tests with Date._strptime monkey-patched

dmikurube:
@muga @sakama After #835, updating `rubyTest`. Can you have a look?


dmikurube:
Thanks! Merging.

PR #878, opened 2017-12-13
Move standard guess plugins to embulk-standards with cleaning up vanilla Ruby tests

dmikurube:
@muga @sakama As a part of cleanups, moving guess plugins for embulk-standards plugins (e.g. CSV) into embulk-standards. Can you PTAL?


dmikurube:
Thanks, both. Merging!

PR #879, opened 2017-12-14
Cleanup unnecessary `embulk/runner` in Ruby

dmikurube:
@sakama @muga The Ruby code `embulk/runner` is no longer needed.


dmikurube:
Thanks!

PR #880, opened 2017-12-14
Make JRuby-related versions via ScriptingContainerDelegate

dmikurube:
Making JRuby version ("9.1.13.0" at this time) and Ruby compatibility version ("2.3.3" then) available through `ScriptingContainerDelegate`. It **will** (not yet) be used for making `.ruby-version` in plugin code generated by `embulk` `bundle`, `migrate`, and `new`.


dmikurube:
Thanks!

PR #881, opened 2017-12-14
Cleanup JRuby instance initialization in JRubyScriptingModule

dmikurube:
@sakama @muga `JRubyScriptingModule` has had bunch of unnecessary initialization code, and ineffective comments. This PR cleans up them.

Can you PTAL?

PR #883, opened 2017-12-15
Merge embulk-cli into embulk-core with cleaning up build.gradle

dmikurube:
@muga @sakama Finally, merging `embulk-cli` into `embulk-core`, and cleaning up `build.gradle`.

It may be complex to read the changes in `build.gradle`. It may be easier to read `build.gradle` after this PR is applied.
* https://github.com/embulk/embulk/blob/809deadfbbbf68f602d7178b37c29f295467d03c/build.gradle


embulk-test/build.gradle
```diff
@@ -1,3 +1,14 @@
+// TODO: Have more details in description and summary.
+description = 'Embulk test helper'
+ext {
+    summary = 'EMbulk test helper'
```
sakama: Minor: 'Embulk'?

dmikurube: Ah, thanks for the good catch! Fixed.

dmikurube:
Thanks! #881 has been merged, then merging this.

PR #884, opened 2017-12-18
Eliminate "import org.jruby..." from CLI and embulk-standards

dmikurube:
After merging embulk-cli to embulk-core (#883), stopping to use JRuby directly from CLI, and embulk-standards.

@muga @sakama Can you PTAL?


dmikurube:
Thanks!

PR #885, opened 2017-12-18
Deprecate "updateJRuby" Gradle task in build.gradle

dmikurube:
The JRuby version can be retrieved at runtime. We don't want to edit the source code from Gradle tasks.


hiroyuki-sato:
@dmikurube 

Just FYI
I  found `updateJRuby` description in README.md
https://github.com/embulk/embulk/blob/master/README.md#update-jruby

dmikurube:
@hiroyuki-sato Thanks for the good catch. Will cleanup README later in bulk.

dmikurube:
Merging...

PR #887, opened 2017-12-19
Use java.time.Instant in org.embulk.spi.time.Timestamp

dmikurube:
Just refactoring. Using `java.time.Instant` in `Timestamp`, instead of pairs of epoch seconds and milliseconds. Their interface is almost the same.

At the same time, moving `Timestamp.fromString` into `TimestampSerDe`, which was package-private, and used only from `TimestampSerDe`.

@sakama @muga Can you take a look?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/time/Timestamp.java
+++ b/embulk-core/src/main/java/org/embulk/spi/time/Timestamp.java
@@ -1,107 +1,97 @@
 package org.embulk.spi.time;
 
-import java.util.regex.Pattern;
-import java.util.regex.Matcher;
-import org.joda.time.format.DateTimeFormat;
-import org.joda.time.format.DateTimeFormatter;
-import org.joda.time.DateTime;
+import java.time.Instant;
+import java.time.ZoneOffset;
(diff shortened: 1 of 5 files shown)


dmikurube:
Ah, failing. Having a look.

dmikurube:
Found the cause, and updated the test.

It was because handling negative values in existing `Timestamp` was different from `java.time.Instant`. In existing `Timestamp`, `-0.9s` could be represented as `-0s -900ms`. In `java.time.Instant`, `-0.9s` is always represented as `-1s + 100ms`.
https://github.com/embulk/embulk/pull/887/files#diff-01070f6c2fe77f8ccd076a97fe29d90aR223

I personally think the behavior of `java.time.Instant` is acceptable and better since:
1. `-1s + 100ms` looks more consistent in terms of `0 <= nanoAdjustment < 1_000_000_000`.
2. The representation of `-1s + 100ms` was also possible in existing `Timestamp`.
3. A unique value should have a unique representation.

@sakama @muga What do you think?

dmikurube:
Thanks!

PR #889, opened 2017-12-20
Indirect org.joda.time.DateTimeZone in config Tasks

dmikurube:
Users of TimestampFormatter, TimestampParser, and DynamicPageBuidler had to use `org.joda.time.DateTimeZone` since their `Task` subclasses had `DateTimeZone`.

This change enables to intercept its `String` representation, and to interpret the time zone ID representation by themselves.

It introduces a hack in `TaskInvocationHandler` to call default implementations in `Task` interfaces.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormatter.java
+++ b/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormatter.java
@@ -1,8 +1,6 @@
 package org.embulk.spi.time;
 
 import java.util.Locale;
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
 import com.google.common.base.Optional;
 import org.jruby.util.RubyDateFormat;
 import org.embulk.config.Config;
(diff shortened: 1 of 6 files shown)


embulk-core/src/main/java/org/embulk/spi/time/TimestampFormatter.java
```diff
@@ -1,8 +1,6 @@
 package org.embulk.spi.time;
 
 import java.util.Locale;
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
```
dmikurube: NOTE: We'll be introducing `java.time...` here soon. Removing `import org.joda...` and using `org.joda...` directly in the main code so that we won't be confused by `DateTimeZone` (Joda-Time) and `ZoneId` (java.time) in the same place.

muga:
@dmikurube Can you tell me the purpose of this PR more? We'll introduce java.time and the `ZoneId`. The goal is to replace joda-time's `DateTimeZone` from SPI. So, I guess that this PR is one of intermediate steps for the goal but, I'm confusing about how to migrate the code for the goal.

dmikurube:
The Config/Task classes will have just `String` getters eventually. The `DateTimeZone` getters are kept for a while (~v0.10?) because some existing plugins are depending. I believe we won't need `ZoneId` getters in Config/Task. Users of Config/Task can parse the time zone ID strings by themselves.

I'm still wondering about constructors and creator methods. They'd have `String` parameters for the time being, but constructors/creators with `ZoneId` parameters may be added later.

muga:
Got it. Thank you for explaining. Sounds good.

dmikurube:
Thanks! Merging.

PR #891, opened 2017-12-20
Implement a time zone parser for java.time.ZoneId

dmikurube:
Moving `TimestampFormat.parseDateTimeZone` to a new class `TimeZoneIds`, and adding a new method to parse a time zone ID string into `java.time.ZoneId`.

@sakama @muga Can you PTAL?


dmikurube:
Thanks!

PR #892, opened 2017-12-20
Reimplement TimestampParser, TimeParsed, and RubyTimeParsed with java.time.ZoneId

dmikurube:
This comes after #893 and #894. It deprecates and reduces the usage of `org.joda.time.DateTimeZone`.


dmikurube:
Thanks!

PR #894, opened 2017-12-21
Refactor fields of RubyTimeParsed

dmikurube:
Some more refactoring in `RubyTimeParsed` so that it can use `RubyTimeParsed`'s fields directly in `toTimestamp` and `asMapLikeRubyHash` without converting getters.

@muga @sakama Can you PTAL?


dmikurube:
Thanks!

PR #896, opened 2017-12-24
Remove a not used argument

yui-knk:


dmikurube:
@yui-knk Thanks for your contribution, but we're still considering whether we process JRuby options in subcommands as commented at: https://github.com/embulk/embulk/blob/3e2b4a3ac0fe2c2d2688aac984655aaa114fa51d/embulk-core/src/main/java/org/embulk/cli/EmbulkRun.java#L749

We'd remove this once we decide we finally don't need, but let us keep this for the time being.

yui-knk:
I see.
Thanks for your quick response :)

PR #898, opened 2017-12-26
Cleanup TimestampParser orders and styles with reimplementing defaultDate with java.time

dmikurube:
Simply cleaning up. Can you have a look, @muga @sakama?


dmikurube:
Thanks!

@muga Merging this. Let me know if you see any concerns.

PR #901, opened 2017-12-26
Add Java-style ("java:") timestamp format to parse timestamps

dmikurube:
It additionally accepts new-style timestamp formats prefixed by `"java:"` and composed from `java.time.DateTimeFormatter`'s format style, such as `"java:EEE, dd MMM uuuu HH:mm:ss ZZZ"`. Some cases could not be supported in the Java-version parser, though.

The spec is at:
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#patterns
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatterBuilder.html#appendPattern-java.lang.String-

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/time/TimeZoneIds.java
+++ b/embulk-core/src/main/java/org/embulk/spi/time/TimeZoneIds.java
@@ -25,6 +25,13 @@ public class TimeZoneIds
     }
 
     /**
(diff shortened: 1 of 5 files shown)


muga:
LGTM

dmikurube:
Thanks!

PR #902, opened 2017-12-28
Deprecate org.embulk.spi.time.TimestampFormat

dmikurube:
`org.embulk.spi.time.TimestampFormat` is essentially just a `String`. We don't need it eventually.


dmikurube:
Thanks!

Issue #860, opened 2017-11-29
Parse time zones in a consistent manner


dmikurube:
Embulk has parsed timestamps based on Ruby's `Date._strptime` since v0.2.0 (Jan, 2015). Although that is reimplemented in Java in #611 since v0.8.27, it keeps compatibility with Ruby's.

Parsing timestamps (v0.2.0):
* https://github.com/embulk/embulk/blob/v0.2.0/embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java#L44
* https://github.com/embulk/embulk/blob/v0.2.0/lib/embulk/java/time_helper.rb#L27

At the same time, their time zones have been parsed wiht Joda-Time's `DateTimeFormat` and `DateTimeZone`.

Parsing time zones (v0.2.0):
* https://github.com/embulk/embulk/blob/v0.2.0/embulk-core/src/main/java/org/embulk/spi/time/TimestampParser.java#L50
* https://github.com/embulk/embulk/blob/v0.2.0/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormat.java#L41-L68

There were some problems in the usage of Joda-Time. The inconsisntency between parsing timestamps and time zones caused problems.

### Ruby-only time zones are not recognized

Ruby's `Date._strptime` recognizes their own time zone formats, such as `"China Standard Time"`, `"JST Daylight Time"`, and `"PST DST"`. They are extracted by `Date._strptime` by its regular expression, but they are not recognized by Joda-Time. Finally, "`2017-01-02 03:04:05 PST DST`" has been ingested as the same as "`2017-01-02 03:04:05 UTC`" until v0.8.36.

### Some daylight time zone IDs are recognized as non-daylight time zones

Embulk has called Joda-Time's `DateTimeFormat.forPattern("z").parseMillis(...)` at first to recognize time zones. It has had problems.

* https://github.com/embulk/embulk/blob/v0.8.38/embulk-core/src/main/java/org/embulk/spi/time/TimestampFormat.java#L74

Joda-Time just converts both `"PST"` and `"PDT"` into `America/Los_Angeles`.

* https://github.com/JodaOrg/joda-time/blob/v2.9.2/src/main/java/org/joda/time/DateTimeUtils.java#L446

[...]

PR #905, opened 2018-01-10
Add Ruby-style ("ruby:") timestamp format to parse timestamps in the compatible way with Time.strptime

dmikurube:
The legacy `TimestampParser` has parsed into components by `DateTime._strptime`, partially followed by `DateTime.strptime`'s manner to resolve it as a `DateTime` object, partially with Joda-Time's `DateTimeZone`, and so on... It was confusing. (See #860.)

This, the new Ruby-style timestamp parser, follows `Time.strptime` of `Time`, which is Ruby's more accepted standard-ish time class. `Time.strptime` has simple and reasonable time zone handling only with numeric time zones, and a few fixed-offset short IDs used in some RFCs.

We can refer Ruby's `Time.strptime` documentation straightforward to describe the parser's behavior.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/time/RubyTimeParsed.java
+++ b/embulk-core/src/main/java/org/embulk/spi/time/RubyTimeParsed.java
@@ -1,8 +1,9 @@
 package org.embulk.spi.time;
 
 import java.math.BigDecimal;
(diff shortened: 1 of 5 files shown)


dmikurube:
Thanks! Merging.

PR #906, opened 2018-01-12
Add Ruby-style ("ruby:") timestamp format to format timestamps in the compatible way with Time.strptime

dmikurube:
Accepting "ruby:" timestamp formats in `TimestampFormatter` in addition to `TimestampParser` at #905, simply for consistency.

There are not so many differences in formatter, not like parser.

@muga @sakama Can you have a look? More test cases would be added sooner in other PRs.


dmikurube:
Thanks!

PR #907, opened 2018-01-14
Add Java-style ("java:") timestamp format to format timestamps with java.time

dmikurube:
After #906, adding Java-style "`java:`" timestamp formatter with `java.time`.

@muga @sakama Can you have a look? More test cases would be added sooner in other PRs.


dmikurube:
Thanks!

PR #908, opened 2018-01-15
Remove unnecessary reference of a Java property "java.io.tmpdir"

dmikurube:
`org.embulk.spi.unit.LocalFileSerDe$LocalFileDeserializer` refers a Java property `java.io.tmpdir`, but that is never used. Let's remove that.

@sakama Can you have a look? (Cc: @muga)


dmikurube:
Thanks! Merging.

PR #909, opened 2018-01-15
Add Embulk design docs in Markdown under docs/design

dmikurube:
Adding internal design documents in the repository. It is nicer to be in the repos so that their revisions are synchronized with the code changes. They'll be separate from `embulk-docs` since they're more Embulk-internal documents, not for embulk.org/docs.

@sakama @muga What do you think?


dmikurube:
Thanks! Merging this.

@muga Please ping if you see any concerns.

PR #912, opened 2018-01-16
Configure M2_REPO from Embulk's system config with moving the default to ~/.embulk/lib/m2/reposotory

dmikurube:
Setting by environment variables is often confusing us. We basically should configure by Embulk's own system config prioritized over OS environment variables.

We may want to stop reading the env `M2_REPO`, not only prioritizing the system config.


dmikurube:
Rebased.

dmikurube:
Thanks. Merging!

PR #913, opened 2018-01-16
Remove the "setVersion" task in build.gradle

dmikurube:
The `setVersion` task is not needed. Let's remove this. @muga @sakama


dmikurube:
Thanks. Merging this as well!

PR #914, opened 2018-01-17
Enable checkstyle for embulk-core's main Java files

dmikurube:
After #582, this enables checkstyle for embulk-core's main Java files with styling the existing code.

@muga @sakama Can you have a look?

PR #915, opened 2018-01-18
Enable checkstyle for embulk-core's test Java files

dmikurube:
After #914, enabling checkstyle for embulk-core tests, then.


dmikurube:
Thanks!

PR #916, opened 2018-01-18
Enable checkstyle for embulk-standards' Java files

dmikurube:
After #915, enabling checkstyle for embulk-standards.


dmikurube:
Merging!

PR #917, opened 2018-01-18
Enable checkstyle for embulk-test' Java files

dmikurube:
At last after #916, enabling checkstyle for embulk-test.


dmikurube:
Merging this last one. Thanks!

PR #920, opened 2018-01-23
Upgrade shadow Gradle plugin to 2.0.2

dmikurube:
Once JRuby was upgraded to 9.1.15.0, it turned out that the Gradle `shadowJar` task failed. Found that upgrading the `shadow` Gradle plugin fixes the issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.3'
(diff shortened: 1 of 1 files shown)


muga:
please go ahead. to be honest i'm not sure that it could be fixed or not though..

dmikurube:
Thanks! I confirmed that this update (shadow 2.0.2) fixed the build problem, and the built binary worked at least for a simple Embulk `run` task.

dmikurube:
Thanks!

PR #930, opened 2018-01-24
Deprecate new things to be deprecated, and suppress uses of deprecated things

dmikurube:
@sakama @muga We're going to deprecate old unused things explicitly by the chance of v0.9. Do you have any other candidates to deprecate?

This PR enables the `-Xlint:deprecation` check, and annotates `@SuppressWarnings("deprecation")` explicitly, and comments with GitHub Issues, as well.


dmikurube:
Thanks! Looks that no other deprecations suggested. Merging this.

PR #931, opened 2018-01-25
Use TimestampFormatter.of and TimestampParser.of instead of their constructors

dmikurube:
The use of `TimestampFormatter.of` and `TimestampParser.of` was not thorough. This PR replaces almost all the constructor calls to `.of`.


dmikurube:
Thanks!

PR #938, opened 2018-01-25
Move resources new_bundle_tempaltes/ to org/embulk/jruby/bundler/template/, and new_plugin_templates/ to org/embulk/plugin/template/

dmikurube:
Just a cleaning-up change. Moving:

* plugin template resources (`embulk new` and `embulk migrate`) to `org/embulk/plugin/template`
* Bundle directory template resources (`embulk mkbundle`) to `org/embulk/jruby/bundler/template`


dmikurube:
Thanks!

PR #939, opened 2018-01-25
Log Embulk core's version when executing EmbulkEmbed: Fix #882

### LINKED ISSUE #882: Print Embulk core's version when executed from EmbulkEmbed

dmikurube:
This tries to address #882 -- logging Embulk's version when executing some methods of `EmbulkEmbed`. Can you have a look? @muga @sakama

The execution log would be like below:

```
2018-01-25 16:07:53.768 +0900 [INFO] (main): BUNDLE_GEMFILE is being set: "/.../Gemfile"
2018-01-25 16:07:53.771 +0900 [INFO] (main): Gem's home and path are being cleared.
2018-01-25 16:07:56.048 +0900 [INFO] (main): Started Embulk v0.9.0-SNAPSHOT
2018-01-25 16:07:56.139 +0900 [INFO] (0001:transaction): Loaded plugin embulk-input-inline (0.1.0)
```

### DIFF
--- a/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
+++ b/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
@@ -104,13 +104,14 @@ public class EmbulkEmbed {
     }
 
     private final LifeCycleInjector injector;
+    private final org.slf4j.Logger logger;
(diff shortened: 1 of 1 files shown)


dmikurube:
Thanks! Merging.

PR #940, opened 2018-01-26
Add an internal design document on deprecations

dmikurube:
Adding another internal document to summarize `@Deprecated` in the Embulk core.

Can you have a look? @muga @sakama (Not in a hurry.)


dmikurube:
Thanks!

Issue #788, opened 2017-09-07
`guess` subcommand read 32KB only as sample data even  I define guess_sample_buffer_bytes explicitly.


hiroyuki-sato:
Related: https://github.com/hiroyuki-sato/embulk-parser-jsonpath/issues/15

## Expect behavior

`guess` sub command read `guess_sample_buffer_bytes` bytes as sample data.

## Actual behavior

`guess` sub command read 32KB only as sample data even log output `Try to read 3,072,000 bytes`.

## Environment

* Embulk 0.8.31
* [embulk-parser-jsonpath](https://github.com/hiroyuki-sato/embulk-parser-jsonpath): 0.2.0

## Details

guess code is below.

```ruby
    class Jsonpath < TextGuessPlugin
      Plugin.register_guess("jsonpath", self)

      def guess_text(config, sample_text)      def guess_text(config, sample_text)
puts "sample_text size: #{sample_text.size}"
```

seed.yml

```yaml
in:
  type: file
  path_prefix: test.json
  parser:
    type: jsonpath
    root: "$.results"
exec:
  guess_sample_buffer_bytes: 3072000
out:
  type: stdout
```

```
ls -lh test.json
-rw-r--r--  1 user  wheel   2.4M Sep  7 08:54 test.json
```

```
embulk guess -I /path/to/embulk-parser-jsonpath/lib -g jsonpath seed.yml
2017-09-07 09:01:07.169 +0900: Embulk v0.8.31
...
2017-09-07 09:01:10.093 +0900 [INFO] (0001:guess): Try to read 3,072,000 bytes from input source
...
sample_text size: 32768
org.jruby.exceptions.RaiseException: (null) unexpected token at '{
      "name": "Jarod Muller",
      "city": "Gradyfurt",
      "street_name": "Lind L'
	at RUBY.load(/Users/hsato/.embulk/jruby/2.3.0/gems/multi_json-1.12.1/lib/multi_json.rb:124)
	at RUBY.process_object(/path/to/home/.embulk/jruby/2.3.0/gems/jsonpath-0.5.8/lib/jsonpath.rb:87)
	at RUBY.enum_on(/path/to/home/.embulk/jruby/2.3.0/gems/jsonpath-0.5.8/lib/jsonpath.rb:73)
	at RUBY.on(/path/to/home/.embulk/jruby/2.3.0/gems/jsonpath-0.5.8/lib/jsonpath.rb:65)
[...]

PR #942, opened 2018-01-27
Customize byte size of sample buffer that GuessParserPlugin tries to read

muga:
This PR enables customizing byte size of sample buffer that `GuessParserPlugin` can tries to read. In #609, `guess_sample_buffer_bytes` option was introduced to custom byte size of sample buffer but, in the case of #788, the option could not work expectedly. #609 enables configuring sample buffer bytes for `GuessExecutor` but, `GuessParserPlugin` still tries to read default bytes 32KB that is not configurable by #609. This PR could configure for the sample buffer bytes. 

In this PR, to pass the custom value of `guess_sample_buffer_bytes` option in `GuessExecutor` to `GuessParserPlugin`, `parser_sample_buffer_bytes` option is introduced. Basically `parser_sample_buffer_bytes` value should be same as `guess_sample_buffer_bytes`. 

It fixes https://github.com/embulk/embulk/issues/788.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/GuessExecutor.java
+++ b/embulk-core/src/main/java/org/embulk/exec/GuessExecutor.java
@@ -122,11 +122,12 @@ public class GuessExecutor {
         GuessExecutorTask task = execConfig.loadConfig(GuessExecutorTask.class);
         guessPlugins.addAll(task.getGuessPlugins());
         guessPlugins.removeAll(task.getExcludeGuessPlugins());
+        final int parserSampleBufferBytes = task.getSampleBufferBytes();
 
-        return guessParserConfig(sample, inputConfig, guessPlugins);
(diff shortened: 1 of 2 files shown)


dmikurube:
@muga Please update `embulk-docs/src/release/release-0.9.0.rst` as well in this PR.

muga:
@dmikurube can you take a look when you have time? I update release-0.9.0.rst as well.

muga:
@dmikurube that's good point. Will change option name with `guess_parser_sample_buffer_bytes`. As a matter of fact, I've not seen that `GuessParserPlugin`, means `system_guess` is used directly by users or plugins but, the plugin could be called by users or plugins. So it's better to rename it.

muga:
@dmikurube thank you for reviewing. merging..

PR #947, opened 2018-01-31
Add a design doc of Timestamp parsing

dmikurube:
@muga @sakama Drafted a design doc of timestamp parsing in Embulk. Can you have a look?


dmikurube:
Thanks!

PR #951, opened 2018-02-05
Allow the same @Config configuration names: Fix #949

### LINKED ISSUE #949: Plugin's Task causes JsonMappingException when the Task extends TimestampFormatter.Task with @Config("default_timezone")
When a plugin has its `Task` interface which extends core's `Task` interface with `default` methods for compatibility, it causes `JsonMappingException` such as:

```
com.fasterxml.jackson.databind.JsonMappingException: Multiple entries with same key: default_timezone=org.embulk.config.TaskSerDe$TaskDeserializer$FieldEntry@2548fc01 and default_timezone=org.embulk.config.TaskSerDe$TaskDeserializer$FieldEntry@4331be30
```

embulk-filter-calcite is the case.
https://github.com/muga/embulk-filter-calcite/blob/v0.1.3/src/main/java/org/embulk/filter/calcite/CalciteFilterPlugin.java#L246-L248

The direct cause is that the extended `interface` has two getter methods with the same `@Config("default_timezone")`.

Options to fix it:

1. If the parent's method has `default` implementation, and it does not have `@Config`, the overridden method behaves like it does not have `@Config`. In calcite's case, `getDefaultTimeZone` behaves like it does not have `@Config`.
2. Multiple getters with the same `@Config` configuration names are allowed.

dmikurube:
This PR is a trial to fix #949, but it changes the behavior of `@Config`. It allows to have duplicated configuration names of `@Config("name")`.

Can I have your thoughts, @muga, @sakama?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/config/TaskInvocationHandler.java
+++ b/embulk-core/src/main/java/org/embulk/config/TaskInvocationHandler.java
@@ -1,6 +1,7 @@
 package org.embulk.config;
 
-import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.Multimap;
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationHandler;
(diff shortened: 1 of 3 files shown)


dmikurube:
Thanks! Added tests for this. Merging this when CI gets happy.

dmikurube:
Passed. Merging!

PR #956, opened 2018-02-08
Remove @Config indirection by default implemantations

dmikurube:
By #951, the indirection is no longer required. It's like reverting #889.


dmikurube:
Thanks!

PR #960, opened 2018-02-09
Fix StdoutOutputPlugin broken by stopping @Config indirection by default implementation at: #956

dmikurube:
The default implementation has been introduced at: #931, and it's to be a kind of reverted as well in addition to #956.


dmikurube:
An unused import remained...

dmikurube:
Passed. Merging... Thanks!

PR #962, opened 2018-02-10
Initialize JRuby's ScriptingContainer lazily

dmikurube:
@sakama @muga Trying to initialize JRuby lazily. Can you have a look?


dmikurube:
Hmm, some tests written in Ruby have got broken... Will have a look.

dmikurube:
Fixed. The problem was that it is not initialized when Ruby tests run as it is executed from Ruby without `ScriptingContainer`.

dmikurube:
Thanks! Merging.

PR #963, opened 2018-02-10
Load non-JRuby plugins without JRuby by deprioritizing JRuby-based plugins

dmikurube:
Finally! It makes Embulk's startup really fast when JRuby-based plugins are not used.

It stops using Guice for `JRubyPluginSource`. Much controversy may exist over using DI (Guice), but  I think we may want to use Guice less in Embulk...


dmikurube:
Thanks!

@muga The problem is the order. Even if `List<PluginSource>` can be specified, I'm not sure how we can specify the order on the Modules side.

dmikurube:
Merging this for v0.9.3. Thanks!

PR #967, opened 2018-02-10
Revert "Remove @Config indirection by default implemantations": Fix #964

dmikurube:
The removal in v0.9.1 did not work perfectly as reported in #964... Reverting the removal would fix the issue.

@sakama @muga Can you have a look?

PR #968, opened 2018-02-10
Prepare for v0.9.3

dmikurube:
@sakama @muga v0.9.3 will be #962 and #963 after #967 on v0.9.2. Can you have a look?


dmikurube:
Thanks!

PR #971, opened 2018-02-11
Use Time objects in Ruby: Fix #970

dmikurube:
@muga @sakama Can you have a look? It is trying to fix #970.

PR #972, opened 2018-02-14
Prepare for v0.9.4

dmikurube:
Not concluded yet, but we may release v0.9.4 with the fix of #971 if we agree. What do you think? @sakama @muga


dmikurube:
Thanks!

Issue #978, opened 2018-03-09
Occasionally failed by "JRuby runtime is not loaded successfully" error


kamikaseda:
# Summary
The Embulk process occationally fails due to following error when doing input-file and output-postgresql.
```
2018-03-08 03:02:53.999 +0900 [INFO] (main): Started Embulk v0.9.4
2018-03-08 03:02:54.477 +0900 [INFO] (0001:transaction): Gem's home and path are set by default: "/root/.embulk/lib/gems"
SystemCallError: Unknown error (SystemCallError) - channel: org.jruby.util.io.ChannelFD@1da5c19d uri:classloader:/META-INF/jruby.home/lib/ruby/gems/shared/specifications/default/minitest-5.4.1.gemspec
              gets at org/jruby/RubyIO.java:2301
          readline at org/jruby/RubyIO.java:2692
     block in data at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/stub_specification.rb:118
              open at org/jruby/RubyIO.java:1156
              open at org/jruby/RubyKernel.java:306
              data at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/stub_specification.rb:116
            valid? at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/stub_specification.rb:207
            select at org/jruby/RubyArray.java:2566
  gemspec_stubs_in at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/specification.rb:751
     default_stubs at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/specification.rb:758
             stubs at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/specification.rb:833
      find_by_path at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/specification.rb:1038
      try_activate at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems.rb:209
           require at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:126
            <main> at <script>:1
org.embulk.jruby.JRubyNotLoadedException: JRuby runtime is not loaded successfully.
        at org.embulk.jruby.ScriptingContainerDelegateImpl.runScriptlet(ScriptingContainerDelegateImpl.java:725)
[...]

PR #979, opened 2018-03-12
Throw more explicit Exceptions from JRuby ScriptingContainerDelegate

dmikurube:
It is not fixing #978, but making it a bit more clear.


dmikurube:
@sakama @muga Can you have a look? `ScriptingContainerDelegate`'s Exception handling has not been perfect. This PR tries not to hide JRuby's original error messages.

dmikurube:
Thanks. Merging!

PR #981, opened 2018-03-20
Accept "embulk bundle --version" to run "bundle --version"

dmikurube:
A small fix. Until v0.9.4, `embulk bundle --version` has been considered the same as `embulk --version`. We has had no way to show Bundler's versions. This PR fixes it.

Can you have a look?


dmikurube:
Thanks!

PR #982, opened 2018-03-20
Show test logs in standard streams

dmikurube:
To see test outputs not only in report html pages, but also in standard output/error in real time.


dmikurube:
Thanks!

PR #985, opened 2018-03-29
[WIP] Implement Fluentd Reporter plugin

sakama:
Implemented Fluentd Reporter plugin based on #984 

### Requirements

Jackson update in embulk-core is required.
https://github.com/embulk/embulk/blob/master/embulk-core/build.gradle#L43-L48

Fluency 1.7.0 is using Jackson 2.7.1 but embulk-core is using 2.6.7.
I couldn't run without updating Jackson.


sakama:
I'm going to implement as standalone plugin.
Close this issue.

PR #994, opened 2018-04-03
Print column names in "stdout" output plugin

dmikurube:
No any "important" feature, but I wanted this when checking a column-renaming feature...

Can you have a look? @sakama

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/StdoutOutputPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/StdoutOutputPlugin.java
@@ -8,6 +8,7 @@ import org.embulk.config.ConfigSource;
 import org.embulk.config.Task;
 import org.embulk.config.TaskReport;
 import org.embulk.config.TaskSource;
+import org.embulk.spi.Column;
 import org.embulk.spi.Exec;
 import org.embulk.spi.OutputPlugin;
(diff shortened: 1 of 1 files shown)


embulk-standards/src/main/java/org/embulk/standards/StdoutOutputPlugin.java
```diff
@@ -40,6 +45,15 @@ public ConfigDiff transaction(ConfigSource config,
...
             OutputPlugin.Control control) {
         final PluginTask task = config.loadConfig(PluginTask.class);
+        if (task.getPrintsColumnNames()) {
+            for (final Column column : schema.getColumns()) {
+                if (column.getIndex() > 0) {
+                    System.out.print(",");
+                }
+                System.out.print(column.getName());
+            }
+            System.out.println("");
```
sakama: Why does this empty output is needed?
Is it intended to ":" ?

dmikurube: Ah, no. It's newline. One option was to use `\n`, but `println` might be better.

sakama: Ah Ok. I overlooked print"ln".

dmikurube:
Thanks!

PR #996, opened 2018-04-04
Eliminate using Guava's Throwables#propagate

dmikurube:
See https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate

Can you have a look? @sakama (Cc: @muga)


dmikurube:
Hmm, it looks that this change broke the "cause" structure of Exception. Will chase.

dmikurube:
Fixed!

dmikurube:
Thanks!

PR #997, opened 2018-04-04
Use java.util.Optional more than com.google.common.base.Optional #1: Config and internal use

dmikurube:
Trying to start using `java.util.Optional` of Java 8, in addition to / instead of Guava's `Optional`.

Can you have a look, @sakama, and @muga as it's related to the basement portions?

In plugins' `Task` interfaces: both `java.util.Optional` and Guava's can be used like https://github.com/embulk/embulk/pull/997/commits/3e24e20dca5b57b441cd81bccb82f2df4eef8aab#diff-828bcc9d321473fc4a546d301ede851bR42 thanks to `Jdk8Module`.

In `TaskSerDe`: `TaskDeserializer` and `ConfigTaskDeserializer` should not be extended from plugins. The `protected` methods `getJsonKey` and `getDefaultJsonString` can be modified to return `java.util.Optional`.

In `ExecSession`: The only method using `Optional` is a `private` constructor, and only `ExecSession.Builder#build` is calling the constructor internally. It can be replaced to `java.util.Optional` safely.

In `ToString`: Guava's `Optional` has been used in its `public` constructor with `@JsonCreator`. The Guava-version should be kept for compatibility for direct callers of the constructor (although deprecated). At the same time, `@JsonCreator` method can switch to `java.util.Optional` safely since Jackson databind handles that.

### DIFF
--- a/embulk-core/build.gradle
+++ b/embulk-core/build.gradle
@@ -44,6 +44,7 @@ dependencies {
     compile 'com.fasterxml.jackson.core:jackson-core:2.6.7'
     compile 'com.fasterxml.jackson.core:jackson-databind:2.6.7'
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-guava:2.6.7'
+    compile 'com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.6.7'
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-joda:2.6.7'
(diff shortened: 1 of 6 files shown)


dmikurube:
Thanks!

@muga Yes and no. We'd have `#2` to replace more, but it is hard to eliminate Guava's `Optional` soon for compatibilities -- they're used in plugin SPI.

muga:
@dmikurube Got it. Thank you.

dmikurube:
Thanks. Merging!

Issue #1003, opened 2018-04-12
[doc] prints_column_names option in stdout output plugin.


hiroyuki-sato:
`prints_column_names` in stdout plugin option introduced in 0.9.6.
It is helpful that this option description is available.

hiroyuki-sato:
This feature undocumented(developer only) feature yet.

PR #1004, opened 2018-04-12
Add stdout output plugin document.

hiroyuki-sato:
Add stdout output plugin document. This PR for #1003.


hiroyuki-sato:
This feature undocumented(developer only) feature yet.

PR #1006, opened 2018-04-13
Get org.jruby.embed.ScriptingContainer injectable again

dmikurube:
`org.jruby.embed.ScriptingContainer` could not be injected since v0.9.0 as `ScriptingContainer` was indirected by `ScriptingContainerDelegate`. But, found that some plugins are injecting it, especially in constructing instances. We at least need to keep this injectable for a while.

https://github.com/embulk/embulk-executor-mapreduce/blob/v0.4.0/embulk-executor-mapreduce/src/main/java/org/embulk/executor/mapreduce/MapReduceExecutor.java#L71

It will be dropped after a while -- maybe in the next major version.

@sakama Can you have a look?


dmikurube:
Thanks! Merging...

PR #1012, opened 2018-04-17
Find a Maven-based plugin JAR with its "direct" dependencies

dmikurube:
@sakama It's totally NOT in a hurry! Can you have a look just when you have time?

This tries finding (only direct) dependencies of a Maven-based plugin JAR.

----

Embulk has expected that a Maven-based plugin is a single JAR (fat-JAR or JARs-in-JAR) so far. But, it is not very good, especially in performance, and in publishing the plugin.

The reason of "single JARs" was that resolving Maven transitive dependencies is not deterministic. When a dependency library `A` depends on another library `B`, `B`'s version is not deterministic. And what is even worse, the dependencies are resolved at Embulk runtime. It may cause problems like: "The plugin works at user P, but does not work at user Q.". It would really confuse ourselves.

The dependency problem is relaxed if all dependencies are guaranteed to have been resolved at compile time. To realize this guarantee, this PR's implementation finds dependencies, but only its "direct" dependencies.

A plugin developer needs to resolve all nested dependencies at compile time, and write all the dependencies directly in `pom.xml` of the plugin JAR. For ease of plugin developers, the Gradle plugin to build a Maven-based Embulk plugin needs to be updated as well.

~~I'd add a design doc for this later.~~ Added a design doc.


docs/design/maven_based_plugins.md
```diff
@@ -0,0 +1,16 @@
+Maven-based Plugins
+====================
+
+Plugins are loaded only through RubyGems except for embulk-standards and Guice modules until v0.8.25. Maven-based JAR plugins are supported since v0.8.26 so that users finally do not need JRuby.
+
+
+Dependency Libraries
+---------------------
+
+A Maven-based plugin was expected to be a single JAR including its dependency libraries (fat JAR or JARs-in-JAR) until v0.9.7. It had had problems especially in publishing Maven-based plugins such as sizes of plugins, and licenses.
```
muga: typo: had had

dmikurube: Ah, I intended past-perfect-tense there.

dmikurube:
Thanks!

PR #1015, opened 2018-04-19
Load a Maven-based plugin JAR with its "direct" dependencies

dmikurube:
After finding "direct" dependencies in #1012, this PR actually loads the out-of-plugin-JAR dependencies. Can you have a look when you have time? @sakama (Cc: @muga) It's not in a hurry.

We may want better tests to load JARs, but that'd be out of `embulk-core`. We'd need much work for such tests.

I've at least confirmed that it successfully loads and runs specially-built `embulk-decoder-commons_compress` with external `org.apache.commons:commons-compress:1.13`.


dmikurube:
Thanks!

Issue #1021, opened 2018-06-13
path_prefix does not match case-insensitve 


jca02266:
For example, following config.yml does not work on Windows. Because the prefix "SAMPLE_" does not match case-inseinsitive.

```
in:
  type: file
  path_prefix: WORK/CSV/SAMPLE_
```

This is quick hack. But, it should be specified a config parameter `file_name_prefix_ignore_case`.
How do you think about it?

```
diff --git a/embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java b/embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
index 4e1cbfe7..4f2a0072 100644
--- a/embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
@@ -42,6 +42,10 @@ public class LocalFileInputPlugin implements FileInputPlugin {
         @ConfigDefault("null")
         Optional<String> getLastPath();
 
+        @Config("file_name_prefix_ignore_case")
+        @ConfigDefault("false")
+        boolean getFileNamePrefixIgnoreCase();
+
         @Config("follow_symlinks")
         @ConfigDefault("false")
         boolean getFollowSymlinks();
@@ -103,6 +107,13 @@ public class LocalFileInputPlugin implements FileInputPlugin {
             int taskCount,
             List<TaskReport> successTaskReports) {}
 
+    boolean pathStartsWith(PluginTask task, Path path, String prefix) {
+        String pathName = path.getFileName().toString();
+
+        boolean ignoreCase = task.getFileNamePrefixIgnoreCase();
+        return pathName.regionMatches(ignoreCase, 0, prefix, 0, prefix.length());
+    }
+
     public List<String> listFiles(PluginTask task) {
         Path pathPrefix = Paths.get(task.getPathPrefix()).normalize();
         final Path directory;
@@ -143,7 +154,7 @@ public class LocalFileInputPlugin implements FileInputPlugin {
                                 parent = CURRENT_DIR;
                             }
                             if (parent.equals(directory)) {
-                                if (path.getFileName().toString().startsWith(fileNamePrefix)) {
+                                if (pathStartsWith(task, path, fileNamePrefix)) {
                                     return FileVisitResult.CONTINUE;
                                 } else {
                                     return FileVisitResult.SKIP_SUBTREE;
[...]

PR #1022, opened 2018-06-18
Match with path_prefix in the proper way of the file system in LocalFileInputPlugin

jca02266:
For example, on MS Windows, the path_prefix is matched case-insensitively.

This PR has been already discussed on the issue #1021.
But you can drop this PR because of some reason.

Thanks.

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
@@ -6,10 +6,12 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.FileSystems;
 import java.nio.file.FileVisitOption;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.PathMatcher;
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
@@ -107,13 +109,17 @@ public class LocalFileInputPlugin implements FileInputPlugin {
         Path pathPrefix = Paths.get(task.getPathPrefix()).normalize();
         final Path directory;
         final String fileNamePrefix;
+        final PathMatcher fileNamePrefixMatcher;
         if (Files.isDirectory(pathPrefix)) {
             directory = pathPrefix;
             fileNamePrefix = "";
+            fileNamePrefixMatcher = null;
         } else {
             fileNamePrefix = pathPrefix.getFileName().toString();
             Path d = pathPrefix.getParent();
             directory = (d == null ? CURRENT_DIR : d);
+
+            fileNamePrefixMatcher = buildPathMatcherForFileNamePrefix(fileNamePrefix);
         }
 
         final ImmutableList.Builder<String> builder = ImmutableList.builder();
@@ -143,7 +149,7 @@ public class LocalFileInputPlugin implements FileInputPlugin {
                                 parent = CURRENT_DIR;
                             }
                             if (parent.equals(directory)) {
-                                if (path.getFileName().toString().startsWith(fileNamePrefix)) {
+                                if (fileNamePrefixMatcher == null || fileNamePrefixMatcher.matches(path.getFileName())) {
                                     return FileVisitResult.CONTINUE;
                                 } else {
                                     return FileVisitResult.SKIP_SUBTREE;
@@ -175,7 +181,7 @@ public class LocalFileInputPlugin implements FileInputPlugin {
                                 parent = CURRENT_DIR;
                             }
                             if (parent.equals(directory)) {
-                                if (path.getFileName().toString().startsWith(fileNamePrefix)) {
+                                if (fileNamePrefixMatcher == null || fileNamePrefixMatcher.matches(path.getFileName())) {
                                     builder.add(path.toString());
                                     return FileVisitResult.CONTINUE;
                                 }
@@ -213,5 +219,30 @@ public class LocalFileInputPlugin implements FileInputPlugin {
                 return Exec.newTaskReport();
             }
         };
+    }
+
+    private PathMatcher buildPathMatcherForFileNamePrefix(String fileNamePrefix) {
+        StringBuilder sb = new StringBuilder();
+        final int len = fileNamePrefix.length();
+        for (int i = 0; i < len; i++) {
+            // Escape the special characters for the FileSystem#getPathMatcher().
+            // See https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-
+            char c = fileNamePrefix.charAt(i);
+            switch (c) {
+                case '*':
+                case '?':
+                case '{':
+                case '}':
+                case '[':
+                case ']':
+                case '\\':
+                    sb.append('\\');
+                    break;
+                default:
+                    break;
+            }
+            sb.append(c);
+        }
+        return FileSystems.getDefault().getPathMatcher("glob:" + sb.toString() + "*");
     }
 }
--- a/embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
+++ b/embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
@@ -44,10 +44,13 @@ public class TestLocalFileInputPlugin {
 
         // It intentionally tests in the platform-aware way, not in the platform-oblivious way.
         if (System.getProperty("os.name").contains("Windows")) {
-            assertEquals(3, files.size());
+            assertEquals(6, files.size());
             assertTrue(files.contains(buildPath("foofoo1")));
+            assertTrue(files.contains(buildPath("FooFoo2")));
             assertTrue(files.contains(buildPath("foofoo4\\foo")));
             assertTrue(files.contains(buildPath("foofoo4\\bar")));
+            assertTrue(files.contains(buildPath("fooFoo6\\foo")));
+            assertTrue(files.contains(buildPath("fooFoo6\\bar")));
         } else if (System.getProperty("os.name").contains("Mac OS")) {
             assertEquals(3, files.size());
             assertTrue(files.contains(buildPath("foofoo1")));
@@ -75,6 +78,7 @@ public class TestLocalFileInputPlugin {
         this.workdir.newFile("directory1/foo1");
         this.workdir.newFile("directory1/foo2");
         this.workdir.newFile("directory1/Foo3");
+        this.workdir.newFile("directory1/bar");
         this.workdir.newFolder("directory2");
         this.workdir.newFile("directory2/bar");
         final LocalFileInputPlugin plugin = new LocalFileInputPlugin();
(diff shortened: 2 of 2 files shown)


dmikurube:
@jca02266 Thanks for your contribution! But I'm afraid it may need some time to start code review cycles. Sorry about that.

embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
```diff
@@ -103,17 +105,44 @@ public void cleanup(TaskSource taskSource,
             int taskCount,
             List<TaskReport> successTaskReports) {}
 
+    String escapeGlobPattern(String filenamePrefix) {
```
dmikurube: Can you :
* Make this method `private`
* Make it return `PathMatcher` directly
* Rename it to `buildPathMatcherForFileNamePrefix`
* Move it at the bottom of this class

embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
```diff
@@ -103,17 +105,44 @@ public void cleanup(TaskSource taskSource,
...
+        StringBuilder sb = new StringBuilder();
+        final int len = filenamePrefix.length();
+        for (int i = 0; i < len; i++) {
+            char c = filenamePrefix.charAt(i);
+            switch (c) {
```
dmikurube: Can you have a comment with a link to  https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String- to describe these special characters are sufficient?

embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
```diff
@@ -103,17 +105,44 @@ public void cleanup(TaskSource taskSource,
...
+            sb.append(c);
+        }
+        return sb.toString();
+    }
+
     public List<String> listFiles(PluginTask task) {
         Path pathPrefix = Paths.get(task.getPathPrefix()).normalize();
         final Path directory;
         final String fileNamePrefix;
+        final PathMatcher pathMatcher;
```
dmikurube: Can you name this local variable `fileNamePrefixMatcher` ?

dmikurube:
Plus, I'm going to add a test for `LocalFileInputPlugin` in #1023 (after adding Windows special handling in directory names in the test).

After merging #1023, wondering if you could update the test to add Windows special handling in base file names as well.

embulk-standards/src/main/java/org/embulk/standards/LocalFileInputPlugin.java
jca02266: I agree.

That comment message has a typo?

```
s/case-insensitively/case-sensitively/
```

dmikurube:
Ah, sorry, my commit message above was incorrect. `s/case-insensitively/case-sensitively/`

```
Before this fix, it matched case-sensitively even on Windows, for example.
```

jca02266:
Thanks for your many suggestion. Can I really rebase those commit?

jca02266:
If you mean that I should rebase first commit only to fix message, Should I remake new PR?

dmikurube:
Yes, rebasing and squashing is okay here in this PR.

jca02266:
Ok, I will squash all tomorrow.

dmikurube:
Yeah, I will ask rebasing again after #1023 is merged, and update the tests there. Tomorrow would be fine.

jca02266:
Please review it, and then I squash again.

embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
```diff
@@ -0,0 +1,173 @@
...
+        // It intentionally tests in the platform-aware way, not in the platform-oblivious way.
+        if (System.getProperty("os.name").contains("Windows")) {
+            assertEquals(6, files.size());
+            assertTrue(files.contains(buildPath("foofoo1")));
+            assertTrue(files.contains(buildPath("FooFoo2")));
+            assertTrue(files.contains(buildPath("foofoo4\\foo")));
+            assertTrue(files.contains(buildPath("foofoo4\\bar")));
+            assertTrue(files.contains(buildPath("fooFoo6\\foo")));
+            assertTrue(files.contains(buildPath("fooFoo6\\bar")));
+        } else if (System.getProperty("os.name").contains("Mac OS")) {
```
dmikurube: I expect Mac OS will need the same care with Windows. (The default file system on Mac OS handles file names case-insensitively.)

dmikurube:
@jca02266 #1023 has been merged!

And please update the PR subject to be the same with the git commit message.

embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
jca02266: Hmm, however, the test at f6d7ef5 passed on my Mac enviroment.

```console
$ uname -a
Darwin baymax.local 17.6.0 Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64 x86_64
```

`System.getProperty("os.name")` return the `Mac OS X`

```console
$ javac -version
javac 1.8.0_131
```

dmikurube: @jca02266 Got it. Reproduced on my environment as well... But if so, it means that this fix is not perfect yet. It is still working inconsistently on Mac OS between the "directory name" part and the "base file name" part.

For the `testListFilesWithDifferentCaseDirectoryPrefix` case:
* `prefix_path:"Directory1/foo"` matched with `directory1/foo1` and `directory1/foo2`
* `prefix_path:"Directory1/foo"` didn't match with `directory1/Foo3`.

We at least need to confirm why this code is working like that before accepting/merging this change.

jca02266: Hmm, `FileSystems.getDefault().getPathMatcher().getClass().getName()` return the `sun.nio.fs.UnixFileSystem$3`.

`FileSystems.getDefault().getPathMatcher(INCORRECT_GLOB_PATTERN)` show following stacktrace:

```
Caused by: java.util.regex.PatternSyntaxException: Missing '] near index 7
[foofoo*
       ^
	at sun.nio.fs.Globs.toRegexPattern(Globs.java:146)
	at sun.nio.fs.Globs.toUnixRegexPattern(Globs.java:211)
	at sun.nio.fs.UnixFileSystem.getPathMatcher(UnixFileSystem.java:294)
...
```

dmikurube: On the other hand, IIUC, Java on Mac OS counts case-insensitively-matched directories in `Files.walkFileTree`, right?

In summary on Mac OS (in my understanding):
* `Files.walkFileTree` includes directories with different cases
* `FileSystems.getDefault().getPathMatcher` does not include directories with different cases

dmikurube: It looks coming from Java's internal implementation:
* `Files.walkFileTree`: (estimated to be) calling the operating system's tree walking API
* `PathMatcher`: matching processed in Java by building `java.util.regex.Pattern` in [`MacOSXFileSystem`](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/43cb25339b55/src/solaris/classes/sun/nio/fs/MacOSXFileSystem.java#l49) in a case-sensitive way

----

To make it at least "consistent", I'd suggest:

* `Files.walkFileTree` is still used in the same way for listing candidates.
* But, `PathMatcher` is applied to the entire path name, not just the base file name part.

By this way, the matching works at least consistent. Case-sensitive matching is applied in Mac OS X , though.

What do you think?

jca02266: I'm considering and trying to imprement your suggestion.
Please wait for a while.
At least, it may become complicated to escape glob meta character because '\' is meta and windows path separater.

jca02266: And If it complicate, I think that the current changes is best. At least, on Windows, it become well. On the other, on Mac, it is not changed. The case-sensitive behaviour of Mac is java's problem.

dmikurube: > And If it complicate, I think that the current changes is best. At least, on Windows, it improved. On the other, on Mac, it is not changed. The case-sensitive behaviour of Mac is java's problem.

It'd be acceptable, and I don't think the behavior of the current implementation on Mac is Java's problem. It's Embulk's problem. The problem is because Embulk is using a combination of two different methods to identify files from a prefix string. After we use only `PathMatcher` to identify files, it'd be acceptable to say "It's Java's problem".

I don't think this will be a dangerous change for Mac users because it causes neither any partial success nor any unintended pick-ups. When this change affects, it fails at all. Users can find it out.

jca02266: real path: /foo
path_prefix: /Foo

In this case, at first, embulk open the directory "Foo" , The Path object has the name "Foo" instead of "foo".
I think We must only use the realpath() system call to get the case sensitive real path, however, realpath() will be resolve symbolic link. So we can't use realpath() because If /foo is linked to /bar, realpath() return the "/bar".

> PathMatcher is applied to the entire path name, not just the base file name part.

Simply, I think we can't do that.

```
public class Sample {
  @Rule
  public TemporaryFolder temp = new TemporaryFolder();

  @Test
  public void test() throws Exception {
    final Path parent = temp.getRoot().toPath();

    temp.newFolder("foo", "BAR");
    temp.newFile("foo/BAR/Baz");

    Files.createSymbolicLink(Paths.get(parent.toString(), "foo/qux"),
                             Paths.get(parent.toString(), "foo/BAR"));

    Path first = Paths.get(parent.toString(), "Foo");

    assertThat(first.toString(), endsWith("Foo"));
    assertThat(first.toRealPath().toString(), endsWith("foo"));
    // If it is passed LinkOption.NOFOLLOW_LINKS, realpath() is not called
    assertThat(first.toRealPath(LinkOption.NOFOLLOW_LINKS).toString(), endsWith("Foo")); // "foo" is expected, but "Foo"

    Path second = Paths.get(parent.toString(), "Foo/qux");

    assertThat(second.toString(), endsWith("Foo/qux"));
    // If Foo/qux is symbolic link, it is resolved.
    assertThat(second.toRealPath().toString(), endsWith("foo/BAR")); // "foo/qux" is expected, but "foo/BAR"
    assertThat(second.toRealPath(LinkOption.NOFOLLOW_LINKS).toString(), endsWith("Foo/qux")); // "foo/qux" is expected, but "Foo/qux"

    EnumSet<FileVisitOption> option = EnumSet.noneOf(FileVisitOption.class);
    int maxDepth = Integer.MAX_VALUE;

    Files.walkFileTree(first, option, maxDepth, new SimpleFileVisitor<Path>() {
      @Override
      public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
        System.out.println("path: " + file.toString());
        System.out.println("real: " + file.toRealPath().toString());
        return FileVisitResult.CONTINUE;
      }
    });
  }
}
```

dmikurube: Sorry, I don't get your point... What do you want to point/claim?

jca02266: oh, sorry.

> Files.walkFileTree is still used in the same way for listing candidates.
> But, PathMatcher is applied to the entire path name, not just the base file name part.

I tried to implemente above, and I found that it is imposible. 

1. We must get the real path (real case)  before applying PathMatcher.
2. We can use the realpath() system call to get the real path.
3. However, realpath() might also resolve symbolic link to real path, so PathMatcher can never match such the path. 

Am I making sense?

If you can't make sense yet (probably, my English is bad), I will implement above code.

dmikurube: @jca02266 Ah, okay... I eventually got your point after some experiments.

Finally, let us accept (merge) the current implementation of yours as the second best choice for the reason you mentioned :
> At least, on Windows, it improved. On the other, on Mac, it is not changed.

But let us consider a bit more whether we'd release this as the next release `0.9.8` or not.

jca02266:
No problem. Thanks a lot.

dmikurube:
Thanks. Merging.

PR #1023, opened 2018-06-26
Add TestLocalFileInputPlugin

dmikurube:
~~4692a47f32aa1c2d0c87258f29928fcdd0725320 is expected to fail on Windows.~~

@sakama @kamatama41 Can you take a look at these tests? It is to check the behavior before and after #1022 (from issue #1021). The behavior would change after #1022 is merged.

### DIFF
--- a/embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
+++ b/embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
@@ -74,6 +74,7 @@ public class TestLocalFileInputPlugin {
         this.workdir.newFolder("directory1");
         this.workdir.newFile("directory1/foo1");
         this.workdir.newFile("directory1/foo2");
+        this.workdir.newFile("directory1/Foo3");
         this.workdir.newFolder("directory2");
         this.workdir.newFile("directory2/bar");
         final LocalFileInputPlugin plugin = new LocalFileInputPlugin();
@@ -112,6 +113,7 @@ public class TestLocalFileInputPlugin {
         this.workdir.newFolder("directory1");
         this.workdir.newFile("directory1/foo1");
         this.workdir.newFile("directory1/foo2");
+        this.workdir.newFile("directory1/Foo3");
         this.workdir.newFolder("directory2");
         this.workdir.newFile("directory2/bar");
         final LocalFileInputPlugin plugin = new LocalFileInputPlugin();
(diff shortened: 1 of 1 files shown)


jca02266:
This test exactly failed on Windows. But, I am wondering if it is what you expected

```
diff --git a/embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java b/embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
index 2625ebbe..c3da43e5 100644
--- a/embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
+++ b/embulk-standards/src/test/java/org/embulk/standards/TestLocalFileInputPlugin.java
@@ -43,8 +43,8 @@ public class TestLocalFileInputPlugin {
         final List<String> files = plugin.listFiles(task);
         assertEquals(3, files.size());
         assertTrue(files.contains(buildPath("foofoo1")));
-        assertTrue(files.contains(buildPath("foofoo4/foo")));
-        assertTrue(files.contains(buildPath("foofoo4/bar")));
+        assertTrue(files.contains(buildPath("foofoo4", "foo")));
+        assertTrue(files.contains(buildPath("foofoo4", "bar")));
     }
 
     private LocalFileInputPlugin.PluginTask buildTask(
@@ -68,13 +68,15 @@ public class TestLocalFileInputPlugin {
         return config.loadConfig(LocalFileInputPlugin.PluginTask.class);
     }
 
-    private String buildPath(final String subPath) {
+    private String buildPath(final String... subPaths) {
         final String workdirPath = this.workdir.getRoot().getPath();
         final StringBuilder pathPrefixBuilder = new StringBuilder(workdirPath);
+        for (String subPath: subPaths) {
             if (!workdirPath.endsWith(File.separator) && !subPath.startsWith(File.separator)) {
                 pathPrefixBuilder.append(File.separator);
             }
             pathPrefixBuilder.append(subPath);
+        }
         return pathPrefixBuilder.toString();
     }
 }
```

dmikurube:
That is one point, but the other point I wanted to confirm is that capital "directory" names match compared to Linux where capital directory names don't match.

dmikurube:
Ah... that's not the case. That case-insensitive "directory" part is in `path_prefix`. Will add such test.

dmikurube:
Finally these are the tests we need. These will fail after #1022 is merged (without updating the tests).

dmikurube:
Thanks!

PR #1038, opened 2018-09-26
Clean up EmbulkEmbed and EmbulkService

dmikurube:
This is just refactoring. Can you have a look, @sakama @kamatama41 ?

* Move `EmbulkService.standardModuleList` to `EmbulkEmbed`
* Reduce usage of Guava
* Use Java 8 `java.util.function.Function` instead of Guava `Function`

----

This will use https://github.com/embulk/guice-bootstrap/pull/12.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
+++ b/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
@@ -1,116 +1,40 @@
 package org.embulk;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkState;
-
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.base.Function;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.inject.Injector;
 import com.google.inject.Module;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.function.Function;
 import org.embulk.config.ConfigDiff;
 import org.embulk.config.ConfigLoader;
 import org.embulk.config.ConfigSource;
 import org.embulk.config.ModelManager;
 import org.embulk.exec.BulkLoader;
+import org.embulk.exec.ExecModule;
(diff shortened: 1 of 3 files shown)


embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
```diff
@@ -1,165 +1,240 @@
...
 import org.embulk.exec.ExecutionResult;
+import org.embulk.exec.ExtensionServiceLoaderModule;
 import org.embulk.exec.GuessExecutor;
 import org.embulk.exec.PartialExecutionException;
 import org.embulk.exec.PreviewExecutor;
 import org.embulk.exec.PreviewResult;
 import org.embulk.exec.ResumeState;
+import org.embulk.exec.SystemConfigModule;
 import org.embulk.exec.TransactionStage;
 import org.embulk.guice.Bootstrap;
```
kamatama41: It seems the import is not used (since past version).

dmikurube: Ah, good catch! Thanks.

embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
```diff
@@ -1,165 +1,240 @@
...
-        public Bootstrap addModules(Iterable<? extends Module> additionalModules) {
-            final List<Module> copy = ImmutableList.copyOf(additionalModules);
-            return overrideModules(new Function<List<Module>, Iterable<Module>>() {
+        public Bootstrap addModules(final Iterable<? extends Module> additionalModules) {
+            final ArrayList<Module> copyMutable = new ArrayList<>();
+            for (final Module module : additionalModules) {
+                copyMutable.add(module);
+            }
+            final List<Module> copy = Collections.unmodifiableList(copyMutable);
+            return overrideModules(new com.google.common.base.Function<List<Module>, Iterable<Module>>() {
```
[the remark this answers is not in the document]
dmikurube: Good catch! I mechanically replaced Guava Function to Java 8 Function, but that'd be better.

dmikurube:
Thanks! Merging this with 22e5b754188df0ccfd83341f1a647e7f509f895b after the CI gets happy.

PR #1039, opened 2018-09-27
[WIP] Support specifying lime delimiter character.

kamatama41:


embulk-core/src/main/java/org/embulk/spi/util/LineDecoder.java
```diff
@@ -135,4 +145,84 @@ public void remove() {
...
         }
     }
+
+    /**
+     * A {@link BufferedReader} that can specify line delimiter character from any one of CR, LF and CRLF.
+     * If not specified, follow the original {@link BufferedReader}'s behavior.
+     */
+    private static class LineReader extends BufferedReader {
+        private static final int UNREAD = -1;
+        private final Newline lineDelimiter;   // should create another enum?
```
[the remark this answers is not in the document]
kamatama41: Created new enum at 51ae1c0

embulk-core/src/main/java/org/embulk/spi/util/LineDecoder.java
```diff
@@ -135,4 +145,84 @@ public void remove() {
...
+        private final char[] buffer;
+        private int offset = UNREAD;
+        private int charsRead = 0;
```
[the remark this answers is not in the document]
kamatama41: Fixed at 
74d1036

embulk-core/src/main/java/org/embulk/spi/util/LineDecoder.java
```diff
@@ -2,14 +2,18 @@
...
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.io.Reader;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CodingErrorAction;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
+
```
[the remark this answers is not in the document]
kamatama41: Fixed at b3b6cbb

embulk-core/src/main/java/org/embulk/spi/util/LineReader.java
```diff
@@ -0,0 +1,105 @@
...
+ * This class is not thread-safe.
+ */
+class LineReader extends BufferedReader {
+    private static final int UNREAD = -1;
+    private final LineDelimiter lineDelimiter;
+    private final char[] buffer;
+    private int offset;
+    private int charsRead;
+
+    LineReader(Reader reader, LineDelimiter lineDelimiter, int bufferSize) {
```
dmikurube: How about making this constructor `private`, and having another `static` initializer method (`.of` or `.create`, or else) to:

1. Return `BufferedReader` created as-is if `lineDelimiter` is `null`.
2. Return `LineReader` if `lineDelimiter` is not `null`.

So that we have less impacts by `LineReader` for existing configurations?

kamatama41: Hmm, since current LineReader without "lineDelimiter" behaves completely same as BufferedReader, I don't think it makes less impact to current config.

dmikurube: Sorry, I meant making it sure and clearer for those who read this code (readers). A simple `BufferedReader` instance returned is clearer for readers, rather than `LineReader` instance with `if (lineDelimiter == null)` check in `readLine`.

kamatama41: Thanks, it's clear now and it makes sense. Fixed at 8317922

kamatama41:
@dmikurube 
I've updated option name and document as discussed here
https://github.com/embulk/embulk/issues/1060#issuecomment-439197304
please take a look again, thank you 🙏

PR #1042, opened 2018-10-05
Refactor LocalFileInputPlugin and its test

dmikurube:
It refactors `LocalFileInputPlugin` after #1040. It's just refactoring, and it shouldn't change any behavior.


dmikurube:
Thanks!

PR #1048, opened 2018-10-11
Build EmbulkEmbed without Life Cycle Management (guice-bootstrap)

dmikurube:
`EmbulkEmbed.Bootstrap#overrideModules` turned out not to be "overriding" standard modules correctly. There must be something wrong in `EmbulkEmbed.Bootstrap#overrideModules`, or guice-bootstrap (even before #1038).

For example, when I tried to override a binding to `ILoggerFactory` over the standard `org.embulk.exec.LoggerProvider` (with Logback) by `EmbulkEmbed.Bootstrap#overrideModules`, `org.embulk.exec.LoggerProvider` was still loaded, and it caused a logging driver conflict in our local environment.

In addition, I couldn't find any `@PostConstruct` and `@PreDestroy` usage with Embulk. Guice-bootstrap may not be required if those annotations are actually not used.

Then, I'm introducing a new `SimpleBootstrap` for trial without guice-bootstrap that overrides correctly. Note that correct overriding may be realized even with guice-bootstrap (probably with fixing guice-bootstrap). But, if those annotations are actually not used, we may also want to eliminate the usage of guice-bootstrap.

I at least confirmed that `org.embulk.exec.LoggerProvider` was correctly overridden with this `SimpleBootstrap`.

----

@sakama @kamatama41 What do you think? At least, this new class does not introduce any incompatibility.

@muga Can I have your thoughts when you have time?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
+++ b/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
@@ -2,10 +2,13 @@ package org.embulk;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.Iterables;
+import com.google.inject.Guice;
 import com.google.inject.Injector;
 import com.google.inject.Module;
+import com.google.inject.util.Modules;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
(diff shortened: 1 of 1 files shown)


dmikurube:
Thanks!

PR #1052, opened 2018-10-26
Make stop_on_invalid_record effective for invalid dates such as 2018-02-31

dmikurube:
We observed that CSV parser's `stop_on_invalid_record: false` was not working as intended when a parsed timestamp was invalid (non-existing) date/time, such as 2018-02-31.

It was because `DateTimeException` in creating a `*DateTime` instance was not caught and wrapped with `TimestampParseException` although `DateTimeParseException` was caught.


dmikurube:
Thanks!

Issue #1031, opened 2018-07-18
"embulk new" command not work in windows 


jo8937:
I have executed command in windows 10 x64

> E:\work> embulk new java-file-output onefile

```
2018-07-18 10:41:05.724 +0900: Embulk v0.9.7
Creating embulk-output-onefile/
java.lang.UnsupportedOperationException
        at sun.nio.fs.WindowsFileSystemProvider.readAttributes(Unknown Source)
        at java.nio.file.Files.readAttributes(Unknown Source)
        at java.nio.file.Files.getPosixFilePermissions(Unknown Source)
        at org.embulk.cli.EmbulkNew.setExecutable(EmbulkNew.java:390)
        at org.embulk.cli.EmbulkNew.newPlugin(EmbulkNew.java:169)
        at org.embulk.cli.EmbulkRun.runSubcommand(EmbulkRun.java:338)
        at org.embulk.cli.EmbulkRun.run(EmbulkRun.java:91)
        at org.embulk.cli.Main.main(Main.java:26)
Failed. Removing the directory created.
```

I have executed same command in cent os6, then success. but not work in windows PC.
JDK version is 1.8.0_171.

PR #1053, opened 2018-10-26
Fix "new" and "migrate" command not work in Windows.

mikoto2000:
See #1031.

NTFS filesystem not support POSIX permission.
I added process in `EmbulkNew#setExecutable` and `EmbulkMigrate#setExecutable`.

1. try/catch of `UnsupportedOperationException`
2. output warning and skip "set executable" process, if `UnsupportedOperationException` was throwd


dmikurube:
@mikoto2000 Thanks for your contribution! Got it. I was not aware of #1031, and your fix looks making it work.

But after some survey, it seemed better for me to use `FileSystem#supportedFileAttributeViews` instead of catching `UnsupportedOperationException` after failing, like I tried in #1054. Can I have your thoughts?

PR #1057, opened 2018-10-30
Check Java versions also for non-Oracle Java: Fix #1056

### LINKED ISSUE #1056: Incorrect Java version detection in OpenJDK.
* OS version: macOS 10.13.6
* Java version: openjdk full version "11.0.1+13"
* Embulk version:  4b8250d8

selfrun java version checker expect the string `java full version xxx`.
But, OpenJDK output `openjdk full version "11.0.1+13"`. 
So, It can't detect Java version 11.

```
./embulk-0.9.9-SNAPSHOT.jar
[WARN] Unrecognized Java version: openjdk full version "11.0.1+13"
```

```bash
    java\ full\ version\ \"11*\")
        echo "[WARN] Embulk does not guarantee running with Java 11." 1>&2
        echo "[WARN] Executing Java with: \"--add-modules java.xml.bind --add-modules=java.se.ee\"" 1>&2
        echo "" 1>&2
        java_args="--add-modules java.xml.bind --add-modules=java.se.ee $java_args"
        ;;
```

dmikurube:
This PR tries to fix #1056. Can you have a look, @sakama @kamatama41 @hiroyuki-sato ?

I assume that the names (`java` or `openjdk`) are lower-case alphabets. When we see other cases, we'll have another fix.

### DIFF
--- a/embulk-core/src/main/bat/selfrun.bat
+++ b/embulk-core/src/main/bat/selfrun.bat
@@ -37,15 +37,15 @@ if "%overwrite_optimize%" == "true" (
 )
 
 for /f "delims=" %%w in ('java -fullversion 2^>^&1') do set java_fullversion=%%w
-echo %java_fullversion% | find "java full version ""1.7" > NUL
+echo %java_fullversion% | find " full version ""1.7" > NUL
 if not ERRORLEVEL 1 (set java_version=7)
-echo %java_fullversion% | find "java full version ""1.8" > NUL
+echo %java_fullversion% | find " full version ""1.8" > NUL
 if not ERRORLEVEL 1 (set java_version=8)
-echo %java_fullversion% | find "java full version ""9" > NUL
+echo %java_fullversion% | find " full version ""9" > NUL
 if not ERRORLEVEL 1 (set java_version=9)
-echo %java_fullversion% | find "java full version ""10" > NUL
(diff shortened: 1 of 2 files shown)


hiroyuki-sato:
* OS: macOS 10.13.6
* Embulk version: 555dabeab
* Java: openjdk full version "11.0.1+13"

Version check works correctly, But I got the following error.

```
%./embulk-0.9.9-SNAPSHOT.jar
[WARN] Embulk does not guarantee running with Java 11.
[WARN] Executing Java with: "--add-modules java.xml.bind --add-modules=java.se.ee"

OpenJDK 64-Bit Server VM warning: Option AggressiveOpts was deprecated in version 11.0 and will likely be removed in a future release.
Error occurred during initialization of boot layer
java.lang.module.FindException: Module java.se.ee not found

% java -fullversion
openjdk full version "11.0.1+13"
```

dmikurube:
@hiroyuki-sato Ah, got it. It looks because `java.se.ee` is removed from pre-install since Java 11.
http://openjdk.java.net/jeps/320

It would need changing packaging of the Embulk core, and that is not in our scope for the time being. I now think about removing Java 11 support from the script for the time being.

dmikurube:
Thanks! Merging it.

Issue #1061, opened 2018-11-17
gem gradle task for Java plugin has warning


kamatama41:
- Embulk: 0.9.9
- OS: MacOS X 10.13.6 (High Sierra)

It might be better to update rake's version in [the plugin template](https://github.com/embulk/embulk/blob/effe8985521311ca7f09864eccd04dd4f1313957/embulk-core/src/main/resources/org/embulk/plugin/template/java/build.gradle.vm#L93).

```
$ embulk9 new java-input foo
2018-11-18 08:34:45.466 +0900: Embulk v0.9.9
Creating embulk-input-foo/

Plugin template is successfully generated.
Next steps:

  $ cd embulk-input-foo
  $ ./gradlew package
$ cd embulk-input-foo
$ ./gradlew gem
> Task :gem
fatal: not a git repository (or any of the parent directories): .git
WARNING:  no homepage specified
WARNING:  open-ended dependency on rake (>= 10.0, development) is not recommended
  if rake is semantically versioned, use:
    add_development_dependency 'rake', '~> 10.0'
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: embulk-input-foo
  Version: 0.1.0
  File: embulk-input-foo-0.1.0.gem


BUILD SUCCESSFUL in 11s
5 actionable tasks: 5 executed
```

kamatama41:
It might be better to update JRuby plugin's template as well.
https://github.com/embulk/embulk/blob/e85d6998d139f65c79bc209207257f410d4e0db9/embulk-core/src/main/resources/org/embulk/plugin/template/ruby/gemspec.vm#L19

PR #1062, opened 2018-11-19
Support only rake 12.x on Java plugin's gemspec

kamatama41:
For #1061 
Open-ended dependency by using `>=` is not recommended by recent `gem` command.
So I made it narrower, only support rake v12.x.

What do you think about this? @sakama  @dmikurube


kamatama41:
thank you for reviewing!

Issue #1064, opened 2018-11-20
Support Java 9+


kamatama41:

kamatama41:
`embulk run` with OpenJDK 9.0.4 causes the following error in my environment.

```
2018-11-20 14:51:00.941 +0900: Embulk v0.9.10-SNAPSHOT
Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
	at org.apache.bval.jsr303.ConfigurationImpl.parseValidationXml(ConfigurationImpl.java:268)
	at org.apache.bval.jsr303.ConfigurationImpl.prepare(ConfigurationImpl.java:258)
	at org.apache.bval.jsr303.ConfigurationImpl.doPrivBuildValidatorFactory(ConfigurationImpl.java:248)
	at org.apache.bval.jsr303.util.SecureActions$5.run(SecureActions.java:131)
	at org.apache.bval.jsr303.util.SecureActions$5.run(SecureActions.java:129)
	at org.apache.bval.jsr303.ConfigurationImpl.run(ConfigurationImpl.java:337)
	at org.apache.bval.jsr303.ConfigurationImpl.buildValidatorFactory(ConfigurationImpl.java:244)
	at org.embulk.config.ModelManager.<init>(ModelManager.java:24)
	at org.embulk.EmbulkEmbed.newSystemConfigLoader(EmbulkEmbed.java:236)
	at org.embulk.EmbulkEmbed$Bootstrap.<init>(EmbulkEmbed.java:94)
	at org.embulk.EmbulkSetup.setup(EmbulkSetup.java:29)
	at org.embulk.cli.EmbulkRun.runSubcommand(EmbulkRun.java:418)
	at org.embulk.cli.EmbulkRun.run(EmbulkRun.java:91)
	at org.embulk.cli.Main.main(Main.java:26)
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.JAXBException
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:582)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:185)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:496)
	... 14 more

```

dmikurube:
How did you run which version?
https://github.com/embulk/embulk/issues/976
https://github.com/embulk/embulk/pull/1049

[...]

PR #1065, opened 2018-11-20
Upgrade Gradle wrapper to the latest

kamatama41:
This is a preparation of #1064 
Current Gradle version (4.1) doesn't work with Java 9+, it's needed to be `4.2.1` or later.

@dmikurube @sakama 
What do you think about this?


dmikurube:
Sorry, I didn't understand this -- Gradle is for building, and #1064 is for running. They don't look related to me, and actually we have already confirmed that released versions (0.9.9) run with Java 9 (with some warning messages).

dmikurube:
But anyways, I basically agree to have later Gradle.

One concern is that later Gradle sometimes needs Gradle plugins to be updated at the same time. And, we sometimes need to update `build.gradle` along with the plugin upgrade. Have you checked existing Gradle tasks working as expected?

kamatama41:
Sorry for my lack explanation. I tried to build the project (and plugin project) with OpenJDK 9.0.4 then the following error was caused. So I wanted to upgrade Gradle.

```shell
$ ./gradlew package                                                                                                                                                                  [embulk-input-foo2]

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine java version from '9.0.4'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org
```

kamatama41:
As I communicated with @dmikurube in another channel about this PR, current build environment should be kept until we actually start to support Java 9+. So I will close this.

dmikurube:
We can leave it open. Once the situation gets settled, we may want to upgrade Gradle version unless we upgrade Java version to build.

PR #1068, opened 2018-11-29
Parse exceeding dates like 2018-02-31 only in the legacy timestamp parser, not in the "ruby:" timestamp parser

dmikurube:
@sakama @kamatama41 As discussed directly,
* We once implement Ruby-compatible behavior with 2018-02-31 (to be 2018-03-03) in v0.9.10.
* Then, we will reject 2018-02-31 in later versions.

Toward this goal, I finally thought that we need this Ruby-compatible behavior only in the "legacy" timestamp parser. I'm reverting a change in #1067 on the "ruby:" timestamp parser.

Can you have a look?

----

JFYI: Embulk has three types of timestamp parsers since v0.9.0.

* Legacy: specified by `format: %Y-%m-...`. It aims to be compatible with Embulk v0.8.
    * We have compatible behavior about 2018-02-31 in v0.9.10.
    * We'll give up compatibility in later versions.
* Ruby: specified like `format: "ruby:%Y-%m-..."`. It aims to be more compliant with original Ruby's `Time.strptime`.
    * We'll have different behavior about 2018-02-31, though.
* Java: specified like `format: "java:uuuu-MM-..."`. It is based on `java.time.format.DateTimeFormatter`.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/time/RubyTimeParsed.java
+++ b/embulk-core/src/main/java/org/embulk/spi/time/RubyTimeParsed.java
@@ -730,6 +730,34 @@ class RubyTimeParsed extends TimeParsed {
                     (this.nanoOfSecond == Integer.MIN_VALUE ? 0 : this.nanoOfSecond),
                     zoneId).withDayOfYear(this.dayOfYear).plusDays(daysRollover);
         } else {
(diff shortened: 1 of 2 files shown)


dmikurube:
Thanks!

PR #1075, opened 2018-12-13
Warn if LifeCycleManager is in use

dmikurube:
I think about removing the LifeCycle support from Embulk. That feature is never used as far is I know.

Before removing that feature, trying to message it to users to expect users report to us when they're using it.


dmikurube:
Thanks! Merging... JFYI, the removal will be done like #1078.

PR #1076, opened 2018-12-13
Cleanup codes

kamatama41:
- Replace all lambda-able anonymous inner classes to lambda c840131
- Remove unnecessary modifiers in interface and enum 9b4b212
- Remove unnecessary type specifications 3442efa

Although this PR is apparently a big change, it doesn't modify any behavior.


kamatama41:
Thank you for the detailed explanation. It makes sense to me.
So I will close this and keep in mind.

PR #1077, opened 2018-12-13
Warn for EmbulkEmbed#destroy

dmikurube:
After #1075, warning for `EmbulkEmbed#destroy` as well. The method is effective only for  `@PreDestroy` methods, but as mentioned in #1075, we don't observe actual usages of `@PreDestroy` around Embulk.

Can you have a look? @sakama @kamatama41


dmikurube:
Thanks!

PR #1079, opened 2018-12-14
Silence LoggerProvider, and configure Logback only when Embulk is executed from CLI

dmikurube:
The details are described in `docs/design/slf4j.md`.

It won't be merged very soon for v0.9.12. It targets later versions such as v0.9.14 or v0.9.15 so that we can confirm the impact of this change before merging.

Can you have a look? @sakama @kamatama41

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -185,8 +185,9 @@ dependencies {
     compile project(':embulk-core')
     compile project(':embulk-standards')
 
-    // Logback is included only in the executable package.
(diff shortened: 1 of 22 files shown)


dmikurube:
Updated the location to initialize Logback: 68658c8a18e22ca13f528cda3452adbdd83f6d05

Logback is initialized in the very beginning of the CLI bootstrap process. @sakama @kamatama41

dmikurube:
Confirmed it looks working fine. Let me merge this, and release v0.9.14 soon.

PR #1083, opened 2018-12-19
Prepare for v0.9.12

dmikurube:
Releasing v0.9.12. Can you have a look, @sakama @kamatama41?

### DIFF
--- a/embulk-docs/src/release/release-0.9.12.rst
+++ b/embulk-docs/src/release/release-0.9.12.rst
@@ -1,0 +1,13 @@
+Release 0.9.12
+==================================
+
+General Changes
+----------------
+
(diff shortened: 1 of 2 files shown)


embulk-docs/src/release/release-0.9.12.rst
```diff
@@ -0,0 +1,13 @@
+Release 0.9.12
+==================================
+
+General Changes
+----------------
+
+* Display file name and file size in log to easier debug which is loaded by FileInputPlugin [#1066]
```
sakama: #1066 shows only file name, don't show file size

dmikurube: Ah, right. Thanks for the good catch! (I wrongly copied it from older git commit message.)

dmikurube:
Releasing v0.9.12!

PR #1084, opened 2018-12-20
Update links in docs to the list of plugins and the Input Plugin Development Guide

dmikurube:
Forgot that https://www.embulk.org/docs has been updated through CircleCI by compiling the `embulk-docs` files... https://github.com/embulk/www.embulk.org/pull/2 and https://github.com/embulk/www.embulk.org/pull/3 were overwritten.

Updating this `embulk-docs` here would update the document. Can you have a look? @sakama @kamatama41

### DIFF
--- a/embulk-docs/src/index.rst
+++ b/embulk-docs/src/index.rst
@@ -55,23 +55,23 @@ Documents
 
    recipe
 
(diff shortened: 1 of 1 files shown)


dmikurube:
Thanks! Will merge.

PR #1085, opened 2018-12-27
Isolate Maven-related dependencies to sub-ClassLoader so that they are hidden from plugins

dmikurube:
It's a big change ... so not in a hurry.

As the first step to hide core's dependency libraries from plugins, this PR tries to hide Apache Maven-related libraries by a separate `ClassLoader`.

In CLI, the executable JAR includes Maven-related JAR files as-is as resources, not with extracting the JARs. They're loaded through a separate `ClassLoader`.

In `EmbulkEmbed`, users (developers) need to call a initializer static method to load Maven-related libraries. (Implementations for CLI (`org.embulk.cli....`) would be hints.)

Can you have a look, @sakama @kamatama41 ? Ask me anything directly if you have any questions. (It may be too complicated to discuss on the text boxes...)


dmikurube:
Then, I believe we're ready to merge this. Let me merge!

PR #1091, opened 2019-01-24
Leave release notes in Git annotated tags

dmikurube:
Releasing a new version of Embulk has been annoying. We had to create a release note in the Git repo (`embulk-docs`), with the right release date, merge that to `master`, and then run `./gradlew release`. Do we really need release notes / change logs as a file in the Git repo?

Instead, Git has annotated tags. GitHub shows annotated tags as "GitHub Releases". That's easier to maintain, that has less commits needed to release.

I've already annotated existing tags. In this PR, we're
* Removing the table of contents of release notes from `index.rst`.
* Leaving only links to GitHub Releases in `release-x.y.z.rst`.
* Remove the checks in `build.gradle`.


dmikurube:
Thanks! Let's apply this, and see embulk.org.

PR #1092, opened 2019-01-24
Fix reStructuredText links for change logs

dmikurube:
Ugh, #1091 was wrong in the link format of reStructuredText. :(  ('`_`' was missing at the end.)

Fixing the mistake here.

@sakama @kamatama41 Can you have a look when you have time?


dmikurube:
Thanks!

PR #1096, opened 2019-01-30
Add an option into JsonParser, which can identify a specific value by JSON Pointer.

kamatama41:
### Usage
```java
JsonParser parser = new JsonParser();
String pointer = "/a/b";
// The value will be "1"
Value value = parser.parse("{\"a\": {\"b\": 1}}", pointer);
```

```java
JsonParser parser = new JsonParser();
String pointer = "/a/b";
InputStream input = toInputStream("{\"a\": {\"b\": 1}}{\"a\": {\"b\": 2}}");
// The stream will return 2 values that are "1" and "2"
JsonParser.Stream stream = parser.open(input, pointer);
```


embulk-core/src/main/java/org/embulk/spi/json/JsonParser.java
```diff
@@ -45,14 +55,19 @@ private static String sampleJsonString(String json) {
...
-            super(createParser(factory, in));
+        public StreamParseContext(JsonFactory factory, InputStream in, String pointer) throws IOException {
+            super(createParser(factory, in, pointer));
         }
 
-        private static com.fasterxml.jackson.core.JsonParser createParser(JsonFactory factory, InputStream in)
-                throws IOException {
+        private static com.fasterxml.jackson.core.JsonParser createParser(
+                JsonFactory factory, InputStream in, String pointer
+        ) throws IOException {
```
dmikurube: nitpick: Ah, I've never seen this style, but our checkstyle accepted this... ;)

I usually have done it like:
```
        private static com.fasterxml.jackson.core.JsonParser createParser(
                JsonFactory factory, InputStream in, String pointer) throws IOException {
```
or,

```
        private static com.fasterxml.jackson.core.JsonParser createParser(
                JsonFactory factory,
                InputStream in,
                String pointer)
                throws IOException {
```

embulk-core/src/main/java/org/embulk/spi/json/JsonParser.java
```diff
@@ -29,11 +31,19 @@ public JsonParser() {
     }
 
     public Stream open(InputStream in) throws IOException {
-        return new StreamParseContext(factory, in);
+        return open(in, null);
+    }
+
+    public Stream open(InputStream in, String pointer) throws IOException {
```
dmikurube: Possibly nice to have a more descriptive variable name what the `pointer` is for.

embulk-core/src/main/java/org/embulk/spi/json/JsonParser.java
```diff
@@ -45,14 +55,19 @@ private static String sampleJsonString(String json) {
...
 
-        private static com.fasterxml.jackson.core.JsonParser createParser(JsonFactory factory, InputStream in)
-                throws IOException {
+        private static com.fasterxml.jackson.core.JsonParser createParser(
+                JsonFactory factory, InputStream in, String pointer
+        ) throws IOException {
             try {
-                return factory.createParser(in);
+                final com.fasterxml.jackson.core.JsonParser baseParser = factory.createParser(in);
+                if (pointer == null || pointer.isEmpty()) {
```
dmikurube: We have no reasons to hesitate using `Optional`, especially in `private` regions.

embulk-core/src/main/java/org/embulk/spi/json/JsonParser.java
kamatama41: Got it. How about `jsonPointer`, `jsonPointerExpression` or `pointerExpression`?

embulk-core/src/main/java/org/embulk/spi/json/JsonParser.java
kamatama41: OK, I fixed the style to the first one.

embulk-core/src/main/java/org/embulk/spi/json/JsonParser.java
kamatama41: Fixed in e9f4f9d

embulk-core/src/main/java/org/embulk/spi/json/JsonParser.java
kamatama41: I see, thank you for the detailed explanation. So I will change it to `offsetInJsonPointer`.

dmikurube: How about changing the method name? (As commented, I'd suggest that way more.)

kamatama41: It makes sense. So I will rename the method.

PR #1097, opened 2019-01-30
Check Git tag status to release

dmikurube:
After #1091, we have had no checks when releasing. This PR checks the Git `HEAD` is in a correct status before running `./gradlew release`.


dmikurube:
Thanks!

PR #1099, opened 2019-02-01
[Experimental] Add an option to point a specific value within a JSON document via Json Pointer expression.

kamatama41:
As we can parse only a specific JSON object within a JSON document by JsonParser since #1096, this option uses it.

## Usage
Sample JSON
```json
{"a": {"b": 1}}
```
Config YAML
```yaml
in:
  type: ...
  parser:
    type: json
    __experimental__json_pointer_to_root: "/a"
```
Result
```
+-------------+
| record:json |
+-------------+
|     {"b":1} |
+-------------+
```

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
@@ -35,6 +35,10 @@ import org.slf4j.LoggerFactory;
 
 public class JsonParserPlugin implements ParserPlugin {
 
+    public JsonParserPlugin() {
+        this.jsonParser = new JsonParser();
+    }
+
     public enum InvalidEscapeStringPolicy {
         PASSTHROUGH("PASSTHROUGH"),
         SKIP("SKIP"),
@@ -147,9 +151,9 @@ public class JsonParserPlugin implements ParserPlugin {
         }
 
         if (task.getJsonPointerToRoot().isPresent()) {
-            return JSON_PARSER.openWithOffsetInJsonPointer(inputStream, task.getJsonPointerToRoot().get());
+            return jsonParser.openWithOffsetInJsonPointer(inputStream, task.getJsonPointerToRoot().get());
         } else {
-            return JSON_PARSER.open(inputStream);
+            return jsonParser.open(inputStream);
         }
     }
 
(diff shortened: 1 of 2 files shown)


embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -207,4 +221,6 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output) {
     private static final Logger logger = LoggerFactory.getLogger(JsonParserPlugin.class);
 
     private static final Pattern DIGITS_PATTERN = Pattern.compile("\\p{XDigit}+");
+
+    private static final JsonParser JSON_PARSER = new JsonParser();
```
dmikurube: It's not an immediate problem because the existing `JsonParser` is immutable, but we may have to change the situation in the future. We can easily forget to update this part in that case.

I'd suggest to declare this as a non-static final member, and initialize it in the constructor. What do you think?

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -57,6 +59,10 @@ public String getString() {
         @Config("invalid_string_escapes")
         @ConfigDefault("\"PASSTHROUGH\"")
         InvalidEscapeStringPolicy getInvalidEscapeStringPolicy();
+
+        @Config("json_pointer_to_root")
```
dmikurube: As discussed f2f, we may change the configuration name later before finalizing the release of this feature. And, we may release Embulk before the finalization.

What do you think:
* Add TODO comment to rename it, and
* Name this configuration so that it is clear this is still "experimental under development": for example `__experimental__json_pointer_to_root` ?

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
kamatama41: From the perspective of current internal implementation of `JsonParser`, creating it multiple times would take a cost because it has `JsonFactory` which is a heavy-weight object.
ref: https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance#basics-things-you-should-do-anyway
That's why I made it reusable in the class. So creating JsonParser per instance sounds good to me 😃 
(Ideally we should make JsonFactory reusable in JsonParser)

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
kamatama41: That makes sense, I will do that.

kamatama41:
Thank you for reviewing! Let me merge.

PR #1101, opened 2019-02-06
[Experimental] Add an option to select a specific value in a record as an Embulk's column for JSON Parser plugin.

kamatama41:
For #1102 
## Usage
Sample JSON
```json
{"id": 1, "name": "Alice", "active": true, "created_at": "2019-01-12 01:23:45"}
```
Config YAML
```yaml
in:
  type: ...
  parser:
    type: json
    __experimental__columns:
      - {name: id, type: long}
      - {name: name, type: string}   
      - {name: active, type: boolean}   
      - {name: created_at, type: timestamp, format: "%Y-%m-%d %H:%M:%S"}   
```
Result 
```
+---------+-------------+----------------+-------------------------+
| id:long | name:string | active:boolean |    created_at:timestamp |
+---------+-------------+----------------+-------------------------+
|       1 |       Alice |           true | 2019-01-12 01:23:45 UTC |
+---------+-------------+----------------+-------------------------+
```

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/util/Timestamps.java
+++ b/embulk-core/src/main/java/org/embulk/spi/util/Timestamps.java
@@ -1,6 +1,7 @@
 package org.embulk.spi.util;
 
 import com.google.common.base.Optional;
+import java.util.HashMap;
 import java.util.Map;
 import org.embulk.config.Task;
 import org.embulk.spi.Column;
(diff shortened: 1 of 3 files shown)


embulk-core/src/main/java/org/embulk/spi/util/Timestamps.java
```diff
@@ -30,6 +31,20 @@ private Timestamps() {}
         return parsers;
     }
 
+    public static Map<Column, TimestampParser> newTimestampColumnParsersAsMap(
```
dmikurube: Sorry that it's really an internal and implicit reason, but actually, we basically don't want to expand existing "library" methods and classes anymore, especially under `spi.util`. (We'll eventually be extracting such utility/library classes as a separate artifact, not in embulk-core.) Once we have it under `spi.util`, plugin developers start using this, and we won't be able to remove it for a long time.

Then, if it can be a private method on `JsonParserPlugin` side, let's have it there.

(And, possibly nice to have that as a TODO comment.)

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -128,6 +152,78 @@ public void run(TaskSource taskSource, Schema schema, FileInput input, PageOutpu
         }
     }
 
+    private static boolean isUseCustomSchema(PluginTask task) {
```
dmikurube: nitpick: sounding a bit unnatural. How about `isSchemaCustomized` or `isUsingCustomSchema`?

embulk-core/src/main/java/org/embulk/spi/util/Timestamps.java
kamatama41: OK, moved it to the inside of the plugin in f45faac

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
kamatama41: Chose the second one (ref: 57938d5)

Issue #1102, opened 2019-02-06
[Feedbacks Wanted] JSON Parser plugin's enhancement


kamatama41:
## Overview

Current built-in JSON parser plugin has only one feature that it can parse multiple JSON documents to Embulk records that have just one Embulk JSON column. We'd like to enhance it with some new features.

## New features (draft)
- Selectable a specific value in JSON document as an Embulk column
```json
{"id": 1, "name": "Alice"}
{"id": 2, "name": "Bob"}
```
```yaml
in:
  type: ...
  parser:
    type: json
    columns:
      - {name: id, type: long}
      - {name: name, type: string}
```

```
+---------+-------------+
| id:long | name:string |
+---------+-------------+
|       1 |       Alice |
|       2 |         Bob |
+---------+-------------+
```

- Support pointing specific value via [JSON Pointer](https://tools.ietf.org/html/rfc6901) expression
```json
{ "result": {"id": 1, "name": "Alice"} }
```
For example, we can specify the nested JSON object as an Embulk record via a pointer `/result`
```
+-------------------------+
|             record:json |
+-------------------------+
| {"id":1,"name":"Alice"} |
+-------------------------+
```

- JSON array can be flattened to each Embulk record

```json
[
  {"id": 1, "name": "Alice"},
  {"id": 2, "name": "Bob"}
]
```

```
+-------------------------+
|             record:json |
+-------------------------+
| {"id":1,"name":"Alice"} |
|   {"id":2,"name":"Bob"} |
+-------------------------+
```

cc: @dmikurube , @sakama, @muga

kamatama41:
Dear users of JSON parser plugin,
The features are experimentally released since Embulk 0.9.5. We would like your feedback to improve it, thank you!

hiroyuki-sato:
@kamatama41 

I found the typo. Embulk 0.9.5 -> Embulk 0.9.15

I would like to know `bob`'s data in the first case.

It that the following?

```json
{"id": 1, "name": "Alice"},
{"id": 2, "name": "Bob"}
```

kamatama41:
@hiroyuki-sato 
Thank you for pointing out, you are right. I fixed them.

hiroyuki-sato:
@kamatama41 I recommend you post this feedback request to embulk-announce.

dmikurube:
The enhancement has been landed for a long time. Let's close this feedback request.

PR #1103, opened 2019-02-06
[Experimental] Add an option to select any value in JSON record by JSON Pointer expression.

kamatama41:
For #1102 and a follow-up of #1101 
## Usage
Sample JSON
```json
{"id": 1, "name": "Alice", "tags": ["a", "b", "c"]}
```
Config YAML
```yaml
in:
  type: ...
  parser:
    type: json
    __experimental__columns:
      - {name: id, type: long}
      - {name: name, type: string}   
      - {name: first_tag, type: string, json_pointer: "/tags/0"}   
```
Result 
```
+---------+-------------+------------------+
| id:long | name:string | first_tag:string |
+---------+-------------+------------------+
|       1 |       Alice |                a |
+---------+-------------+------------------+
```

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
@@ -86,7 +86,7 @@ public class JsonParserPlugin implements ParserPlugin {
     }
 
     public interface OptionalColumnConfig extends Task, TimestampParser.TimestampColumnOption {
-        @Config("json_pointer")
+        @Config("relative_json_pointer_from_root")
         @ConfigDefault("null")
         Optional<String> getJsonPointer();
     }
--- a/embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
@@ -8,6 +8,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
@@ -84,7 +85,11 @@ public class JsonParserPlugin implements ParserPlugin {
         Optional<SchemaConfig> getSchemaConfig();
     }
 
-    interface TimestampColumnOption extends Task, TimestampParser.TimestampColumnOption { }
+    public interface OptionalColumnConfig extends Task, TimestampParser.TimestampColumnOption {
+        @Config("json_pointer")
+        @ConfigDefault("null")
+        Optional<String> getJsonPointer();
+    }
 
     @Override
     public void transaction(ConfigSource configSource, Control control) {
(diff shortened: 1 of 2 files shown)


embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -83,6 +85,12 @@ public String getString() {
         Optional<SchemaConfig> getSchemaConfig();
     }
 
+    public interface OptionalColumnConfig extends Task {
+        @Config("json_pointer")
```
dmikurube: Not very confident, but we may want to call it a kind of `relative`, IIUC?

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -228,6 +255,18 @@ private PageBuilder newPageBuilder(Schema schema, PageOutput output) {
...
     }
 
+    private Value parseColumnValueWithOffsetInJsonPointer(String valueAsJsonString, String jsonPointer) {
+        try {
+            return jsonParser.parseWithOffsetInJsonPointer(valueAsJsonString, jsonPointer);
+        } catch (JsonParseException e) {
+            /*
+             * When JsonParseException is thrown, it would be an error that the given JSON pointer doesn't match with the JSON object.
+             * We would return NULL when the pointer doesn't match, not throw Exception.
+             */
```
dmikurube: We may want to discuss about this -- what we should do in case of parsing errors. For example in case `stop_on_invalid_record: true` ?

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
kamatama41: Yes, it can be relative if `__experimental__json_pointer_to_root` is specified.. so something like `relative_json_pointer_from_root` might be better, which is long, though.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
kamatama41: If we discuss the specification of this, we would need to consider about the case of "when the specified key (or JSON pointer) doesn't exist in a record with custom schema". Current version always returns NULL in both case inexistent key and pointer.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
kamatama41: Renamed to `relative_json_pointer_from_root`. How about this?
ref: 7672bae

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
kamatama41: As a result of F2F communication, in the first version, we concluded that the JSON parser returns NULL if the given pointer doesn't match within a JSON record, which means it's not needed to change current version. cc: @dmikurube

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
dmikurube: Sounds reasonable. At the same time, it could be worth looking into https://tools.ietf.org/id/draft-handrews-relative-json-pointer-00.html while it's still draft.

embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
dmikurube: Sounds. good. Just nice to leave more comments here -- from the viewpoint of application-level, not only from the Embulk-core's code-level.

kamatama41:
Thank you for reviewing~!

PR #1104, opened 2019-02-08
[Experimental] Add an option to flatten JSON array to each Embulk record for JSON Parser plugin

kamatama41:
For #1102 and a follow-up of #1103 

## Usage
Sample JSON
```json
[
  {"id": 1, "name": "Alice"},
  {"id": 2, "name": "Bob"},
]
```
Config YAML
```yaml
in:
  type: ...
  parser:
    type: json
    __experimental__flatten_json_array: true
```
Result 
```
+-------------------------+
|             record:json |
+-------------------------+
| {"id":1,"name":"Alice"} |
|   {"id":2,"name":"Bob"} |
+-------------------------+
```

### DIFF
--- a/embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
+++ b/embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
@@ -7,6 +7,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


embulk-standards/src/main/java/org/embulk/standards/JsonParserPlugin.java
```diff
@@ -129,18 +134,23 @@ public void run(TaskSource taskSource, Schema schema, FileInput input, PageOutpu
                     Value value;
                     while ((value = stream.next()) != null) {
                         try {
-                            if (!value.isMapValue()) {
-                                throw new JsonRecordValidateException(
-                                        String.format("A Json record must not represent map value but it's %s", value.getValueType().name()));
+                            final Iterable<Value> recordValues;
+                            if (task.getFlattenJsonArray()) {
+                                if (!value.isArrayValue()) {
```
dmikurube: I a little bit thought about other types of "flattening", but it must be too much at the moment. It may be a bit worth just thinking an extensible configuration format, though.

(By other types of flattening, I mean, for example, doubly-nested arrays, or JSON objects with considering a key as a column.)

kamatama41: As a result of f2f communication with @dmikurube, we will go with current version so far and consider such complex feature in the time when it's actually required.

kamatama41:
Thank you for reviewing! Let me merge.

PR #1106, opened 2019-02-15
Do not warn Maven libraries uninitialized when using self-contained JARs

dmikurube:
One follow-up to #1085. #1085 unintentionally showed a warning `Dependencies for Maven are uninitialized. Expected to use classes loaded by the parent ClassLoader.` when using the self-contained JARs == in case of the all-in-one CLI package.

This PR fixes that.


dmikurube:
Thanks!

Issue #1110, opened 2019-02-25
Unsupport nested-JAR style JAR plugins


dmikurube:
Currently, Embulk supports JAR-based Embulk plugins to load their dependencies in three ways.

1. Expand all dependencies flat into the plugin JAR
2. Embed all dependencies nested in the plugin JAR
3. Specify dependencies as Maven dependencies of the plugin artifact

1 is the most primitive way, but basically not recommended because the package easily gets bigger, and their license issues are not very clear.

2 was implemented around #792 to relax the problems of 1. Dependency JARs are embedded in the main plugin JAR as-is, and they're loaded by `org.embulk.plugin.PluginClassLoader`. But, the package is still big, and `PluginClassLoader` has gotten much complicated by this change.

3 is implemented the most recently in #1012 and #1015. Dependency libraries are not included in the main plugin JAR, but loaded through `<dependencies>` declarations in the plugin's `pom.xml`. To make the loaded versions clear, Embulk finds only the "direct" dependencies, without transitioned dependencies.

----

Maven-based JAR plugins are not very popular yet, and it should be safe to remove some features. Because we already have 3, and 2 is too complicated, let's remove the feature of 2.

kamatama41:
Done.

kamatama41:
Ah, It should be removed after new version is released.

dmikurube:
v0.9.17 is out. Closing!

PR #1111, opened 2019-02-28
Unsupport nested-JAR style JAR plugins.

kamatama41:
For #1110 
All codes that relate to`embeddedJarPathsInNestedJar` in `PluginClassLoader` are removed.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
+++ b/embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
@@ -3,7 +3,6 @@ package org.embulk.plugin;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
@@ -203,32 +202,6 @@ public class PluginClassLoader extends URLClassLoader {
             URL childUrl = findResource(name);
             if (childUrl != null) {
                 return childUrl;
-            }
-        }
-
-        return null;
-    }
-
-    @Override
-    public InputStream getResourceAsStream(final String resourceName) {
-        final boolean childFirst = isParentFirstPath(resourceName);
-
-        if (childFirst) {
-            final InputStream childInputStream = super.getResourceAsStream(resourceName);
-            if (childInputStream != null) {
-                return childInputStream;
-            }
-        }
-
-        final InputStream parentInputStream = getParent().getResourceAsStream(resourceName);
-        if (parentInputStream != null) {
-            return parentInputStream;
-        }
-
-        if (!childFirst) {
-            final InputStream childInputStream = super.getResourceAsStream(resourceName);
-            if (childInputStream != null) {
-                return childInputStream;
             }
         }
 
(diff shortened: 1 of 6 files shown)


embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
```diff
@@ -518,400 +327,8 @@ public InputStream getInputStream() throws IOException {
...
-            }
-        }
-        if (lastException != null) {
-            throw new ClassNotFoundException(className, lastException);
-        } else {
-            throw new ClassNotFoundException(className);
-        }
-    }
-
     private InputStream getResourceAsStreamFromChild(final String resourceName) {
```
dmikurube: We may be able to remove this method to change the caller side to call `super.getResourceAsStream(resourceName)` directly.

Also, the caller side `getResourceAsStream` may be removed. `URLClassLoader#getResourceAsStream` is actually calling `getResource` to find the resource as below.
http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/net/URLClassLoader.java#l232

`getResource` is overridden in `PluginClassLoader` in the same way where the overridden `getResourceAsStream` is doing.

embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
```diff
@@ -301,7 +194,7 @@ public void addUrl(URL url) {
      */
     @Override
     public URL findResource(final String resourceName) {
```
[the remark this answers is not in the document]
kamatama41: Yes, I overlooked both `if` and `else` is the same logic 😅, fixed in e55755c

embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
```diff
@@ -338,32 +218,18 @@ public URL run() {
      */
     @Override
     public Enumeration<URL> findResources(final String resourceName) throws IOException {
```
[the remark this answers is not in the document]
kamatama41: Same as above, fixed in e55755c

embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
kamatama41: I just removed the method and call `super.getResourceAsStream(resourceName)` directly in ca688e9

dmikurube: Thanks!

Plus, as a straight inference looking in `URLClassLoader.java`, I guessed we can remove the overridden `PluginClassLoader#getResourceAsStream` as well.

What the overridden `PluginClassLoader#getResourceAsStream` does is almost the same as the overridden `PluginClassLoader#getResource`, and then just loading it as `InputStream` in addition.

The same operation could be done just by the non-overridden `URLClassLoader#getResourceAsStream` calling `getResource`, where `getResource` is actually overridden as `PluginClassLoader#getResource`.

kamatama41: I see. Removed it in 90ab1f6

kamatama41:
Thanks. I found another unused variable and removed it in df33bdd.

PR #1115, opened 2019-03-04
Remove general help of `-R` options to have only `-R--dev`

dmikurube:
The command line option `-R` no longer has general meanings. Replacing it to a description of only `-R--dev`.

@sakama Can you have a look?


dmikurube:
Thanks!

PR #1117, opened 2019-03-04
Add more tests for TestJarPluginLoader

kamatama41:
A part of #1110.
Before #1111, I added a test for nested style Jar plugin. In addition, I wanted to test `PluginClassLoader` in the `JarPluginLoader`. So I did the following changes:

- Create new project ~`embulk-test-maven-plugin`~ `test-helpers` and move `ExampleJarSpiV0` and `ExampleDependencyJar` into the project. This is intended to make the classes out of the scope of the default class loader.
  - `main` project has class to be loaded as plugin's main class
  - The `main` project depends on the `deps` project.
  - The project can be built as both the flat Jar style and the nested Jar style.
- ~Added 3 Jar files into `embulk-core/src/test/resources/jars`, which are pre-built on the `embulk-test-maven-plugin`.~
- Deploy 3 Jar files into ~`embulk-core/build/embulk-test-maven-plugin`~ `test-helpers/build/embulk-test-maven-plugin`, which are built before running `embulk-core:test`

What do you think about @dmikurube


kamatama41:
@dmikurube 
I think we can do it. The reason why I pre-built and put JARs into the Git repo is to simplify build and test process. If we build the JARs on-demand, we need to add the process before running tests. I didn't want to do it. But, as your opinion makes sense to me, I will try it 😃

dmikurube:
Thanks for your consideration!

kamatama41:
@dmikurube 
I've changed it to build and deploy JARs on `@BeforeClass` of `TestJarPluginLoader` in e4b6cb2
(It might be better to remove the binaries from Git history by rebasing, though)
How about this?

dmikurube:
@kamatama41 Thanks!

Hmm, I personally like the latter way, to build test JAR files from an existing `build.gradle` tree. It's true that the approach is a bit hard to understand, but at the same time
* having a totally-independent `build.gradle` (`embulk-test-maven-plugin/build.gradle` has no relationship with the top-level `build.gradle` IIUC) is usually unexpected by other developers, and
* running Gradle in a nested manner from test code is not very intuitive, as well.

As we'll have some more hacks around ClassLoaders, we'll need more tests like this. Then, what do you think:
1. Have a top-level subproject named like `test-helpers` on the top directory, not only for this Maven plugin tests.
2. Build this `embulk-test-maven-plugin` now, and more JARs not to be loaded automatically in the future, in the subproject
3. Locate (deploy) the built JAR files into build directories required

?

dmikurube:
Ah, eventually when merging, let's do the "Squash" merge. That will remove the binaries from the history.

kamatama41:
Thank you for your suggestion. It makes sense to me. I will try it 😃

kamatama41:
Thank you for reviewing. Let me merge.

dmikurube:
Thanks! Then released Embulk v0.9.16. https://github.com/embulk/embulk/releases/tag/v0.9.16

PR #1120, opened 2019-03-06
Fix option names for JSON parser plugin's new features.

kamatama41:
For #1102 
After an internal discussion with @dmikurube and @sakama, we decided option names as following:
-  `__experimental__json_pointer_to_root` -> `root`
- `__experimental__flatten_json_array` -> `flatten_json_array`
- `__experimental__columns` -> `columns`
- `relative_json_pointer_from_root` -> `pointer`


kamatama41:
Thanks!

PR #1121, opened 2019-03-06
Add descriptions for new options for JSON parser plugin.

kamatama41:
For #1102 
As we fixed option names in #1120, it's time to add them into the official doc.

### DIFF
--- a/embulk-docs/src/built-in.rst
+++ b/embulk-docs/src/built-in.rst
@@ -353,13 +353,25 @@ The ``json`` parser plugin parses a JSON file that contains a sequence of JSON o
 Options
 ~~~~~~~~
 
-+----------------------------+----------+----------------------------------------------------------------------------------------------------------------+------------------------------+
-| name                       | type     | description                                                                                                    |          required?           |
-+============================+==========+================================================================================================================+==============================+
-| stop\_on\_invalid\_record  | boolean  | Stop bulk load transaction if a file includes invalid record (such as invalid json)                            | ``false`` by default         |
-+----------------------------+----------+----------------------------------------------------------------------------------------------------------------+------------------------------+
-| invalid\_string\_escapes   | enum     | Escape strategy of invalid json string such as using invalid ``\`` like ``\a``. (PASSTHROUGH, SKIP, UNESCAPE)  | ``PASSTHROUGH`` by default   |
-+----------------------------+----------+----------------------------------------------------------------------------------------------------------------+------------------------------+
++----------------------------+----------+-----------------------------------------------------------------------------------------------------------------+----------------------------------------+
+| name                       | type     | description                                                                                                     |          required?                     |
++============================+==========+=================================================================================================================+========================================+
+| stop\_on\_invalid\_record  | boolean  | Stop bulk load transaction if a file includes invalid record (such as invalid json)                             | ``false`` by default                   |
++----------------------------+----------+-----------------------------------------------------------------------------------------------------------------+----------------------------------------+
+| invalid\_string\_escapes   | enum     | Escape strategy of invalid json string such as using invalid ``\`` like ``\a``. (PASSTHROUGH, SKIP, UNESCAPE)   | ``PASSTHROUGH`` by default             |
++----------------------------+----------+-----------------------------------------------------------------------------------------------------------------+----------------------------------------+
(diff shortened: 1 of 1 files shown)


embulk-docs/src/built-in.rst
```diff
@@ -353,13 +353,25 @@ The ``json`` parser plugin parses a JSON file that contains a sequence of JSON o
...
++----------------------------+----------+-----------------------------------------------------------------------------------------------------------------+----------------------------------------+
+| flatten_json_array         | boolean  | Set true when regard elements in a JSON array as multiple Embulk records.                                       | ``false`` by default                   |
++----------------------------+----------+-----------------------------------------------------------------------------------------------------------------+----------------------------------------+
+| columns                    | hash     | Columns (see below)                                                                                             | optional                               |
++----------------------------+----------+-----------------------------------------------------------------------------------------------------------------+----------------------------------------+
+| default\_timezone          | string   | Time zone of timestamp columns. This can be overwritten for each column using ``column``                        | ``UTC`` by default                     |
```
dmikurube: Yeah, `default_timezone`, `default_timestamp_format`, and `default_date` can be configured actually, but in a rough direction, we're going to deprecate them.

Sorry that I didn't let you know, but let's keep them undocumented.

embulk-docs/src/built-in.rst
```diff
@@ -376,6 +388,23 @@ if you set invalid\_string\_escapes and appear invalid JSON string (such as ``\a
 
 (\*1): Throwing an exception.
 
+The ``columns`` option can extract any value as an Embulk's column.
```
dmikurube: How about :
```
The ``columns`` option declares the list of columns, and the way how to extract JSON values into Embulk columns.
```

embulk-docs/src/built-in.rst
```diff
@@ -376,6 +388,23 @@ if you set invalid\_string\_escapes and appear invalid JSON string (such as ``\a
...
+
++----------+-----------------------------------------------------------------------------------------------------------+
+| name     | description                                                                                               |
++==========+===========================================================================================================+
+| name     | Name of the column. Select a key within a JSON record or any name can be allowed with ``pointer`` option. |
```
dmikurube:
```
Name of the column. The JSON value with this name is extracted if `pointer` is not specified.
```

embulk-docs/src/built-in.rst
```diff
@@ -376,6 +388,23 @@ if you set invalid\_string\_escapes and appear invalid JSON string (such as ``\a
...
++----------+-----------------------------------------------------------------------------------------------------------+
+| type     | Type of the column (same as CSV parser's one)                                                             |
++----------+-----------------------------------------------------------------------------------------------------------+
+| pointer  | Set as a JSON Pointer expression if you want to specify any value in a JSON record as the column's value  |
++----------+-----------------------------------------------------------------------------------------------------------+
+| timezone | Set if you want to specify time zone for the timestamp column                                             |
```
dmikurube: For a similar reason with above, let's keep only `format` here without `timezone` and `date`.

embulk-docs/src/built-in.rst
```diff
@@ -376,6 +388,23 @@ if you set invalid\_string\_escapes and appear invalid JSON string (such as ``\a
...
+The ``columns`` option can extract any value as an Embulk's column.
+
++----------+-----------------------------------------------------------------------------------------------------------+
+| name     | description                                                                                               |
++==========+===========================================================================================================+
+| name     | Name of the column. Select a key within a JSON record or any name can be allowed with ``pointer`` option. |
++----------+-----------------------------------------------------------------------------------------------------------+
+| type     | Type of the column (same as CSV parser's one)                                                             |
++----------+-----------------------------------------------------------------------------------------------------------+
+| pointer  | Set as a JSON Pointer expression if you want to specify any value in a JSON record as the column's value  |
```
dmikurube: How about :
```
Pointer to the child node to be extracted as the column, expressed as a JSON Pointer (optional).
```

dmikurube: Ah, it may not be `child`, but should be `descendant`.

kamatama41:
@dmikurube 
Thank you for your suggestion. I have updated following them in ef3c2fa

PR #1124, opened 2019-03-11
Use "element_at" instead of "pointer" in the built-in JSON parser plugin

dmikurube:
The configuration name `pointer` was the last remaining consideration when we discussed the names. What do you think: `element_at` instead?

```
    columns:
      - {name: id, type: long}
      - {name: name, type: string}
      - {name: first_tag, type: string, element_at: "/tags/0"}
```

The noun "element" came from https://json.org/. It is enough explicit that it extracts a JSON element from that position. We can naturally extend it later like `element_at: "jsonpath:tags[0]"`.

What do you think, @kamatama41 @sakama ?

### DIFF
--- a/embulk-docs/src/built-in.rst
+++ b/embulk-docs/src/built-in.rst
@@ -384,17 +384,17 @@ if you set invalid\_string\_escapes and appear invalid JSON string (such as ``\a
 
 The ``columns`` option declares the list of columns, and the way how to extract JSON values into Embulk columns.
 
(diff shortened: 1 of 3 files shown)


dmikurube:
Yeah, I believe those should be fine because they're internal names, and they're exactly "JSON Pointers" there.

I chose `element_at` only in the external (user) configuration name. In case we consider "`jsonpath:`" extension, I think we'll have other internal methods.

dmikurube:
Thanks!

PR #1123, opened 2019-03-08
Load Apache Commons CLI library in a separate ClassLoader

dmikurube:
Continued from #1085. It isolates Apache Common CLI library to be hidden from plugins.

It's a kind of a big change. The changes are broken down into several commits to be more readable.

@sakama Can you have a look?

----

The all-in-one JAR will be structured like below.
```
  lib/
  lib/embulk-deps-maven-0.9.16-SNAPSHOT.jar
+ lib/embulk-deps-cli-0.9.16-SNAPSHOT.jar
  lib/maven-aether-provider-3.3.9.jar
  lib/maven-model-builder-3.3.9.jar
  lib/maven-artifact-3.3.9.jar
  lib/aether-impl-1.1.0.jar
  lib/aether-spi-1.1.0.jar
  lib/aether-util-1.1.0.jar
  lib/aether-api-1.1.0.jar
+ lib/commons-cli-1.3.1.jar
  lib/maven-model-3.3.9.jar
  lib/maven-repository-metadata-3.3.9.jar
  lib/maven-builder-support-3.3.9.jar
  lib/plexus-utils-3.0.22.jar
  lib/plexus-component-annotations-1.6.jar
  lib/plexus-interpolation-1.21.jar
...
- org/embulk/third_party/org/apache/commons/cli/
- org/embulk/third_party/org/apache/commons/cli/AlreadySelectedException.class
- org/embulk/third_party/org/apache/commons/cli/AmbiguousOptionException.class
- org/embulk/third_party/org/apache/commons/cli/BasicParser.class
...
```

And, `MANIFEST.MF` will be like below.
```
  Manifest-Version: 1.0
+ Embulk-Resource-Class-Path-Cli: /lib/embulk-deps-cli-0.9.16-SNAPSHOT.j
+  ar /lib/commons-cli-1.3.1.jar
  Embulk-Resource-Class-Path-Maven: /lib/embulk-deps-maven-0.9.16-SNAPSH
   OT.jar /lib/maven-aether-provider-3.3.9.jar /lib/maven-model-builder-
   3.3.9.jar /lib/maven-artifact-3.3.9.jar /lib/plexus-utils-3.0.22.jar 
   /lib/maven-model-3.3.9.jar /lib/maven-builder-support-3.3.9.jar /lib/
   plexus-component-annotations-1.6.jar /lib/plexus-interpolation-1.21.j
   ar /lib/aether-impl-1.1.0.jar /lib/aether-spi-1.1.0.jar /lib/aether-a
[...]

PR #1127, opened 2019-03-11
Load Apache Velocity in the sub ClassLoader

dmikurube:
Follow-up to #1123 to isolate Apache Velocity in addition. Velocity will be loaded by the same ClassLoader instance with Apache Commons CLI.

What do you think, @sakama ?


dmikurube:
Thanks!

PR #1131, opened 2019-03-25
Test with Logback so that logs can be checked

dmikurube:
SLF4J logs have not been displayed during tests. It adds Logback to `testCompile` to make them displayed.

@sakama Can you have a look?


dmikurube:
Thanks!

> Latest stable version seems 1.2.3?

Looks so. But going to use the same version with the main code. Planning to upgrade both main and test in the next chance.

PR #1132, opened 2019-03-25
Self-update from https://dl.embulk.org

dmikurube:
Current Embulk depends on Bintray and its behavior in `selfupdate`, but we'd like to be independent. This PR makes `selfupdate` to depend on only `dl.embulk.org`.

@sakama Can you have a look?


dmikurube:
Thanks, too!

PR #1133, opened 2019-03-27
Provide system config options to disable selected standard plugins

dmikurube:
For example, consider we run :

```
embulk run -X standards.output.stdout.disabled=true test.yml
```

If `test.yml` contains `out: type: stdout`, it fails with an error :

```
org.embulk.config.ConfigException: OutputPlugin 'stdout' is not found.
```

@sakama Can you have a look?


dmikurube:
Thanks!

PR #1134, opened 2019-03-27
Cleanup EmbulkEmbed and EmbulkService

dmikurube:
Retry of #1038 in a bit different manner. We don't need to upgrade Guice after #1078. We can do this kind of refactoring more safely.

* Reorder methods and fields.
* Set system config through JSON-formatted `String` in `Bootstrap`. No use of "System config loader" outside.
* Move `EmbulkService.standardModuleList` to `EmbulkEmbed`
* Reduce usage of Guava
* Use Java 8 `java.util.function.Function` instead of Guava `Function`
* Use statically-initialized `Logger`

@sakama Can you have a look?


dmikurube:
Thanks!

PR #1135, opened 2019-03-28
Process a new system config "gem_path", to work the same way as GEM_PATH

dmikurube:
Allowing Embulk to configure `GEM_PATH`, in addition to `GEM_HOME`, through system configs.
https://stackoverflow.com/questions/11277227/whats-the-difference-between-gem-home-and-gem-path

By configuring `GEM_PATH`, users can look for Embulk plugins from multiple Gem directories.

@sakama Can you have a look when you have time?


dmikurube:
Thanks!

PR #1140, opened 2019-04-10
Abandon FindBugs

dmikurube:
FindBugs is no longer maintained.
https://mailman.cs.umd.edu/pipermail/findbugs-discuss/2016-November/004321.html

And we have had unnecessary dependency on `com.google.code.findbugs:annotations:3.0.0` for FindBugs.

We're not effectively using FindBugs. It's time to abandon FindBugs. We may enable SpotBugs instead in the future.


dmikurube:
@sakama Can you have a look?

dmikurube:
Thanks!

PR #1141, opened 2019-04-11
Cleanup Buffer, BufferAllocator, and PooledBufferAllocator, and deprecate ByteSize

dmikurube:
Cleaning up `Buffer`, `BufferAllocator`, and `PooledBufferAllocator`. Along with it,

* Reorder methods and fields.
* `PooledBufferAllocator` to be instantiated without Guice and system config.
* Deprecate `ByteSize`, which has been used only in `PooledBufferAllocator`. No plugins using this class are found.
* Use `BigDecimal` to parse byte size strings, instead of `double`.
* Reduce usage of Guava.

@sakama Can you have a look when you have time? (Not in a hurry)


dmikurube:
Thanks!

PR #1144, opened 2019-04-15
Make PooledBufferAllocator package-private

dmikurube:
Follow-up to #1141. `PooledBufferAllocator` does not need to be public.

@sakama Can you have a look?


dmikurube:
Thanks!

PR #1145, opened 2019-04-15
Document planned deprecations

dmikurube:
Just an internal document update. @sakama Can you have a look when you have time?


dmikurube:
Thanks!

PR #1146, opened 2019-04-15
Remove unnecessary uses of SystemConfig

dmikurube:
Another cleanup. Removing unnecessary use of system configs. @sakama Can you have a look when you have time? (No hurry.)


dmikurube:
Thanks!

PR #1157, opened 2019-07-08
Remove unnecessary injections

dmikurube:
Simply removing unnecessary injections. Can you have a look, @sakama ?


dmikurube:
Thanks!

PR #1158, opened 2019-07-08
Fix EmbulkEmbed.Bootstrap.setSystemConfig, which was not working

dmikurube:
Was a bug. `EmbulkEmbed.Bootstrap.setSystemConfig(ConfigSource)` was totally not working. (But it was not used in `embulk-core`. `embulk-core` uses `setSystemConfigFromJson` instead.)

Fixing the bug, and letting `setSystemConfigFromJson` to call `setSystemConfig` internally.

@sakama Can you have a look?


dmikurube:
Thanks!

PR #1160, opened 2019-07-08
Set system config also in java.util.Properties

dmikurube:
Embulk's system config has been represented with `ConfigSource`, the same with Embulk's user config based on YAML/JSON/Jackson. `ConfigSource` requires a lot, including library dependencies, and their initialization.

On the other hand, system config is sometimes wanted in early steps of Embulk's bootstrap. Then, System config should not depend on many dependencies.

Representing system config with standard `java.util.Properties` would help. `Properties` is actually less expressive than Jackson, but in almost all the use-cases, we have put just string:string mappings in system config, except for a couple of internal use-cases.

This commit is the first step to replace system config to use `Properties`. It additionally stores system config in `Properties`, accepts system config set from `Properties`, and uses `Properties` to set system config from CLI.

It still uses the `ConfigSource` version of system config inside embulk-core. They will be replaced with the `Properties` version in later Embulk versions.


dmikurube:
The following system configs used standard expect arrays. We may need special handling for them.

* `guess_plugins`
* `jruby_classpath`
* `jruby_load_path`

dmikurube:
@shroman Can you have a look when you have time?

#1201, #1204 and this would be the key parts of v0.9.22. (#1205 is not included.)

embulk-core/src/test/java/org/embulk/jruby/TestJRubyInitializer.java
```diff
@@ -3,31 +3,21 @@
...
                 null,
                 LoggerFactory.getLogger(TestJRubyInitializer.class),
                 "/gem/home",
                 null,
                 true,
-                loadPath,
-                classpath,
-                options,
+                "/load/path",
+                "/classpath",
```
shroman: these paths are with separator, I would add an additional test (if you don't consider it as a separate refactoring etc.)

dmikurube: Thanks for the good catch! Sounds reasonable. In the same test case `testArguments`, added another path for `load_path` and `classpath` in ee06c77b752451afb7ce2028bf85fd109a046114 to test separation by the path separator.

PR #1162, opened 2019-07-10
Remove unnecessary injection in TempFileAllocator

dmikurube:
After #1157, another unnecessary injection found.

@sakama Can you have a look when you get time?


dmikurube:
Thanks!

PR #1163, opened 2019-07-10
Use EmbulkSystemProperties in embulk-core and its Guice Modules

dmikurube:
`EmbulkSystemProperties` was introduced in #1160 (v0.9.22) to replace the system config represented by `ConfigSource`.

After it is added in v0.9.22, starting to use it in embulk-core v0.10.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
+++ b/embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
@@ -39,7 +39,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class EmbulkEmbed {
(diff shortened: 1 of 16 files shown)


embulk-core/src/main/java/org/embulk/EmbulkEmbed.java
```diff
@@ -56,12 +56,13 @@ public Bootstrap() {
             this.started = false;
         }
 
-        @Deprecated  // To be removed. Plugins should not call this by themselves.
+        @Deprecated  // To be removed. Users and plugins should not call this by themselves.
```
shroman: Good time to remove deprecated code ;) Maybe for a separate pt though.

dmikurube:
Thanks!

PR #1165, opened 2019-07-19
Load Maven dependencies only of "compile" and "runtime" scopes

dmikurube:
By #1015, Maven-based plugins are loaded only with its "direct" dependency libraries, without its "transitive" dependencies ( == dependencies of dependencies).

This PR restricts the load only to `compile` and `runtime` scope. It will work with Embulk plugins built with the updated Gradle plugin: https://github.com/embulk/gradle-embulk-plugins/pull/24

----

@sakama Can you have a look? (No hurry.)

I plan to release v0.9.18 after this PR. The date may be a bit later, though.


dmikurube:
Thanks!

PR #1175, opened 2019-08-16
Support new rules for bundled Rename Filter plugin

sakama:
This PR is a proposal of new options that introduces 2 new rename rules for bundled Rename Filter plugin.

* `upper_to_lower_underscore` as known as **"snake case"**
  e.g. "UserData" to "user_data"
* `upper_to_lower_hyphen` as known as **"kebab case"**
  e.g. "UserData" to "user-data"

I think both of them are popular formats.
With the current rename rule, we have to set all 1 to 1 convert rule manually. This is the only way but the official document says "not recommended"
https://www.embulk.org/docs/built-in.html#columns-not-recommended

```yaml
filters:
- type: rename
  columns:
    Name: name
    UsersData: user_data
```

### Note

There might be some variations for both formats when the string starts with `_` or `-`.
I used `com.google.common.base.CaseFormat` for both rules and this library adds `_` or `-` even if the string starts with such characters.

* To lower underscore case

| Original string | CaseFormat (guava) | Other valiation (doesn't add new "_") |
----|---- |---- 
| _UserData       | __user_data        | _user_data      |
| __UserData      | ___user_data        | __user_data      |

* To lower hyphen case

| Original string | CaseFormat (guava) | Other valiation (doesn't add new "-") |
----|---- |---- 
| -UserData       | --user-data        | -user-data      |
| --UserData      | ---user_data        | --user_data      |

### DIFF
--- a/embulk-docs/src/built-in.rst
+++ b/embulk-docs/src/built-in.rst
@@ -653,23 +653,27 @@ Renaming rules
 
 The ``rules`` is an array of rules as below applied top-down for all the columns.
 
-+-------------------------+----------------------------------------------------------------------------------------+
-| rule                    | description                                                                            |
-+=========================+========================================================================================+
-| character\_types        | Restrict characters by types. Replace restricted characteres.                          |
-+-------------------------+----------------------------------------------------------------------------------------+
-| first\_character\_types | Restrict the first character by types. Prefix or replace first restricted characters.  |
-+-------------------------+----------------------------------------------------------------------------------------+
-| lower\_to\_upper        | Convert lower-case alphabets to upper-case.                                            |
-+-------------------------+----------------------------------------------------------------------------------------+
-| regex\_replace          | Replace with a regular expressions.                                                    |
-+-------------------------+----------------------------------------------------------------------------------------+
-| truncate                | Truncate.                                                                              |
-+-------------------------+----------------------------------------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


dmikurube:
I understand that need, but it still has some difficulties there.

----

One is case variations (mostly about case **before** renaming). The code looks assuming `UpperCamel`, but source column names can be more varied: `UPPER_SNAKE`, `UPPER-HYPHEN`, `lower_underscore`, and else. If we support `UpperCamel` to `lower_underscore`, we have no reason to support `UPPER_SNAKE` to `UpperCamel`, and so many "source x destination" variations.

Implementing just a part of them now is fine. But, the configuration syntax at least needs to be extensible for such more variations. It is expectedly hard to maintain all the "source x destination" rename rules.

Once we have these two rules here now, it would be so hard to change them. We'll need to be cautious there.

One idea of possible config to be extensible may be like below?

```
- type: rename
  rules:
  - rule: case_format
    from: upper_camel
    to: lower_underscore
```

----

The other is Guava. We are actually trying to eliminate Guava usage in the Embulk core because it has dependency problems with plugins. Now, it is hard for us to upgrade Guava in Embulk core without any compatibility impact, and plugins cannot use newer Guava even by themselves. Therefore, I wanted to avoid adding new usage of Guava in Embulk.

----

Or, another easy way can be to implement an independent filter plugin, such as `rename_advanced`? Plugins out of the Embulk core can be weird, but more controllable with less thoughts about compatibility.

What do you think?

Issue #1151, opened 2019-06-12
Improve TempSpace


tvhung83:
### Problems
* Temp space dir contains [space in name](https://github.com/embulk/embulk/blob/v0.9.17/embulk-core/src/main/java/org/embulk/spi/ExecSession.java#L101), eg. `2019-05-15 11:35:48.617 UTC`
* There is a rare issue of temp file not found > add more logs when embulk [clean up temp files](https://github.com/embulk/embulk/blob/fd9c0c0234033e40808478544ffcf8832e4b2aa0/embulk-core/src/main/java/org/embulk/exec/BulkLoader.java#L378)

### Expected result
* Rename temp space to not include space
* Add logs during clean up (but not too verbose)

dmikurube:
To be fixed in v0.9.20 by #1190 and #1189... to be confirmed.

dmikurube:
It looks to be done in Embulk v0.9.20. We no longer observe the "temp file not found" errors. Let me close it!

Issue #1142, opened 2019-04-11
sporadic FileNotFoundException


kulmam92:
Issue Type: Bug Report
-----------------------

* Write the following environmental information.
  * OS version: windows 2016
  * Java version: JRE8
  * Embulk version: 0.9.17
  * Your Embulk configuration (YAML)
  * Plugin versions
* Write all what you did, e.g. your commands executed
* Write what you expected and observed actually

I have a workflow that execute 10 embulk session in parallel multiple times.
Below is the part of workflow.

```
+execute_channel:
  for_each>: {j: [0,1,2,3,4,5]}
  _do:
    +get_pipeline_config:
      py>: scripts.main.get_pipeline_config_channel
      require: 'scripts/main'  

    +execute_pipeline:
      for_each>:
        i: ${TOTAL_NUMBER}
      _parallel: true
      _do:
        +step1:
          _export:
            EMBULK_SQL_STRING: ${SQL_STRING[i]}
            EMBULK_BLOB_CONTAINER: ${BLOB_CONTAINER[i]}
            EMBULK_BLOB_FULPATH: ${BLOB_FULLPATH[i]}
            EMBULK_BLOB_FILENAME: ${BLOB_FILENAME[i]}
            EMBULK_BLOB_FORMAT: ${BLOB_FORMAT[i]}
            EMBULK_BLOB_PIPE_ID: ${PIPE_ID[i]}
          sh>: embulk '"-J-O"' run .\scripts\config.yml.liquid
```

One of the issue that I face is sporadic file not found error like below.

```
2019-04-11 15:28:02.641 +0000: Embulk v0.9.17
2019-04-11 15:28:03.422 +0000 [WARN] (main): DEPRECATION: JRuby org.jruby.embed.ScriptingContainer is directly injected.
2019-04-11 15:28:07.500 +0000 [INFO] (main): Gem's home and path are set by default: "C:\Users\SnowflakeETL_prd_svc\.embulk\lib\gems"
2019-04-11 15:28:10.954 +0000 [INFO] (main): Started Embulk v0.9.17
2019-04-11 15:28:11.032 +0000 [INFO] (0001:transaction): Loaded plugin embulk-input-sqlserver (0.9.3)
2019-04-11 15:28:11.094 +0000 [INFO] (0001:transaction): Loaded plugin embulk-output-azure_blob_storage (0.1.7)
[...]

PR #1181, opened 2019-09-27
Log stacktrace at TempFileSpace#cleanup

dmikurube:
Errors are sometimes observed that temporary files are not found.
* https://github.com/embulk/embulk/issues/1142
* https://github.com/embulk/embulk/issues/1151

To investigate this issue, adding a log message per TempFileSpace#cleanup.


dmikurube:
Thanks!

PR #1183, opened 2019-10-03
Bump up to v0.9.20-SNAPSHOT

dmikurube:


dmikurube:
@shroman Could you have a look? (#1189 and #1190 will follow.)

PR #1186, opened 2019-10-08
hashCode() for classes overriding equals(). fixes #1185

### LINKED ISSUE #1185: Overriding equals but not overriding hashCode
As for the current use, it's a non-issue, but it is a potential problem.
Need to be fixed to avoid violation of the general contract for Object.hashCode().

Classes affected:
`QuoteCharacter`, `EscapeCharacter`, `RubyTimeFormatToken`, `RubyTimeFormat`, `Immediate`.

shroman:
adding missing hashCode() for classes overriding equals().

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/time/RubyTimeFormat.java
+++ b/embulk-core/src/main/java/org/embulk/spi/time/RubyTimeFormat.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * RubyTimeFormat represents a Ruby-compatible time format.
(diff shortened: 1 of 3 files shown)


shroman:
@dmikurube sure. Btw, what is the specific reason not to include it into the next release? Do we make a distinction between types of fixes to include to some types of releases?

dmikurube:
We've confirmed that v0.9.20 fixed #1151, and #1196 has been merged. Let's go merging this.

As ed17c5680fbb6df5997c9adb94e2a55cee4b1b35 was not compliant with the CI, we may want to squash the commits so that .

----

(I usually did local squash and force push for such a case so far so that we have a "merge commit" in the commit history. But now, we may want to change the custom in this OSS `embulk` repo.

Ping me if you want to talk about it, or please go ahead with merging. But at least, please keep the PR number (`#1186`) in the squashed commit message so that we can find this PR later from Git commit history.)

shroman:
@dmikurube +1 for squashing with pr number.

PR #1188, opened 2019-10-22
Closing JarInputStream on failure. fixes #1184

### LINKED ISSUE #1184: Potential resource leak in SelfContainedJarFile
SelfContainedJarFile doesn't close `JarInputStream`.
https://github.com/embulk/embulk/blob/master/embulk-core/src/main/java/org/embulk/deps/SelfContainedJarFile.java#L60

This is a potential resource leak.

shroman:
Closing JarInputStream on failure.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/deps/SelfContainedJarFile.java
+++ b/embulk-core/src/main/java/org/embulk/deps/SelfContainedJarFile.java
@@ -55,7 +55,7 @@ final class SelfContainedJarFile {
             return;
         }
 
-        final JarInputStream jarInputStream;
+        JarInputStream jarInputStream = null;
         try {
             jarInputStream = new JarInputStream(inputStream, false);
         } catch (final IOException ex) {
@@ -64,6 +64,7 @@ final class SelfContainedJarFile {
             this.manifest = null;
             this.innerResources = null;
             this.innerResourcesBinary = null;
+            closeQuiet(jarInputStream);
             return;
         }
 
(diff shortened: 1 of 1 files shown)


embulk-core/src/main/java/org/embulk/deps/SelfContainedJarFile.java
```diff
@@ -55,7 +55,7 @@
             return;
         }
 
-        final JarInputStream jarInputStream;
+        JarInputStream jarInputStream = null;
```
dmikurube: Did you have special reasons why we needed to remove `final` and assign `null` here?

embulk-core/src/main/java/org/embulk/deps/SelfContainedJarFile.java
```diff
@@ -64,6 +64,7 @@
             this.manifest = null;
             this.innerResources = null;
             this.innerResourcesBinary = null;
+            closeQuiet(jarInputStream);
```
dmikurube: If the execution goes into this `catch` block, it means that it failed in constructing an instance of `JarInputStream`, then `jarInputStream` is not assigned. I don't think we need to close it, then?

shroman: you are right, I missed the place. Modified.

embulk-core/src/main/java/org/embulk/deps/SelfContainedJarFile.java
shroman: Not needed if the next issue you raised is addressed. thanks.

PR #1189, opened 2019-10-28
Create a TempFileSpace with ISO 8601's basic format

dmikurube:
`TempFileSpace`'s directory names have been like `"2019-05-15 11:35:48.617 UTC"`,
but whitespaces and colons could make everything weird.

This change makes it in ISO 8601's basic format, such as `"20190515T113548Z"`.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/ExecSession.java
+++ b/embulk-core/src/main/java/org/embulk/spi/ExecSession.java
@@ -1,6 +1,9 @@
 package org.embulk.spi;
 
 import com.google.inject.Injector;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
 import java.util.Optional;
 import org.embulk.config.Config;
 import org.embulk.config.ConfigDefault;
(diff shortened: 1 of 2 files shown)


dmikurube:
It addresses a part of: #1151

dmikurube:
@shroman Can you have a look?

dmikurube:
1. Unfortunately, `ExecSession` is a kind of difficult one to (unit) test... it's seriously tightly coupled with Guice `Injector`. :(
2. For the call from `ExecSession`, yes. But we have no guarantee that it is the only caller. We try to keep things unless we have guarantee or before we go into the v0.10 development series.

shroman:
1. I meant enforcing it with tests for `TempFileSpaceAllocator#newSpace`
2. For others, ideally they will take care of it by doing the parameter's transformations before calling it ideally. This is a tiny thing, but let's introduce a special comment for places that has to be improved in v0.10 or have an issue open for it.

dmikurube:
@shroman Thanks! Got it.

Added 2 tests in ba84e919c74eebfd0edc9c60adaa5621f322d9ea. They both now pass as-is. But the latter would be changed to fail (Exception?) from some point in the v0.10 series.

In addition, another PR #1192 is to add a warning log to call it with ISO 8601's basic format.

dmikurube:
As the returned `TempFileSpace` does not open its properties public, so I wanted at least to test `newSpace` would fail (throw an Exception) or not. We actually have a way to assert by trying to create a temporary file actually, and checking the path of the created file. But thought it could be overkilling to test `TempFileSpaceAllocator`. (It's almost testing `TempFileSpace`.)

Let me consider it later soon, but going with it for now for v0.9.20.

PR #1182, opened 2019-09-27
Test coverage and comments for TempFileSpace.

shroman:
Added unit tests and comments to `TempFileSpace`.

PR #1190, opened 2019-10-29
Guarantee TempFileSpace to be unique

dmikurube:
A `TempFileSpace` has been created in a directory with a timestamp, such as `"/tmp/2019-05-15 11:35:48.617 UTC"`.
But, when multiple Embulk processes (or tasks) are running, they can easily conflict.

To avoid such conflicts, this change starts to use `Files#createTempDirectory` to create a `TempFileSpace`.
Java VM guarantees that the directory created is unique.

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createTempDirectory-java.nio.file.Path-java.lang.String-java.nio.file.attribute.FileAttribute...-

The old constructor `TempFileSpace(File)` keeps the original behavior for compatibility,
but it is now deprecated. It should not be called from plugins, though.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/TempFileSpace.java
+++ b/embulk-core/src/main/java/org/embulk/spi/TempFileSpace.java
@@ -7,58 +7,131 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Responsible to manage a temporary directory (a space for temporary files) for a task.
+ *
+ * <p>Plugins are expected to get this through {@link Exec#getTempFileSpace}. Do not create an instance of this directly.
+ */
 public class TempFileSpace {
-    public TempFileSpace(final File dir) {
-        if (dir == null) {
-            throw new IllegalArgumentException("dir is null");
+    private TempFileSpace(final boolean isUnique, final Path baseDir, final String prefix, final File exactDir) {
+        this.isUnique = isUnique;
+        if (isUnique) {
+            this.baseDir = baseDir;
+            this.prefix = prefix;
+            this.exactDir = null;
+        } else {  // Backward-compatible behavior for `new TempFileSpace(File)`. To be removed.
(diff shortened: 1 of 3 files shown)


dmikurube:
@shroman It still needs tests, but before that, can you have a quick look at this for now? (Another commit for testing would follow later.)

dmikurube:
In 7e3b1cf53fb5ad1bbc03083d58446cf1e765bdf8, added tests that are based on @shroman's #1182.

embulk-core/src/main/java/org/embulk/spi/TempFileSpace.java
```diff
@@ -7,58 +7,135 @@
...
             throw new TempFileException(ex);
         }
     }
 
+    /**
+     * Cleans up the temporary file space, and everything in the space.
+     *
+     * <p>It is to be called along with Embulk's cleanup process. Plugins should not call this directly.
+     */
     public synchronized void cleanup() {
```
shroman: `Plugins should not call this directly` <- can we make it package-private?

shroman:
I think we can close #1182

embulk-core/src/main/java/org/embulk/spi/TempFileSpace.java
```diff
@@ -81,8 +158,57 @@ public FileVisitResult postVisitDirectory(final Path dirOnVisit, final IOExcepti
...
+
+    // The exact directory specified through the constructor TempFileSpace(File).
+    //
+    // Available only when created through the constructor TempFileSpace(File).
+    //
+    // TODO: Remove it once the constructor TempFileSpace(File) goes away.
+    private final File exactDir;
+
+    // The temporary directory created when creating the first temporary file.
+    private Optional<Path> tempDirectoryCreated;
```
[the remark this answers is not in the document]
dmikurube: Yeah, I actually thought a bit about it. But anyways, we can rethink about this as it's totally internal. ;)

embulk-core/src/main/java/org/embulk/spi/TempFileSpace.java
dmikurube: Possibly, but this method has been `public` for long, and we may call this out of this package even while in `embulk-core`.

dmikurube:
Thanks for the good catch! Addressed them. Will merge after CI seems happy.

PR #1195, opened 2019-11-14
Prepare a Gradle task to upload onto GitHub Package Registry

dmikurube:
It's not a functional change, but a change to upload the built artifacts to GitHub Package Registry. To be used in #1196.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,17 @@ buildscript {
     }
 }
 
+def getRevision = { ->
+    def stdout = new ByteArrayOutputStream()
+    exec {
+        workingDir "${rootDir}"
+        commandLine "git", "rev-parse", "HEAD"
+        standardOutput = stdout
+    }
+    return stdout.toString().trim()
+}
+def revision = getRevision()
+
 allprojects {  // Applies all projects including the root project as well.
     apply plugin: 'java'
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -130,13 +142,28 @@ configure(subprojects.findAll { ['embulk-core', 'embulk-deps-cli', 'embulk-deps-
...
             }
         }
+
+        // publishMavenPublicationToGithubRepository
+        repositories {
+            maven {
+                name = "github"
+                url "https://maven.pkg.github.com/embulk/embulk"
+                credentials {
+                    username = project.hasProperty("github_package_user") ? github_package_user : ""
```
shroman: shall we have in README something about specifying user and token?

build.gradle
```diff
@@ -20,6 +20,17 @@ buildscript {
     }
 }
 
+def getRevision = { ->
+    def stdout = new ByteArrayOutputStream()
```
shroman: what will happen if this block fails?
will it stop the build process or become `null`?

build.gradle
dmikurube: I don't think we need. This is basically core-team internal only for GitHub Actions. We won't expect individual developers never upload from their environment.

build.gradle
dmikurube: Hmm  it may stop. But thought we don't need to cover the case.

build.gradle
shroman: Right, just saving time for new core developers ;) Up to you.

build.gradle
shroman: Ok, I was just wondering if you want to cover it. LGTM.

dmikurube:
Thanks!

PR #1197, opened 2019-11-18
Refactor dependency class loaders to be extensible easier

dmikurube:
We plan to load more dependency libraries to be loaded through `DependencyClassLoader`. But now, we need annoying messy steps to extend them.

This pull request refactors `EmbulkDependencyClassLoaders` and `EmbulkSelfContainedJarFiles` to be parameterized so that it can be extended more easily.

The pull request consists of two steps:
* 1d2a6aca582315e45e78875815ac7d82da04ef74: focuses on `EmbulkDependencyClassLoaders`
* 3f2ed2f34d570322402964bc21a61fad87a59e42: focuses on `EmbulkSelfContainedJarFiles`

---

They are expected to be used like below. It is for the case of Embulk as the all-in-one executable package from CLI.
https://github.com/embulk/embulk/blob/v0.9.20/embulk-core/src/main/java/org/embulk/cli/Main.java#L60-L61

Users with `EmbulkEmbed` will need to call them by themselves.

Unfortunately it has been difficult to test as a unit test because it runs in the `static` context intentionally.

---

To find a corresponding `ClassLoader`, we need :

```
// Before
ClassLoader loader = EmbulkDependencyClassLoaders.ofMaven();
```

```
// After
ClassLoader loader = EmbulkDependencyClassLoaders.of(DependencyCategory.MAVEN);
```

### DIFF
--- a/embulk-core/src/main/java/org/embulk/deps/DependencyCategory.java
+++ b/embulk-core/src/main/java/org/embulk/deps/DependencyCategory.java
@@ -1,0 +1,23 @@
+package org.embulk.deps;
+
+public enum DependencyCategory {
+    CLI("CLI", EmbulkSelfContainedJarFiles.Type.CLI),
+    MAVEN("Maven", EmbulkSelfContainedJarFiles.Type.MAVEN),
+    ;
+
+    private DependencyCategory(final String name, final EmbulkSelfContainedJarFiles.Type selfContainType) {
+        this.name = name;
+        this.selfContainType = selfContainType;
+    }
+
+    public String getName() {
+        return this.name;
+    }
+
+    public EmbulkSelfContainedJarFiles.Type getSelfContainType() {
+        return this.selfContainType;
+    }
+
+    private final String name;
+    private final EmbulkSelfContainedJarFiles.Type selfContainType;
+}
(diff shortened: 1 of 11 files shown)


dmikurube:
@shroman Can you have a look when you have time?

embulk-core/src/main/java/org/embulk/deps/DependencyCategory.java
```diff
@@ -0,0 +1,23 @@
+package org.embulk.deps;
+
+public enum DependencyCategory {
+    CLI("CLI", "Embulk-Resource-Class-Path-Cli"),
+    MAVEN("Maven", "Embulk-Resource-Class-Path-Maven"),
```
shroman: It's worth having a simple test case just as a design contract, in case these are changed in gradle file.

shroman:
lgtm
To lower the hurdles of starting with Embulk for new devs, it's worth to have a simple README in https://github.com/embulk/embulk/tree/v0.9.20/embulk-core/src/main/java/org/embulk/cli to explain `EmbulkDependencyClassLoaders` and `EmbulkSelfContainedJarFiles` need to called (with an example). Up to you.

embulk-core/src/main/java/org/embulk/deps/DependencyCategory.java
dmikurube: Sounds reasonable. Added in 7734acb4fb05f0ae14c1beaa9406ed838e13b059

dmikurube:
Thanks! Added tests.

For explaining `EmbulkDependencyClassLoaders` and `EmbulkSelfContainedJarFiles`, let me consider where to add. Filed an issue #1202.

shroman:
excellent! looks good!

PR #1200, opened 2019-11-28
Throw a readable Exception against illegal prefix/suffix for a temp file (Fixes #1193)

### LINKED ISSUE #1193: "/" in prefix does not work for TempFileSpace#createTempFile since v0.9.20
Embulk v0.9.20 started to use `java.nio.file.Files.createTempFile`, instead of `java.io.File.createTempFile`, to create a temp file.

But, these two APIs had a difference when "/" is included in its `prefix`.

`File.createTempFile("fooA/barA", ".tmp", new File("/tmp"));`
==> Creates `/tmp/barA9003569178568216717.tmp`

`Files.createTempFile(Paths.get("/tmp"), "fooB/barB", ".tmp");`
==> Throws `IllegalArgumentException`

At least `embulk-output-ftp` has been affected by this change.
https://github.com/embulk/embulk-output-ftp/pull/26

We may want :
* Explicit validation and more understandable `Exception` about it, or
* Cover that case on the embulk-core side.

dmikurube:
It throws a more readable Exception against illegal prefix/suffix for a temp file (#1193).

@shroman Can you have a look?

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/TempFileSpace.java
+++ b/embulk-core/src/main/java/org/embulk/spi/TempFileSpace.java
@@ -105,7 +105,18 @@ public class TempFileSpace {
         try {
             this.createTempDirectoryIfRequired();
 
-            final Path tempFile = Files.createTempFile(this.tempDirectoryCreated.get(), prefix, "." + fileExt);
+            final Path tempFile;
+            try {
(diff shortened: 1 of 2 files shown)


dmikurube:
Thanks!

PR #1201, opened 2019-11-28
Load Netty Buffer library in a separate DependencyClassLoader

dmikurube:
After refactoring at #1197 and #1203, as the first effort of isolating existing dependency libraries, we are moving this Netty Buffer to a separate `DependencyClassLoader`.

Because the caller of Netty Buffer was only `PooledBufferAllocator`, made it an entry point to the `DependencyClassLoader` part.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,4 @@ env:
     - GRADLE_TASK=':embulk-core:check'
     - GRADLE_TASK=':embulk-standards:check'
     - GRADLE_TASK=':embulk-test:check'
+    - GRADLE_TASK=':embulk-deps-buffer:check'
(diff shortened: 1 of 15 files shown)


dmikurube:
@shroman Can you have a look? It's not hurry as it won't be included in next v0.9.21, but to be in v0.9.22.

It'll be one example of how we "isolate a dependency library to a sub ClassLoader".

dmikurube:
Thanks!

Basically agreed to have more Javadocs, but checkstyle is sometimes ritual and dogmatic. We can proceed with efforts to add more Javadocs in parallel to normal improvements. And then we may remove these suppressions when they're unnecessary finally.

Just ritual Javadocs can sometimes be worse than no Javadocs. Nice to dive deeper into background of the classes/methods, and leave them in the Javadocs, like done in: https://github.com/embulk/embulk/pull/1190/files#diff-80a18ead6a79aee5a1516dc78aefbad3R39

PR #1203, opened 2019-12-02
Fix refactored EmbulkDependencyClassLoaders to process JARs out of executable-JAR

dmikurube:
The "double initialization" check for JARs out of self-contained JAR was wrong in: 1d2a6aca582315e45e78875815ac7d82da04ef74
It was working for self-contained executable JAR, but it was not working for the EmbulkEmbed case, or other testing cases.


dmikurube:
A follow-up fix for #1197. Can you have a look, @shroman ?

dmikurube:
Thanks!

PR #1204, opened 2019-12-02
Load Airlift Slice library in a separate DependencyClassLoader

dmikurube:
After refactoring at #1201, we are moving this Airlift's Slice to the separate DependencyClassLoader for `Buffer`, the same as Netty Buffer.

Airlift's Slice is called from `PageBuilder` and `PageReader`. The entry point is `io.airlift.slice.Slice` here.


dmikurube:
Thanks!

PR #1205, opened 2019-12-03
Load ICU4J library in a separate DependencyClassLoader

dmikurube:
Next to #1201 (Netty Buffer) and #1204 (Airlift Slice), we're moving ICU4J that has been used for charset detection in Embulk's guess.

The entry point is just `CharsetDetector`. But as `CharsetDetector#detect` returns `CharsetMatch`, we have `CharsetMatch`, too.

### DIFF
--- a/embulk-deps/guess/src/main/java/org/embulk/deps/guess/CharsetDetectorImpl.java
+++ b/embulk-deps/guess/src/main/java/org/embulk/deps/guess/CharsetDetectorImpl.java
@@ -1,0 +1,22 @@
+package org.embulk.deps.guess;
+
+import com.ibm.icu.text.CharsetDetector;
+
+public final class CharsetDetectorImpl extends org.embulk.deps.guess.CharsetDetector {
+    public CharsetDetectorImpl() {
+        this.detector = new CharsetDetector();
+    }
+
+    @Override
+    public CharsetDetectorImpl setText(final byte[] in) {
+        this.detector.setText(in);
+        return this;
+    }
+
+    @Override
+    public CharsetMatchImpl detect() {
+        return new CharsetMatchImpl(this.detector.detect());
+    }
+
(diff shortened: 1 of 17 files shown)


embulk-deps/guess/src/main/java/org/embulk/deps/guess/CharsetDetectorImpl.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Override
+    public CharsetDetectorImpl setText(final byte[] in) {
+        this.detector.setText(in);
+        return this;
+    }
+
+    @Override
+    public CharsetMatchImpl detect() {
+        return new CharsetMatchImpl(this.detector.detect());
```
shroman: Not quite sure, but won't it delegate the loading request to the parent class loader?

dmikurube: If I'm not making mistakes...
* This `CharsetMatchImpl` is `org.embulk.deps.guess. CharsetMatchImpl ` defined in the same package `embulk-deps-guess` (just below) loaded by the sub ClassLoader.
* The `detector` instance is `com.ibm.icu.text.CharsetDetector` (as in `import`), which is ICU4J's also loaded by the sub ClassLoader.
* Then, `detector.detect()` returns `com.ibm.icu.text.CharsetMatch`, which is in ICU4J again loaded by the sub ClassLoader, and that is used to construct `org.embulk.deps.CharsetMatchImpl`.

So, I believe it completes in the sub ClassLoader, but if you still see points in question, please let me know.

shroman: forgot it's in ` embulk-deps/guess`

dmikurube:
Thanks!

PR #1207, opened 2019-12-05
Check dependency vulnerability semi-automatically with org.owasp:dependency-check-gradle

dmikurube:
GitHub's vulnerability alert does not support Gradle unfortunately, but we'd want some checks for vulnerability. As the very first step, introducing a dependency checker Gradle plugin in our build.gradle.


dmikurube:
Thanks!

PR #1211, opened 2019-12-12
Load SnakeYaml in a separate DependencyClassLoader.

shroman:
moving Yaml conversion to a separate DependencyClassLoader.
Related PRs: #1201 (Netty Buffer) and #1204 (Airlift Slice).

Used in: org.embulk.config.ConfigLoader and org/embulk/EmbulkRunner

PR #1216, opened 2019-12-24
Embed embulk-deps-config with SnakeYaml in executable JARs

dmikurube:
Sorry that I was not aware of this at #1211. The separated `embulk-deps-config` was not included in the finally-generated executable JAR files.

This PR would add them.


dmikurube:
@shroman Could you have a look?

dmikurube:
Thanks!

PR #1224, opened 2020-03-17
Remove "embulk selfupdate" to the latest version

dmikurube:
At first, `embulk selfupdate` won't support updating to the latest version.


dmikurube:
Thanks!

PR #1225, opened 2020-03-17
Upgrade to Gradle 6.2.2

dmikurube:
Through v0.10, we don't take too much care about compatibility. Then, we try updating things to the latest!

It enables dependency locks along with that.


dmikurube:
Thanks!

PR #1226, opened 2020-03-17
Rename embulk-test to embulk-junit4

dmikurube:
`embulk-test` assumes JUnit 4. But we have JUnit 5 in the world, and JUnit 5 is totally different from JUnit 4. On plugin's side, they will need to switch from JUnit 4 to JUnit 5 intentionally and explicitly.

We rename `embulk-test` to `embulk-junit4` so that JUnit 4 and JUnit 5 can co-exist for a while.


shroman:
two cents: maybe `junit` without `4`? Upgrading to `5` in near future may be feasible.

dmikurube:
@shroman It's intentional. In the migration period, we would have both `embulk-junit4` and `embulk-junit5`.

dmikurube:
Thanks!

PR #1227, opened 2020-03-23
Build an executable JAR file without the Gradle Shadow plugin

dmikurube:
Embulk's executable JAR no longer needs package relocation. The Gradle Shadow plugin is no longer needed. Dependency JARs can be unzipped in build.gradle.

The Gradle Shadow plugin is too complicated inside, and has some potential problems as commented in `build.gradle`. It's reasonable to stop using it if we need just a simple thing. It was still reasonable to use it when we needed package relocation, though.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,6 @@ plugins {
     id "java"
     id "maven-publish"
     id "jacoco"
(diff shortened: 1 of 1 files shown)


dmikurube:
Thanks!

PR #1229, opened 2020-03-23
Stop uploading the executable JAR to Bintray, and with declaring about Bintray more explicitly

dmikurube:
Embulk's executable binaries were uploaded to Bintray, and `selfupdate` was choosing the "latest" version by Bintray's automated redirect from: `https://bintray.com/embulk/maven/embulk/_latestVersion`. The URL was hard-coded in Embulk.

It was a lock-in. Too bad. We removed the hard-code, and started to download from `https://dl.embulk.org/...` in #1132 (v0.9.17). But, Embulk executables older than v0.9.17 were still trying to access Bintray. We needed to keep releasing executable binaries to Bintray.
(At the same time, we started to release executable binaries to GitHub Releases, and https://dl.embulk.org/...` is redirected to GitHub Releases.)

Now, we want older Embulk users to stop getting auto-updated to v0.10 series. We have to stop uploading executable binaries to Bintray, then.

On the other hand, each Maven artifact (`embulk-core`, `embulk-standards`, ...) is still uploaded to Bintray.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,18 @@ ext {
     summary = 'Embulk, a plugin-based parallel bulk data loader'
 }
 
-configure(subprojects.findAll { ['embulk-core', 'embulk-deps-buffer', 'embulk-deps-cli', 'embulk-deps-guess', 'embulk-deps-config', 'embulk-deps-maven', 'embulk-standards', 'embulk-junit4'].contains(it.name) }) {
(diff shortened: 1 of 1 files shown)


dmikurube:
@shroman Good catch! Left a note above.

dmikurube:
Thanks!

PR #1223, opened 2020-03-17
Bump up to v0.10.0-SNAPSHOT

dmikurube:
v0.9.23 looks working fine. It's time to go v0.10!

At the same time, it once removes the setting of GitHub Actions as it is failing.

PR #1196, opened 2019-11-14
Build with GitHub Actions to upload onto GitHub Package Registry every build

dmikurube:
WIth the GitHub Action, artifacts are uploaded to GitHub Package Registery every build. 

For this Pull Request at `2d44f83bf15952360013b6d2d1d445201a835c00`,

Action :
* https://github.com/embulk/embulk/commit/2d44f83bf15952360013b6d2d1d445201a835c00/checks

Artifacts :
* https://github.com/embulk/embulk/packages/54702?version=0.9.21-SNAPSHOT-2d44f83bf15952360013b6d2d1d445201a835c00
* https://github.com/embulk/embulk/packages/54714?version=0.9.21-SNAPSHOT-2d44f83bf15952360013b6d2d1d445201a835c00
* https://github.com/embulk/embulk/packages/54715?version=0.9.21-SNAPSHOT-2d44f83bf15952360013b6d2d1d445201a835c00
* https://github.com/embulk/embulk/packages/54716?version=0.9.21-SNAPSHOT-2d44f83bf15952360013b6d2d1d445201a835c00
* https://github.com/embulk/embulk/packages/54717?version=0.9.21-SNAPSHOT-2d44f83bf15952360013b6d2d1d445201a835c00

PR #1231, opened 2020-03-24
Test with GitHub Actions on Ubuntu

dmikurube:
* It disables Travis-CI at the same time.
* It still tests on Windows by Appveyor.
* It does not upload to GitHub Package Registry, which started in #1196, and was disabled in #1223.
* It disables updating https://www.embulk.org/docs/ for a while. The update would be done in a different way later.


dmikurube:
Thanks!

PR #1232, opened 2020-03-25
Split `org.embulk.spi.Buffer` into a separate artifact `embulk-api`

dmikurube:
As described in `docs/design/api_and_spi.md`, going to split some classes into different artifacts `embulk-api` and `embulk-spi` so that Embulk plugins will need to depend only on them, not depending on `embulk-core`.

`Buffer` is the first step because it does not depend on other classes, and used from other classes. `Buffer` in `embulk-api` is an abstract class. The implementation is `BufferImpl` in `embulk-core`.

`embulk-api` and `embulk-spi` would be released to Maven Central, not Bintray nor JCenter. Common build tools often refer Maven Central by default. Other Maven repositories like JCenter usually inherit Maven Central.

The `embulk-api` artifact is to have the `LICENSE` file from :
https://www.apache.org/licenses/LICENSE-2.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ plugins {
     id "java"
     id "maven-publish"
     id "jacoco"
-    id "com.jfrog.bintray" version "1.8.4"
+    id "com.jfrog.bintray" version "1.8.4" apply false
     id "org.ajoberstar.grgit" version "4.0.1"
 }
 
@@ -42,7 +42,11 @@ ext {
     summary = 'Embulk, a plugin-based parallel bulk data loader'
 }
 
-def subprojectNamesReleased = [
+def subprojectNamesReleasedInMavenCentral = [
+    "embulk-api",
+]
+
+def subprojectNamesReleasedInBintray = [
     "embulk-core",
     "embulk-deps-buffer",
     "embulk-deps-cli",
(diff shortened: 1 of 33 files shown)


build.gradle
```diff
@@ -186,6 +177,103 @@ configure(subprojects.findAll { subprojectNamesReleased.contains(it.name) }) {
...
+                            name = "Satoshi Akama"
+                            email = "satoshiakama@gmail.com"
+                        }
+                        developer {
+                            name = "Dai MIKURUBE"
+                            email = "dmikurube@treasure-data.com"
+                        }
+                        developer {
+                            name = "Shinichi Ishimura"
+                            email = "shiketaudonko41@gmail.com"
```
dmikurube: @kamatama41 Do you mind if we include you, with your e-mail address in Embulk's Git log, as one of the `<developers>` in POM of Embulk Maven artifacts? It is to be uploaded in Maven Central. (We haven't have `<developers>` in Bintray, but it is mandatory in Maven Central.)

kamatama41: @dmikurube 
No problem, I'm so proud of it 😃 
(I haven't checked recent Embulk topics for a long time, though)

dmikurube: Thanks! Yeah, we're happy to have you in the list. :)

embulk-api/src/main/html/overview.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
shroman, line 1: why do we need this doc? How about having `package-info.java` instead?

dmikurube: It's not very important, but just wanted to have some fundamental HTML elements from this in the top-level `index.html`. IIUC, `package-info.java` is for the package-level, not the top-level.

dmikurube:
Thanks!

PR #1237, opened 2020-03-30
Split some `org.embulk.config` classes into `embulk-api`

dmikurube:
Following #1232, moving some interfaces in `org.embulk.config` to `embulk-api`, too.


dmikurube:
Yeah, I'm trying to add Javadocs along with the spilt. ;)

PR #1239, opened 2020-04-23
Remove direct injection of org.jruby.embed.ScriptingContainer (Fix #1007)

### LINKED ISSUE #1007: Drop injection of `org.jruby.embed.ScriptingContainer`
JRuby runtime is now indirected from Embulk. Injecting `org.jruby.embed.ScriptingContainer` is to be dropped after a while.

dmikurube:
`org.jruby.embed.ScriptingContainer` has still been injected for some plugins still getting `ScriptingContainer`, but we are going to unsupport it in v0.10. The JRuby runtime will be loaded just optionally at all, under a sub ClassLoader.


dmikurube:
Thanks!

PR #1242, opened 2020-04-28
Split Exception classes into `embulk-api`

dmikurube:


shroman:
I find it more natural to have exceptions in packages of classes that use them. Can you provide the reasoning why moving them is necessary?

dmikurube:
Simply because plugins need to throw those Exceptions, and Embulk (core) is going to handle those Exceptions in it.

As explained separately, plugins are going to depend only on embulk-api and embulk-spi.

dmikurube:
Thanks!

PR #1245, opened 2020-05-13
Stop using org.jruby.util.RubyDateFormat, and replace with embulk-util-rubytime under a sub ClassLoader (Fix #830)

### LINKED ISSUE #830: Avoid using org.jruby.util.RubyDateFormat in TimestampFormatter

dmikurube:
See #830 for the details. In short, this use of `org.jruby.util.RubyDataFormat` is the blocker to make JRuby optional.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,6 +53,7 @@ def subprojectNamesReleasedInBintray = [
     "embulk-deps-guess",
     "embulk-deps-config",
     "embulk-deps-maven",
+    "embulk-deps-timestamp",
     "embulk-standards",
     "embulk-junit4",
 ]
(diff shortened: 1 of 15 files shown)


dmikurube:
We'll have more cleanups to depend more on `embulk-util-rubytime`. All those implementation classes will go away finally then.

But anyways, the formatting specification (e.g. `%Y%m%d`) is officially declared to be "Ruby's" (compatible) in Embulk, even though we no longer use (J)Ruby's implementation as-is. I think we'll keep the name of Ruby somewhere.

PR #1246, opened 2020-05-15
Add a new `toJson` method in `org.embulk.config.DataSource`

dmikurube:
We're going to separate the config processing part (mapping `ConfigSource`/`TaskSource` into `(Plugin)Task` interface) into another utility artifact `embulk-util-config` to process configs on plugin's side.

To realize that, we have to have a method to retrieve `DataSource`'s internal structure without leaking Jackson instances, but as of now, we only have `getAttributes` and `getObjectNode` to return Jackson instances.

To retrieve `DataSource`'s internal structure, this PR is to add a new `toJson` method to stringify `DataSource`'s internal JSON structure so that `embulk-util-config` can rebuild its own JSON structure.

### DIFF
--- a/embulk-api/src/main/java/org/embulk/config/ConfigSource.java
+++ b/embulk-api/src/main/java/org/embulk/config/ConfigSource.java
@@ -31,11 +31,17 @@ public interface ConfigSource extends DataSource {
     /**
      * Loads this configuration as a task class like {@code PluginTask}.
      *
+     * <p>This method will be deprecated. Loading configs will be replaced by another library
+     * {@code embulk-util-config} on plugin's side.
+     *
      * @param <T>   the task class to load this configuration as
      * @param taskType  the task class to load this configuration as
      * @return a task class instance loaded
      */
-    <T> T loadConfig(Class<T> taskType);
+    default <T> T loadConfig(Class<T> taskType) {
+        throw new UnsupportedOperationException(
+                "ConfigSource#loadConfig does not work with: " + this.getClass().getCanonicalName());
+    }
 
     /**
      * Returns a nested value under {@code attrName}.
--- a/embulk-core/src/main/java/org/embulk/config/DataSourceImpl.java
+++ b/embulk-core/src/main/java/org/embulk/config/DataSourceImpl.java
@@ -177,6 +177,11 @@ public class DataSourceImpl
         return this;
     }
 
+    @Override
+    public String toJson() {
+        return this.model.writeObject(this.data);
+    }
+
     private static void mergeJsonObject(ObjectNode src, ObjectNode other) {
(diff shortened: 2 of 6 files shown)


embulk-api/src/main/java/org/embulk/config/ConfigSource.java
```diff
@@ -31,11 +31,17 @@
     /**
      * Loads this configuration as a task class like {@code PluginTask}.
      *
+     * <p>This method will be deprecated. Loading configs will be replaced by another library
```
shroman, line 34: Nit: Having `TODO` or `@Deprecated` (probably not what you want, since don't deprecate now) would make it easier to find it later.

embulk-core/src/main/java/org/embulk/config/DataSourceImpl.java
```diff
@@ -128,28 +128,40 @@ public DataSourceImpl set(String attrName, Object v) {
...
-            throw new IllegalArgumentException("DataSourceImpl#setNested aceepts only DataSourceImpl.");
+            final String vJsonStringified = v.toJson();
+            if (vJsonStringified == null) {
+                throw new ConfigException(new NullPointerException("DataSource#setNested accepts only valid DataSource"));
+            }
+            final JsonNode vJsonNode = this.model.readObject(JsonNode.class, vJsonStringified);
+            if (!vJsonNode.isObject()) {
+                throw new ConfigException(new ClassCastException("DataSource#setNested accepts only valid JSON object"));
+            }
+            this.data.set(attrName, (ObjectNode) vJsonNode);
```
shroman, line 144: We don't need this cast since checked if it's `JsonObject`.

embulk-api/src/main/java/org/embulk/config/ConfigSource.java
dmikurube: Yeah, but it's not deprecated yet as you said. The purpose of this comment is to describe it "will be" deprecated.

Also, this is published Javadoc, not our internal comments. TODO may not fit for Javadoc.

embulk-core/src/main/java/org/embulk/config/DataSourceImpl.java
dmikurube: Ah, that's true. `ObjectNode#set` receives `JsonNode`.
https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/node/ObjectNode.html#set-java.lang.String-com.fasterxml.jackson.databind.JsonNode-

But I think it's okay to keep it as it has no cost, and we can explicitly declare it should be `ObjectNode` here. If you think it's to be removed, please let me know.

dmikurube:
Thanks!

PR #1248, opened 2020-05-26
Implement a default implementation for org.embulk.config.DataSource#toJson

dmikurube:
The `toJson` method is newly added in DataSource. Nice to have a default implementation
for it so that an unexpected combination of newer DataSource and older DataSourceImpl
won't cause ununderstandable errors.


dmikurube:
Thanks!

PR #1249, opened 2020-05-29
Not to throw ConfigException directly from SerDe

dmikurube:
`JsonSerializer` is basically expected to throw `JsonGenerationException`, not `ConfigException` (which is `RuntimeException`).

(It was added in https://github.com/embulk/embulk/pull/1246/files#diff-a6bb4457efbba2979dc038a940f2f6e0.)


dmikurube:
Thanks!

PR #1250, opened 2020-05-29
Use "api" in "java-library" to build embulk-core

dmikurube:
Since v0.10.0 after Embulk started using "implementation" to declare dependencies instead of "compile", embulk-core's dependencies are defined as "runtime" instead of "compile" in their pom.xml.

https://bintray.com/embulk/maven/embulk/0.9.23#files/org%2Fembulk%2Fembulk-core%2F0.9.23
https://bintray.com/embulk/maven/embulk/0.10.0#files/org%2Fembulk%2Fembulk-core%2F0.10.0

Embulk plugins have expected that embulk-core's dependencies are declared as "compile" in pom.xml. So, plugins didn't declare such dependencies of embulk-core as their own dependencies.

As of now, keeping them as "compile" would be better because:
* This change was unexpected.
* We plan to remove many finally from direct dependencies of embulk-core.
  Plugins are expected to follow when the removal happens, not by this unexpected change.
  Impacting plugins multiple times is not very good.

### DIFF
--- a/embulk-core/build.gradle
+++ b/embulk-core/build.gradle
@@ -11,8 +11,8 @@ processResources.from("${buildDir}/embulk_gem_as_resources", "${buildDir}/depend
 configurations {
     // com.google.inject:guice depends on asm and cglib but version of the libraries conflict
     // with ones bundled in jruby-complete and cause bytecode compatibility error
-    implementation.exclude group: "asm", module: "asm"
-    implementation.exclude group: "org.sonatype.sisu.inject", module: "cglib"
+    api.exclude group: "asm", module: "asm"
+    api.exclude group: "org.sonatype.sisu.inject", module: "cglib"
     compileClasspath.resolutionStrategy.activateDependencyLocking()
     runtimeClasspath.resolutionStrategy.activateDependencyLocking()
 }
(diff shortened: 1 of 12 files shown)


embulk-core/build.gradle
```diff
@@ -23,30 +23,32 @@ repositories {
...
-    implementation "com.fasterxml.jackson.module:jackson-module-guice:2.6.7"
-    implementation "org.slf4j:slf4j-api:1.7.12"
-    implementation "org.jruby:jruby-complete:" + rootProject.jrubyVersion
-    implementation "javax.validation:validation-api:1.1.0.Final"
-    implementation "org.apache.bval:bval-jsr303:0.5"
-    implementation "joda-time:joda-time:2.9.2"
-    implementation "org.msgpack:msgpack-core:0.8.11"
+    // Embulk plugins have expected that embulk-core's dependencies are declared as "compile" in pom.xml.
+    // Although those dependencies would be finally removed from embulk-core's direct dependencies,
+    // those "compile" declaration would be kept so that plugin developers won't be confused unnecessarily.
```
shroman, line 28: This comment says this will be moved to `implementation` eventually? Do I understand correctly `removed`?

dmikurube: Eventually, many of them are "removed" from this artifact `embulk-core`, and go to `embulk-deps-*` to be behind sub ClassLoaders.

Some may remain in `embulk-core`. TBD to keep `api` or back to `implementation`.

PR #1255, opened 2020-06-16
Remove subcommands `migrate`, `exec`, and `irb` which are no longer supported

dmikurube:


dmikurube:
Thanks. Actually, a kind of CHANGELOG is in git tag annotations (and GitHub Releases https://github.com/embulk/embulk/releases).

shroman:
Ah, indeed 👍

PR #1256, opened 2020-06-16
Remove ConfigSource-based system config (Fix #1159)

### LINKED ISSUE #1159: Simplify system config to be in java.util.Properties
Embulk's "System Config" has been represented in `ConfigSource`, the same structure as user Embulk configs (YAML) based on Jackson. But in almost all the use-cases, we just put `String`:`String` mappings in System Config, except for some internal use-case.

We need many (initialization, libraries, ...) to use `ConfigSource`. But at the same time, we'd like to use the System Configs earlier in Embulk's process.

It'd be very easy to use System Configs in the very beginning, by representing System Configs with Java's standard `java.util.Properties`.

dmikurube:
`ConfigSource`-based system config is no longer used. In v0.10, it's time to remove it!


dmikurube:
Thanks!

PR #1257, opened 2020-06-16
Stop Timestamp-based transaction_time

dmikurube:
The `transaction_time` feature has been using Guava Optional and Joda-Time-based `Timestamp`, but they can be replaced to `java.util.Optional<java.util.Instant>`.

We are keeping `Timestamp`-based APIs for now, but they may be removed later.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
+++ b/embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
@@ -330,7 +330,7 @@ public class BulkLoader {
             List<Optional<TaskReport>> inputTaskReports = (inputTaskStates == null) ? null : getInputTaskReports();
             List<Optional<TaskReport>> outputTaskReports = (outputTaskStates == null) ? null : getOutputTaskReports();
             return new ResumeState(
-                    exec.getSessionExecConfig(),
+                    exec.newConfigSource().set("transaction_time", exec.getTransactionTimeString()),
                     inputTaskSource, outputTaskSource,
                     inputSchema, executorSchema,
                     inputTaskReports, outputTaskReports);
--- a/embulk-core/src/main/java/org/embulk/spi/ExecSession.java
+++ b/embulk-core/src/main/java/org/embulk/spi/ExecSession.java
@@ -1,10 +1,14 @@
 package org.embulk.spi;
 
 import com.google.inject.Injector;
+import java.time.Instant;
+import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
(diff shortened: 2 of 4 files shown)


embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
```diff
@@ -330,7 +330,7 @@ public ResumeState buildResumeState(ExecSession exec) {
             List<Optional<TaskReport>> inputTaskReports = (inputTaskStates == null) ? null : getInputTaskReports();
             List<Optional<TaskReport>> outputTaskReports = (outputTaskStates == null) ? null : getOutputTaskReports();
             return new ResumeState(
-                    exec.getSessionExecConfig(),
+                    exec.newConfigSource().set("transaction_time", exec.getTransactionTimeString()),
```
shroman, line 333: looks like `exec.getSessionExecConfig()` already doing the same thing

embulk-core/src/main/java/org/embulk/spi/ExecSession.java
```diff
@@ -196,4 +222,62 @@ public boolean isPreview() {
     public void cleanup() {
         tempFileSpace.cleanup();
     }
+
+    private static Optional<Instant> toInstantFromString(final String string) {
```
shroman, line 226: I'd be more confident having some way to test these private methods, but I leave it to you.

embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
dmikurube: Yeah, but `Exec.getSessionExecConfig` is now `@Deprecated`, and we intend it to be removed finally. That's there only for compatibilify for a while. Rather than `@SuppressWarnings("deprecation")`, I chose to do the same here.

embulk-core/src/main/java/org/embulk/spi/ExecSession.java
dmikurube: Agreed. Added tests in fe530f0ffd14e706df86157c80d14ad056ac700a

embulk-core/src/main/java/org/embulk/exec/BulkLoader.java
shroman: oops, true:) I missed the annotation

embulk-core/src/main/java/org/embulk/spi/ExecSession.java
shroman: Thanks!

dmikurube:
Thanks!

PR #1258, opened 2020-06-16
Call LoggerFactory.getLogger more directly to prepare for removing logger binding from Guice

dmikurube:
`Exec.getLogger` is going to be replaced with SLF4J's direct `LoggerFactory.getLogger`, and its background `ILoggerFactory` Guice binding would be removed.

This is the last step of the replacement inside `embulk-core`.


dmikurube:
Thanks!

PR #1259, opened 2020-06-22
Construct and maintain PluginClassLoaderFactory out of Guice

dmikurube:
We're going to get Guice out of Embulk at least from the core components because :

* Guice `Injector` is now used as a global variable container. Everything is accessing global values there with `getInjector` and `getInstance`.
* Guice is now composing the key part of the core, and it is taking Guava along. We have the "version update deadlock" problem also in Guava, like Jackson -- who would update Guava at first, the core or plugins. To release Embulk from this Guava problem, Guice (with Guava) needs to be removed from the core.

This PR is the first step of the efforts.


dmikurube:
It may be easier to look at each commit, not the entire PR.

embulk-core/src/main/ruby/embulk/java_plugin.rb
```diff
@@ -6,7 +6,15 @@ class JavaPlugin
     def self.classloader(dir)
       jars = Dir["#{dir}/**/*.jar"]
       urls = jars.map {|jar| java.io.File.new(File.expand_path(jar)).toURI.toURL }
-      factory = Java.injector.getInstance(Java::PluginClassLoaderFactory.java_class)
+      begin
+        expected_temporary_variable_name = Java::org.embulk.jruby.JRubyPluginSource::PLUGIN_CLASS_LOADER_FACTORY_VARIABLE_NAME
+      rescue => e
+        raise PluginLoadError.new "Java's org.embulk.jruby.JRubyPluginSource does not define PLUGIN_CLASS_LOADER_FACTORY_VARIABLE_NAM unexpectedly."
```
shroman, line 12: typo: `PLUGIN_CLASS_LOADER_FACTORY_VARIABLE_NAME`

embulk-core/src/main/ruby/embulk/java_plugin.rb
```diff
@@ -6,7 +6,15 @@ class JavaPlugin
...
+      end
+      if expected_temporary_variable_name != "$temporary_internal_plugin_class_loader_factory__"
+        raise PluginLoadError.new "Java's org.embulk.jruby.JRubyPluginSource does not define PLUGIN_CLASS_LOADER_FACTORY_VARIABLE_NAM correctly."
```
shroman, line 15: typo `PLUGIN_CLASS_LOADER_FACTORY_VARIABLE_NAME`

dmikurube:
Thanks for the good catch! Going to merge.

PR #1262, opened 2020-06-24
Replace some SerDe classes to independent Jackson Modules

dmikurube:
For unknown reasons, some Jackson Serializers and Deserializers (`*SerDe`) have been implemented to register only into `ObjectMapperModule` (Guice `Module`) only for `ModelManager`, not as common Jackson `Module`s. (Note that: Guice `Module` and Jackson `Module` are totally different concepts.)

But, they can be implemented as Jackson `Module`s, and that would be more general-purpose. As we want to use them for general `ObjectMapper` (in next #1263), they would be implemented as Jackson `Module`s.


dmikurube:
Thanks!

PR #1263, opened 2020-06-24
Stop using ModelManager in PreviewPrinter

dmikurube:
To reduce dependency onto `ModelManager`, stopping to use `ModelManager` in `PreviewPrinter`.

`PreviewPrinter` is used only in the `embulk preview` command -- especially only when called via CLI. Not used from an embedded program using `EmbulkEmbed`.


dmikurube:
Thanks!

PR #1266, opened 2020-07-08
Remove embulk-docs which has moved to https://www.embulk.org/

dmikurube:
It removes duplicated contents in README.md and else as well.

----

https://www.embulk.org/docs/ was maintained via this `embulk-docs` with automated builds, but it's now maintained through https://github.com/embulk/embulk.github.io, too. We no longer need this `embulk-docs` subproject.


dmikurube:
Thanks!

PR #1267, opened 2020-07-15
Test ToStringMap

dmikurube:


dmikurube:
@shroman @himadieievsv Along with https://github.com/embulk/embulk-filter-calcite/pull/41, can you have a look at this?

Btw, as far as I searched for plugins we have access, `embulk-filter-calcite` is the only one plugin using this `ToString` or `ToStringMap`. Then, I think we're going to deprecate this.

dmikurube:
Thanks!

PR #1271, opened 2020-07-27
Log against loading classes, that are removed in JEP 320, from the parent ClassLoader

dmikurube:


embulk-core/src/main/java/org/embulk/plugin/PluginClassLoader.java
```diff
@@ -254,6 +269,124 @@ private boolean isParentFirstPath(String name) {
...
+        "javax.transaction",
+    };
+
+    private static Set<String> JEP_320_PACKAGES =
+            Collections.unmodifiableSet(new HashSet<String>(Arrays.asList(JEP_320_PACKAGES_ARRAY)));
+
     private final List<String> parentFirstPackagePrefixes;
     private final List<String> parentFirstResourcePrefixes;
+
+    private boolean hasJep320LoggedWithStackTrace;
```
shroman, line 391: A matter of preference, but it seems to me that having this flag is not very advantageous -- such logs are not likely to be observed frequently and it's fine to have them verbose. Also it may help to find transitive dependency issues faster.

dmikurube: When one class is loaded from one set of feature (e.g. JAXB), that is typically not the only one class. Related classes are loaded one after another in typical use-cases. So, when one class is logged, a lot more are logged. Most of them are not very informative, though.
https://github.com/embulk/embulk/pull/1271/files#diff-b05395f45e2d8aad0e598eaf4185c738R278

Also, the log is dumped almost always when the affected plugin is used, even though the user (including ourselves) is aware of that.

dmikurube:
Thanks!

PR #1274, opened 2020-08-04
Move FileInputPlugin and FileOutputPlugin to embulk-spi, and Transactional and TransactionalFile* to embulk-api

dmikurube:


dmikurube:
Thanks!

PR #1279, opened 2020-08-12
Merge all embulk-deps-* to embulk-deps

dmikurube:
Although we initially expected some scenarios that we might need to have hidden dependencies isolated per purpose/category, for example, scattered Jackson annotations might conflict something else.

After some investigation and cleanup work done, we concluded they can be cleaned up to a singleton `DependencyClassLoader` for all hidden dependencies.

This change merges all `embulk-deps-*` subprojects to a single `embulk-deps` subproject.


build.gradle
```diff
@@ -320,7 +305,7 @@ def listEmbedDependencies = { rootModuleName, prefix ->
         }
     }
     if (list.isEmpty()) {
-        throw new GradleException('Failed to collect libraries to embed: no dependencies found for: "' + rootModuleName + '"')
+        throw new GradleException('Failed to collect libraries to embed: no dependencies found')
```
shroman, line 308: `rootModuleName` is not useful anymore?

dmikurube: Ah, yeah, maybe. I kept it for now just because it is still used in the condition `firstLevelDependency.moduleName == rootModuleName`, and I wanted to make this change as small as possible (it's already big enough).

We'll want to revisit the enumeration logic, and simplify it more.

dmikurube:
Thanks! Merging it, and releasing v0.10.10.

PR #1281, opened 2020-08-13
Explicitly show the dump of an Exception for JEP 320 is a dummy Exception

dmikurube:
The log message introduced in #1271 is shown like:

```
2020-08-13 08:22:41.828 +0000 [INFO] (0001:transaction): Class javax.annotation.PostConstruct is loaded by the parent ClassLoader, which is removed by JEP 320. The plugin needs to include it on the plugin side. See https://github.com/embulk/embulk/issues/1270 for more details.
java.lang.Exception: null
	at org.embulk.plugin.PluginClassLoader.logInfoIfJep320Class(PluginClassLoader.java:281) [embulk-core-0.10.8.jar:0.10.8]
```

`java.lang.Exception: null` is much confusing. We'll want the stack trace to be shown as `DUMMY` explicitly.


dmikurube:
Thanks!

PR #1289, opened 2020-08-18
Replace Jackson annotations in Type to Jackson Module

dmikurube:
Replacing Jackson annotation on `org.embulk.spi.type.Type` to Jackson `Module`-based SerDe implementation.

This is needed to move `Type` to `embulk-api` that is Jackson-independent. As its side effect, `Type` will not be able to get converted out of `org.embulk.config.*` (including `ModelManager`), but we accept it.

We'll need another update for `embulk-util-config` once the new `embulk-api` including `Type` is released.


shroman (approved):
LGTM

But I have a question: why is `TypeDeserializer` public and `TypeSerializer` private inner?

himadieievsv (approved):
LGTM

dmikurube:
For `TypeDeserializer`, it's because it has already been `public`. I saw some plugins are accessing that class...

shroman:
Ah, got it!

dmikurube:
Thanks!

PR #1290, opened 2020-08-18
Replace Jackson annotations in Column to Jackson Module

dmikurube:
Replacing Jackson annotation on `org.embulk.spi.Column` to Jackson `Module`-based SerDe implementation.

This is needed to move `Column` to `embulk-api` that is Jackson-independent. As its side effect, `Column` will not be able to get converted out of `org.embulk.config.*` (including `ModelManager`), but we accept it.

We'll need another update for `embulk-util-config` once the new `embulk-api` including `Column` is released.

----

This PR consists of two commits:
* c16ffeb6ef8d65b8890faf05c4332dfa93de0073: Just add a test of mapping JSON to `Column`.
* 9470d2553891be2ae3b37f5cdf381af47e26d257: Actual replacement.

It is confirmed that c16ffeb6ef8d65b8890faf05c4332dfa93de0073 succeeds, and 9470d2553891be2ae3b37f5cdf381af47e26d257 still succeeds with the replacement.


shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1291, opened 2020-08-25
Replace Jackson annotations in Schema to Jackson Module

dmikurube:
Similar to #1290, replacing Jackson annotation on `org.embulk.spi.Schema` to Jackson `Module`-based SerDe implementation.

This is needed to move `Schema` to `embulk-api` that is Jackson-independent. As its side effect, `Schema` will not be able to get converted out of `org.embulk.config.*` (including `ModelManager`), but we accept it.

We'll need another update for `embulk-util-config` once the new `embulk-api` including `Schema` is released.

----

This PR consists of two commits:
* eafe1cc41e48d918acee5a717209e74d81db856b: Just add a test of mapping JSON to `Schema`.
* 0eb374f02822a48a8ae3f613ad01f84506267cb4: Actual replacement.

It is confirmed that eafe1cc41e48d918acee5a717209e74d81db856b succeeds, and 0eb374f02822a48a8ae3f613ad01f84506267cb4 still succeeds with the replacement.


shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1294, opened 2020-08-26
Deprecate org.embulk.spi.time.Timestamp

dmikurube:
This deprecation does not mean immediate removal of `Timestamp`, but as javadoc'ed for `Timestamp`, `java.time.Instant` can be a perfect alternative of `org.embulk.spi.time.Timestamp`, and we are going to use `Instant` instead of `Timestamp` almost everywhere.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/PageReader.java
+++ b/embulk-core/src/main/java/org/embulk/spi/PageReader.java
@@ -1,5 +1,6 @@
 package org.embulk.spi;
 
+import java.time.Instant;
 import org.embulk.deps.buffer.Slice;
 import org.embulk.spi.time.Timestamp;
 import org.msgpack.value.Value;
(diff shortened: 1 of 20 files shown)


embulk-core/src/main/java/org/embulk/spi/PageReader.java
```diff
@@ -97,19 +98,42 @@ public String getString(int columnIndex) {
...
+     *
+     * @deprecated Use {@link #getTimestampInstant(int)} instead.
+     */
+    @Deprecated
+    @SuppressWarnings("deprecation")  // https://github.com/embulk/embulk/issues/1292
     public Timestamp getTimestamp(int columnIndex) {
+        return Timestamp.ofInstant(this.getTimestampInstant(columnIndex));
+    }
+
+    public Instant getTimestampInstant(final Column column) {
```
himadieievsv, line 124: Probably just `getInstant` name would be descriptive enough. What do you think?

embulk-core/src/main/java/org/embulk/spi/PageReader.java
```diff
@@ -97,19 +98,42 @@ public String getString(int columnIndex) {
...
+        // TODO check type?
+        return this.getTimestampInstant(column.getIndex());
+    }
+
+    public Instant getTimestampInstant(final int columnIndex) {
```
himadieievsv, line 129: The same here

embulk-core/src/main/java/org/embulk/spi/PageReader.java
dmikurube: Yeah, I also thought about that, but its naming convention has used Embulk's type names, not Java type names here. (See `getJson` while its Java type is `org.msgpack.value.Value`.)

That was the reason why I kept the word `Timestamp` here, but I understand `getInstant` is simpler.

@trung-huynh Can we have your idea?

himadieievsv: Or we can have just `getTimestamp `  if following Embulk's type names

dmikurube: Actually cannot because there already is `getTimestamp` to return `org.embulk.spi.time.Timestamp`, and that cannot be removed/replaced immediately because many existing plugins are calling that.

himadieievsv (approved):
LGTM

dmikurube:
Thanks! Merging.

PR #1295, opened 2020-08-26
Move Type and related classes, including Column and Schema, to embulk-api

dmikurube:
Moving `Type`, `Column`, `Schema`, and related classes to `embulk-api`.

Unlike other "APIs", some of them include implementations. As of now, I think we accept that because :
* `Type` and related classes are considered as a kind of constants in Embulk API.
* Implementations of `Column` and `Schema` should not be variated. Their implementations should be only ones.
    * (I thought to mark them `final`. We may actually do that later in another PR.)


himadieievsv (approved):
LGTM

dmikurube:
Thanks!

PR #1296, opened 2020-08-26
Suppress internal deprecation warnings for org.embulk.spi.time.Timestamp

dmikurube:
Follow-up to #1294.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/Exec.java
+++ b/embulk-core/src/main/java/org/embulk/spi/Exec.java
@@ -9,7 +9,6 @@ import org.embulk.config.ModelManager;
 import org.embulk.config.TaskReport;
 import org.embulk.config.TaskSource;
 import org.embulk.plugin.PluginType;
-import org.embulk.spi.time.Timestamp;
 import org.slf4j.Logger;
 
 public class Exec {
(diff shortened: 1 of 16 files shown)


embulk-core/src/main/java/org/embulk/spi/Exec.java
```diff
@@ -41,8 +40,9 @@ public static Injector getInjector() {
     }
 
     @Deprecated
-    public static Timestamp getTransactionTime() {
-        return Timestamp.ofInstant(session().getTransactionTimeInstant());
+    @SuppressWarnings("deprecation")  // https://github.com/embulk/embulk/issues/1292
+    public static org.embulk.spi.time.Timestamp getTransactionTime() {
```
himadieievsv, line 44: Just out of curiosity, why we removed  `org.embulk.spi.time.Timestamp` from import section and use it in code?

dmikurube: That's because the class `org.embulk.spi.time.Timestamp` is deprecated by itself. If an entire class is deprecated, `import ...;` is warned for deprecation, and `@SuppressWarnings("deprecation")` does not work for `import`.

trung-huynh (approved):

himadieievsv (approved):
lgtm

dmikurube:
Thanks!

PR #1300, opened 2020-08-27
Replace TimestampFormatter and TimestampParser with hidden embulk-util-timestamp

dmikurube:
It's a big PR. It's to replace Embulk's `TimestampFormatter` and `TimestampParser` with `embulk-util-timestamp`'s. By replacing, we can remove a lot of implementations for Ruby-compatible date/time from `embulk-core`, and some heavy tests with JRuby.

They will be also removed eventually as all plugins should finally use `embulk-util-timestamp` by themselves. They remain here just to call `embulk-util-timestamp` for plugin compatibility.

JFYI, `embulk-util-timestamp` is loaded behind `DependencyClassLoader` on `embulk-core`. So, they won't conflict with plugin's `embulk-util-timestamp`.


embulk-core/src/test/java/org/embulk/spi/time/TestRubyTimeFormat.java
```diff
@@ -1,372 +0,0 @@
-package org.embulk.spi.time;
```
dmikurube, line 1: The same tests in https://github.com/embulk/embulk-util-rubytime/blob/v0.3.2/src/test/java/org/embulk/util/rubytime/TestFormat.java

embulk-core/src/test/java/org/embulk/spi/time/TestTimeZoneIds.java
```diff
@@ -1,18 +0,0 @@
-package org.embulk.spi.time;
```
dmikurube, line 1: Tested in https://github.com/embulk/embulk-util-timestamp/blob/v0.2.1/src/test/java/org/embulk/util/timestamp/TestLegacyDateTimeZones.java

embulk-core/src/test/java/org/embulk/spi/time/TestTimestampParser.java
```diff
@@ -493,20 +493,20 @@ public void testRuby__strptime__fail() {
...
-        testRubyToParse("7", "%u", 0L);
-        testRubyToParse("0", "%w", 0L);
+        failRubyToParse("7", "%u");  // "ruby:" version fails to parse this.
+        failRubyToParse("0", "%w");  // "ruby:" version fails to parse this.
         failRubyToParse("7", "%w");
 
         failRubyToParse("Sanday", "%A");
         failRubyToParse("Jenuary", "%B");
-        testRubyToParse("Sundai", "%A", 0L);
+        failRubyToParse("Sundai", "%A");  // "ruby:" version fails to parse this.
```
dmikurube, line 509: Those are known, and intentional.

embulk-core/src/test/ruby/monkey_strptime/mri/date/test_date_strptime.rb
```diff
@@ -1,515 +0,0 @@
-# frozen_string_literal: false
```
dmikurube, line 1: Tested in https://github.com/embulk/embulk-util-rubytime/blob/v0.3.2/src/jrubyTest/resources/ruby/test/date/test_date_strptime.rb

embulk-core/src/main/java/org/embulk/spi/time/TimestampFormatter.java
```diff
@@ -139,30 +95,35 @@ public static TimestampFormatter of(final Task task, final Optional<? extends Ti
...
+    @Deprecated  // This method will be removed sooner when Joda-Time is removed from Embulk during v0.10.
     public org.joda.time.DateTimeZone getTimeZone() {
-        if (this.delegate == null) {
-            throw new RuntimeException("FATAL: Unexpected execution path of TimestampFormatter without delegate.");
-        }
-        return this.delegate.getTimeZone();
+        return TimeZoneIds.parseJodaDateTimeZone(this.zoneIdString);
     }
 
+    @SuppressWarnings("deprecation")  // https://github.com/embulk/embulk/issues/1292
```
himadieievsv, line 103: looks like suppression like that will have no effect because class `TimestampFormatter` is also market as deprecated.

embulk-core/src/main/java/org/embulk/spi/time/TimestampFormatter.java
```diff
@@ -139,30 +95,35 @@ public static TimestampFormatter of(final Task task, final Optional<? extends Ti
...
-        }
-        return this.delegate.format(value);
+        // It may throw DateTimeException.
+        return this.delegate.format(value.getInstant());
     }
 
     // Receiving LineEncoder as a parameter is deprecated. TimestampFormatter should have fewer dependencies inside.
-    // It won't be removed very soon at least until Embulk v0.10.
     @Deprecated
+    @SuppressWarnings("deprecation")  // https://github.com/embulk/embulk/issues/1292
```
himadieievsv, line 111: The same here, since method is deprecated we can omit  `@SuppressWarnings("deprecation")`

himadieievsv (approved):
lgtm, with just few notices

shroman (approved):
lgtm

embulk-core/src/main/java/org/embulk/spi/time/TimestampFormatter.java
dmikurube: Actualy `@SuppressWarnings("deprecation")` warned. Hmm, there may be some difference in building environment? Let me have another look later.

embulk-core/src/main/java/org/embulk/spi/time/TimestampFormatter.java
dmikurube: ditto.

dmikurube (commented):
Thanks!

PR #1303, opened 2020-08-31
Suppress internal deprecation warnings for TimestampFormatter and TimestampParser

dmikurube:
We have had a lot of deprecation warnings when building `embulk-core` since #1273. They have been deprecated for plugins, but the core still uses them for a while.

To avoid so noisy warnings, adding `@SuppressWarnings` there.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/util/DynamicColumnSetterFactory.java
+++ b/embulk-core/src/main/java/org/embulk/spi/util/DynamicColumnSetterFactory.java
@@ -4,8 +4,6 @@ import org.embulk.config.ConfigException;
 import org.embulk.config.ConfigSource;
 import org.embulk.spi.Column;
 import org.embulk.spi.PageBuilder;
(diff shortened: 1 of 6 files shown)


himadieievsv (approved):
LGTM

shroman (approved):
lgtm

I wonder why you switched to fully qualified names for `org.embulk.spi.time` classes.

dmikurube:
@shroman That's the same reason with https://github.com/embulk/embulk/pull/1296#discussion_r480017466 -- because those entire classes are `@Deprecated`.

When an entire class is `@Deprecated`, just `import` causes deprecation warning, and `@SuppressWarnings("deprecation")` does not work for `import`.

Those classes would be finally removed, but it would take some more time in `embulk-core`. We needed to notify plugin developers about those deprecations earlier. That is the reason why I hurried to deprecate them while the `embulk-core` internal needed a lot of `@SuppressWarnings` as not ready yet.

dmikurube:
Thanks!

PR #1307, opened 2020-08-31
Suppress internal deprecation warnings for org.embulk.spi.util.Pages

dmikurube:
We have had a lot of deprecation warnings when building `embulk-core` since #1273. They have been deprecated for plugins, but the core still uses them for a while.

To avoid so noisy warnings, adding `@SuppressWarnings` there.


himadieievsv (approved):
LGTM

trung-huynh (approved):

shroman (approved):
lgtm

PR #1308, opened 2020-08-31
Deprecate ModelManager

dmikurube:
`ModelManager` is one of the biggest source of complicated dependencies in `embulk-core`. Once plugins start using `embulk-util-config` instead of `ConfigSource#loadConfig` and `TaskSource#loadTask`, it will finally be unnecessary.

To show it will be finally away, marking it `@Deprecated`. Plugins should not use it!


himadieievsv (approved):
LGTM

shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1310, opened 2020-09-02
Relax the restriction in building Column from JSON without index

dmikurube:
`Column` has been deserialized through `ColumnJacksonModule` since #1290, but I found that `ColumnJacksonModule` is more strict than `@JsonCreator` -- `index` was able to be empty in `@JsonCreator`.

Although we've observed no problems with it, relaxing it back just in case.


trung-huynh (approved):
LGTM

shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1312, opened 2020-09-02
Prepare org.embulk.spi.ExecInternal to contain Exec methods not for plugins

dmikurube:
It is planned to move `org.embulk.spi.Exec` to `embulk-api`, but some methods should be removed before that because :
* `getInjector`: `Injector` would be hidden from plugins.
* `getModelManager`: `ModelManager` is deprecated.
* `newPlugin`: not for plugins, and its argument `PluginType` blocks from hiding Jackson from plugins.

But, they are still needed in `embulk-core`, and some (hacky) plugins are using `newPlugin`.

Then, this commit introduces `ExecInternal` for `embulk-core` inside, and hacky plugins so that they can technically call it (although calling it is discouraged -- no guarantee in compatibility).


shroman (approved):
lgtm

dmikurube:
Thanks! I'm going to release v0.10.15 -- it could be a very experimental version... ;)

PR #1318, opened 2020-09-14
Move TempFileSpace to embulk-api

dmikurube:
`TempFileSpace` is provided for plugins through `Exec.getTempFileSpace`, so `TempFileSpace` should be in `embulk-api`. But it now includes actual implementation.

This PR :
1. Splits `TempFileSpace` to interface and implementation. (The constructor and the creator method are no longer available.)
2. Moves `TempFileSpace` (interface) to `embulk-api`.


shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1320, opened 2020-09-14
Move PageBuilder to embulk-api

dmikurube:
`PageBuidler` is now considered as a part of Embulk API. Although calling its constructor has been the only way to construct `PageBuilder` from plugins, `PageBuilder` should no longer provide its own constructor for plugins in `embulk-api`. Instead, `Exec` is going to have `Exec.getPageBuilder`.

* v0.10: `PageBuilder`'s constructor is marked as `@Deprecated`, and `Exec.getPageBuilder` is added. But Embulk v0.9-compatible plugins still have to call the constructor because Embulk v0.9 has only the constructor, no `Exec.getPageBuilder`.
* v0.11: `PageBuilder`'s constructor will still be kept, but plugins should start using `Exec.getPageBuilder` instead.
* v1.0: `PageBuilder`'s constructor would be removed at some early point of v1.0.


himadieievsv (approved):
LGTM

dmikurube:
Thanks!

PR #1322, opened 2020-09-14
Move PageReader to embulk-api

dmikurube:
`PageReader` is now considered as a part of Embulk API. Although calling its constructor has been the only way to construct `PageReader` from plugins, `PageReader` should no longer provide its own constructor for plugins in `embulk-api`. Instead, `Exec` is going to have `Exec.getPageReader`.

* v0.10: `PageReader`'s constructor is marked as `@Deprecated`, and `Exec.getPageReader` is added. But Embulk v0.9-compatible plugins still have to call the constructor because Embulk v0.9 has only the constructor, no `Exec.getPageReader`.
* v0.11: `PageReader`'s constructor will still be kept, but plugins should start using `Exec.getPageReader` instead.
* v1.0: `PageReader`'s constructor would be removed at some early point of v1.0.


trung-huynh (approved):
LGTM

shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1327, opened 2020-09-23
Move Exec and ExecSession to embulk-api

dmikurube:
Finally! It should be the last API for plugins. I believe that `embulk-api` and `embulk-spi` should be ready for plugins to build only with depending on them without `embulk-core`. Some more supplemental classes in `embulk-api` and utility libraries (`embulk-util-*`) may be needed later, though.


shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1332, opened 2020-10-13
Upgrade depended slf4j-api to 1.7.30

dmikurube:
Upgrading `slf4j-api` to the latest `1.7`. It is required to use `SubstituteLogger` in next PRs #1333 and #1334.

We don't use `slf4j-api` `1.8.*` nor `2.0.*` because :
* SLF4J 1.8 will never be "released".
  * http://mailman.qos.ch/pipermail/slf4j-user/2020-August/001748.html
* SLF4J 2.0 has a different mechanism to find a logging driver. It can cause severe incompatibility problems with plugins.

### DIFF
--- a/embulk-api/build.gradle
+++ b/embulk-api/build.gradle
@@ -11,5 +11,5 @@ configurations {
 
 dependencies {
     api "org.msgpack:msgpack-core:0.8.11"
-    api "org.slf4j:slf4j-api:1.7.12"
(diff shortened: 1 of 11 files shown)


shroman (approved):
wow, it's like 4 year jump from `1.7.12` :)
LGTM

himadieievsv (approved):
LGTM

dmikurube:
Thanks!

PR #1334, opened 2020-10-14
Find Embulk's home-like directories: embulk_home, m2_repo, gem_home, and gem_path

dmikurube:
It is to set Embulk System Properties not only from the command line, but also from environment variables and a config file (`embulk.properties`). It also finds a kind of "home" directories from the file system, and the Embulk System Properties.

JFYI: it basically does not affect `EmbulkEmbed` use-case. It's only about CLI use-case. The callers of `EmbulkEmbed` needs to configure Embulk System Properties `gem_home`, `gem_path`, and `m2_repo` explicitly.

Before:
* Embulk had no concept of "Embulk home".
* Embulk was looking for "Gem home", "Gem path", and "local Maven repository" by an ad-hoc rule.
    * Env `GEM_HOME` and `GEM_PATH`
    * Fixed directories under `~/.embulk/lib`

After:
* Embulk will have a concept of "Embulk home".
    * If Embulk System Property `embulk_home` is set from the command line or environment variables, adopt it.
    * If a directory which is named `.embulk` which has `embulk.properties` directly under itself is found in parent directories of the current working directory, assume the directory is "Embulk home".
    * If neither of them are found, use the fixed directory `~/.embulk`.
* Embulk will load `embulk.properties` as Embulk System Properties if exists.
* Embulk will initialize Embulk System Properties `gem_home`, `gem_path`, and `m2_repo` unconditionally.
    * If those Properties are set explicitly (from command line, environment variables, `embulk.properties`), adopt them.
    * If they are not set, use `<embulk_home>/lib/...`.

Then, the behavior is compatible if `.embulk/embulk.properties` does not exist in parent directories of the current working directory.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -287,8 +287,8 @@ dependencies {
     implementation project(':embulk-standards')
 
     // Logback and jansi are included only in the executable package. (jansi for logback colors to work on Windows.)
-    implementation 'ch.qos.logback:logback-classic:1.1.3'
-    implementation 'org.fusesource.jansi:jansi:1.11'
+    implementation "ch.qos.logback:logback-classic:1.2.3"
+    implementation "org.fusesource.jansi:jansi:1.18"
 
     embed(project(':embulk-deps')) {
         exclude group: 'org.apache.commons', module: 'commons-lang3'  // Included in embulk-core.
--- a/embulk-core/src/main/java/org/embulk/cli/EmbulkRun.java
+++ b/embulk-core/src/main/java/org/embulk/cli/EmbulkRun.java
@@ -3,7 +3,6 @@ package org.embulk.cli;
 import java.io.IOException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
-import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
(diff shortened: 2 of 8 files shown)


dmikurube:
@shroman @himadieievsv @trung-huynh Ah, I noticed that I've pushed a half-baked files on this PR. I'll fix it -- please wait for a while.

dmikurube:
@shroman @himadieievsv @trung-huynh Updated the PR. It's now ready.

shroman:
@dmikurube At a glance, I can't find where `embulk_home` is being handled in this PR. Is the description correct?

dmikurube:
`findEmbulkHome()` in `EmbulkSystemPropertiesBuilder.java` tries to find it from command-line, environment variables, and the file system.
https://github.com/embulk/embulk/pull/1334/files#diff-e15319229fbc216e7f78228e441af34200b30eab7e71f62daa688a428b886ffbR54

Then, everything is searched for based on the `embulkHome` variable.
https://github.com/embulk/embulk/pull/1334/files#diff-e15319229fbc216e7f78228e441af34200b30eab7e71f62daa688a428b886ffbR57

And finally it's (re)set as an `embulk_home` Embulk System Property.
https://github.com/embulk/embulk/pull/1334/files#diff-e15319229fbc216e7f78228e441af34200b30eab7e71f62daa688a428b886ffbR78

Do they answer your question?

The `embulk_home` property is not used in later processing as of now, actually.

shroman:
@dmikurube sorry, I didn't notice github didn't expand that file and couldn't search -- need to get PR locally :)

dmikurube:
Ah, yeah, GitHub sometimes automatically collapse some long files... ;)

trung-huynh:
@dmikurube , just went though the whole PR, is the description correct:
> Embulk will initialize Embulk System Properties gem_home, gem_path, and m2_repo from "Embulk home`.
If those Properties are set explicitly, adopt them.
If they are not set, use <embulk_home>/lib/....

As `gem_home`, `gem_path` and `m2_repo` with command line props are still prioritized than from `Embulk Home`.

dmikurube:
@trung-huynh Ah, that was not accurate, thanks!

I revised the PR description to:

> * Embulk will initialize Embulk System Properties `gem_home`, `gem_path`, and `m2_repo` unconditionally.
>    * If those Properties are set explicitly (from command line, environment variables, `embulk.properties`), adopt them.
>    * If they are not set, use `<embulk_home>/lib/...`.

trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1337, opened 2020-10-21
Initialize embedded JAR resources in the very beginning, without SLF4J

dmikurube:
Before upcoming changes to refactor command line processing, and Embulk system properties, changing the loader of JAR resources (in the command-line Embulk executable binary) to log just with `System.err`, not with SLF4J.

It is because :
* They are loaded (initialized) in the very beginning of Embulk CLI, when SLF4J and Logback are not initialized.
* We can use `SubstituteLogger` of SLF4J for delayed logging, but their all "logs" are about critical cases. When they are logged, Embulk is failing.
* For critical error messages, not SLF4J logging, but `System.err` would fit well.


shroman (approved):
LGTM

dmikurube:
Thanks!

PR #1338, opened 2020-10-26
Remove embedded gems: msgpack, bundler, liquid

dmikurube:
After removing JRuby from Embulk executable binary, it also removes embedded RubyGem libraries: `msgpack`, `bundler`, and `liquid`.

* `msgpack` is mandatory for Ruby-based Embulk plugins. Users would have to install it with `embulk gem install msgpack` by themselves.
* `bundler` has not been mandatory, but some CLI users have used it for maintaining plugin installation. Actually `bundler` is preinstalled in some JRuby versions, so it depends on users which JRuby to use.
* `liquid` has not been mandatory, but some CLI users have used it for templating Embulk configs. JFYI: its's not available for `EmbulkEmbed` users. Users would have to install it with `embulk gem install liquid` by themselves.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -279,6 +279,7 @@ repositories {
 
 configurations {
     embed
+    jruby
 }
 
 dependencies {
@@ -293,6 +294,9 @@ dependencies {
         exclude group: 'org.apache.commons', module: 'commons-lang3'  // Included in embulk-core.
         exclude group: 'com.google.guava', module: 'guava'  // Included in embulk-core.
     }
+
+    // Only to install the msgpack gem in embulk/build/.
+    jruby "org.jruby:jruby-complete:9.1.15.0"
 }
 
 def listEmbedDependencies = { rootModuleName, prefix ->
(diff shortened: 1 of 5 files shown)


shroman (approved):
lgtm

dmikurube:
I found that it's not working well, too. Taking a deeper look...

dmikurube:
Ugh, the problem was not just this PR, but my local msgpack gem was too new.

As written in the PR description,
> Users would have to install it with `embulk gem install msgpack` by themselves.

But as of now, Embulk expects `msgpack:1.1.0` although `embulk gem install msgpack` installs the latest `msgpack:1.3.3` automatically. When I installed the specific `msgpack:1.1.0`, I confirmed it worked.

Let me think deeper about it, but I believe we can go ahead with this PR.

dmikurube:
If a user installs `msgpack` just by `embulk gem install msgpack`, `msgpack:1.3.3` is installed as of 2020-11-17, and it does not match `~> 1.1.0`.
https://rubygems.org/gems/msgpack/versions/1.3.3-java

`~>` means that the first digit and the second digit are the same.
https://guides.rubygems.org/patterns/#declaring-dependencies

632ec5647b9d3b0702a792e2966b8fb6c41fcd22 changes the restriction from `~> 1.1.0` to `>= 1.1.0` to keep `embulk-core` work even with newer `msgpack`,

dmikurube:
The PR is ready for review again -- the difference is only 632ec5647b9d3b0702a792e2966b8fb6c41fcd22 .

shroman (approved):
Looks good, thanks for checking! Looking at the code only doesn't work sometimes :)

trung-huynh (approved):
LGTM

himadieievsv (approved):
LGTM

dmikurube:
Thanks!

PR #1341, opened 2020-10-27
Move PreviewPrinter to embulk-deps

dmikurube:
`*PreviewPrinter` needed to use Jackson in its mechanism. To move Jackson behind `DependencyClassLoader` eventually, they need to be moved behind `DependencyClassLoader` as well.


shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1342, opened 2020-11-18
Cleanup and refactor PluginClassLoader and its Factory

dmikurube:
`PluginClassLoader` had had a mechanism to load a JARs-in-JAR-style nested JAR plugin, but it has been removed since v0.9.17 by: #1111

This commit is to remove some remaining code for nested JAR plugins.


shroman (approved):
lgtm

PR #1356, opened 2020-12-09
Load a Maven-based plugin from a simple plugin type with Embulk system properties

dmikurube:
It is to look for a Maven-based plugin even when only a simple plugin type is specified like `type: ftp` in user configuration, if a corresponding Embulk system property is configured.

For example, assume a user configuration is like below:

```
in:
  type: ftp
  host: ...
  ...
```

Embulk has looked only for a RubyGem-based plugin so far in this case.

By this pull request, Embulk looks for a Maven-based plugin if an Embulk system property `plugins.input.ftp` is set like `maven:org.embulk:ftp:0.3.0`.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/plugin/MavenPluginType.java
+++ b/embulk-core/src/main/java/org/embulk/plugin/MavenPluginType.java
@@ -6,6 +6,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


shroman (approved):
lgtm

himadieievsv (approved):

dmikurube:
While v0.10.25 is not 100% confirmed yet, this should be okay. Thanks for reviewing it! Merging...

PR #1358, opened 2020-12-11
Reenable multiple types of self-contained JAR resource files

dmikurube:
Types of self-contained JAR resource files have been once disabled in #1279, but now we have to reenable it for a different purpose.

Embulk executable binary is going to self-contain `embulk-standards` plugin JARs in upcoming changes. Types of self-contained JAR files need to be reenabled so that the self-contained plugin JARs are loaded through `PluginClassLoader`s, not `DependencyClassLoader`.

This is needed for #1360.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/deps/EmbulkSelfContainedJarFiles.java
+++ b/embulk-core/src/main/java/org/embulk/deps/EmbulkSelfContainedJarFiles.java
@@ -24,22 +24,24 @@ public final class EmbulkSelfContainedJarFiles {
 
     public static final class StaticInitializer {
         private StaticInitializer() {
-            this.jarFileResources = new ArrayList<>();
+            this.jarFileResources = new HashMap<>();
         }
 
-        public StaticInitializer addJarFileResource(final String resource) {
-            this.jarFileResources.add(resource);
-            return this;
-        }
-
-        public StaticInitializer addJarFileResources(final Collection<String> resources) {
-            this.jarFileResources.addAll(resources);
+        public StaticInitializer addJarFileResources(final String category, final Collection<String> resources) {
+            this.jarFileResources.compute(category, (dummy, old) -> {
+                final ArrayList<String> targetList = (old != null) ? old : new ArrayList();
+                targetList.addAll(resources);
+                return targetList;
+            });
             return this;
         }
 
         public StaticInitializer addFromManifest(final Manifest manifest) {
             final Attributes attributes = manifest.getMainAttributes();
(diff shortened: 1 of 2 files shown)


embulk-core/src/main/java/org/embulk/deps/EmbulkSelfContainedJarFiles.java
```diff
@@ -128,5 +142,9 @@ static Resource getSingleResource(final String targetResourceName) {
         return list;
     }
 
-    private static final ArrayList<String> JAR_RESOURCE_NAMES = new ArrayList<>();
+    // Category for embulk-core's own self-contained dependencies.
+    public static final String CORE = "";
```
shroman, line 146: Safer to give it some real value to avoid passing empty strings by mistake in future?

embulk-core/src/main/java/org/embulk/deps/EmbulkSelfContainedJarFiles.java
```diff
@@ -86,9 +94,13 @@ static Resource getSingleResource(final String targetResourceName) {
         return resourceToReturn;
     }
 
-    static Collection<Resource> getMultipleResources(final String targetResourceName) {
+    static Collection<Resource> getMultipleResources(final String targetResourceName, final String category) {
+        if (category == null) {
+            return null;
```
shroman, line 99: Let's return an empty list here. Because in `DependencyClassLoader#findResource` we iterate over the returned value of this method.

embulk-core/src/main/java/org/embulk/deps/EmbulkSelfContainedJarFiles.java
dmikurube: Also a reasonable point. I at first chose an empty string so that it will never conflict with given categories accidentally (expected to be an embedded plugin name: such as `embulk-parser-csv` as of now (*1)), but yeah, mistakes can easily bring an empty string there.
(*1): https://github.com/embulk/embulk/pull/1360/files#diff-02ae5d423faebe1c8fd3e17f337912a9b5e29157d471a7603cd7053c2e587b72R79

In 33f9d8bf622a72b6762a0f6d736b91f32b67d9b2, set it as `"$embulk-deps$"` which explicitly seems like a special thing.

dmikurube (commented):
Thanks for your comments, @shroman. Made some changes for your suggestion. What do you think?

shroman (approved):
lgtm! thanks for addressing my comments!

trung-huynh (approved):
LGTM

himadieievsv (approved):

PR #1359, opened 2020-12-11
Refactor DependencyClassLoader to separate to SelfContainedJarAwareURLClassLoader

dmikurube:
It prepares for `PluginClassLoader` to be able to load from JAR resources self-contained in the Embulk JAR file. `PluginClassLoader` is going to extend `SelfContainedJarAwareURLClassLoader`, too.

This is needed for #1360.

PR #1360, opened 2020-12-11
Embed standard non-guess plugins as self-contained JARs, and load them through PluginClassLoader

dmikurube:
After #1358 and #1359, this PR starts loading (non-guess) `embulk-standards` plugins as self-contained JAR files through `PluginClassLoader`.

They have been loaded on the top-level class loader so far, registered through `StandardPluginModule` and `StandardPluginExtension`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -307,6 +307,10 @@ dependencies {
         transitive = false
     }
 
+    subprojectNamesOfStandardPlugins.each { pluginName ->
+        embed(project(":${pluginName}"))
+    }
+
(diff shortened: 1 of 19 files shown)


himadieievsv (approved):

dmikurube:
In bb448d72d16fae225ca9721daceb9ee42a1b6009, changed an order a bit in `/build.gralde` so that it won't conflict with `master`.

shroman (approved):
lgtm

PR #1361, opened 2020-12-11
Stop registering standard guess plugins in StandardPluginModule, instead from the JAR Manifest via Embulk system property "default_guess_plugins"

dmikurube:
This PR stops registering guess plugins in `StandardPluginModule` and `StandardPluginExtension`, then removes those two classes.

The guess plugins are now registered via Embulk system property `default_guess_plugins`, which is set initially from the JAR Manifest.


trung-huynh (approved):
LGTM

embulk-core/src/main/java/org/embulk/exec/GuessExecutor.java
```diff
@@ -56,8 +50,7 @@
     }
 
     public static void registerDefaultGuessPluginTo(Binder binder, PluginType type) {
-        Multibinder<PluginType> multibinder = Multibinder.newSetBinder(binder, PluginType.class, ForGuess.class);
-        multibinder.addBinding().toInstance(type);
+        throw new UnsupportedOperationException("GuessExecutor.registerDefaultGuessPluginTo no longer works.");
```
shroman, line 53: extreme nit :) `no longer works` sounds like it's broken, more accurate wording could be better.

shroman (approved):
lgtm

embulk-core/src/main/java/org/embulk/exec/GuessExecutor.java
dmikurube: Thanks! Agreed. Changed the wording to `is no longer available`.

dmikurube:
Thanks!

PR #1363, opened 2020-12-15
Modernize embulk-standards plugins with v0.11-ready API/SPI

dmikurube:


dmikurube:
At first thought to finish this modernization first before converting guess plugins to Java, but actually, Java conversion should come first because :
* Modernization includes migrating to `embulk-util-config`, not `ConfigSource#loadConfig`.
* CSV guess includes a call of `ConfigSource#load_config`.
    * https://github.com/embulk/embulk/blob/v0.10.23/embulk-standards/src/main/ruby/embulk/guess/csv.rb
* Ruby-based plugins cannot use `embulk-util-config` (and other Java libraries) without contaminating the top-level class loader.

dmikurube:
This PR is split into 20 PRs #1381 - #1400 for smaller units to review. Closing this PR.

PR #1366, opened 2021-01-07
Reimplement standard guess plugins in Java, except for csv

dmikurube:
"Standard" guess plugins are implemented in (J)Ruby in `embulk-standards` (`embulk.gem` since v0.10.24). In that Ruby code, the CSV guess calls `CSVParserPlugin` to perform a guess about CSV.

This dependency is blocking `CSVParserPlugin` from being "modernized" for Embulk v0.10 to use `embulk-util-config` instead of `embulk-core`'s `org.embulk.config.*` classes. To unblock this, we are going to reimplement standard guess plugins in Java from Ruby.

This pull request is the first step before CSV -- reimplementing all other bzip2, gzip, and JSON guess, which are not very complicated. (On the other hand, CSV guess is complicated, to be done in another pull request.)


shroman (approved):
lgtm
Compared with the Ruby part (I hope I got it right:) )

trung-huynh (approved):
LGTM, compared with Ruby counter-part

dmikurube:
Thanks!

PR #1367, opened 2021-01-18
Release Embulk's Ruby code as a separate RubyGem

dmikurube:
After v0.10.21's JRuby separation to a sub class loader, It turned out that JRuby seems sometimes (randomly) failing to find `.rb` files embedded in `embulk-core-*.jar` as Java resources. The root cause is not 100% confirmed, but it looks like a problem in some versions of JRuby. It may happen when looking for `.rb` files in a different JAR file from JRuby's, or when looking for `.rb` files from a different class loader.

It separates Embulk's Ruby code as a separate Ruby gem to be released in https://rubygems.org/ to workaround the problem, and also to have cleaner JRuby separation in Embulk packages.

Users need to install `embulk` gem by themselves when running a RubyGem-packaged plugin. But, users have already needed to install `msgpack` gem. It is not a big difference.

----

As a side effect, build files have been much cleaned up around Ruby and Ruby tests. :tada:


shroman (approved):
lgtm

trung-huynh (approved):
LGTM

dmikurube:
Thanks! Merging.

PR #1370, opened 2021-01-20
Fail explicitly with embulk.gem mismatching with the running Embulk

dmikurube:
It turned out that existing v0.9.* users with Bundler (`mkbundle`) may fall into a weird situation that `embulk.gem` 0.10.24 is installed by Bundler, caused by releasing `embulk.gem` 0.10.24 in rubygems.org. 

Most Bundler users should have had the following declaration in their `Gemfile` (by our recommendation when we started v0.9):
```
gem 'embulk'
```

It installs the latest `embulk.gem`, then `embulk.gem` 0.10.24 is installed.

However, we cannot get back. Then, we're going to introduce a version check in `embulk.gem` v0.10.25+ between the Embulk core (Java) and `embulk.gem` (Ruby). The check fails if the running Embulk version does not match the loaded `embulk.gem` version, and suggests how the user should do based on their use-case.

In case the user uses v0.9.* with Bundler, it suggests to update their `Gemfile` as below, and it should work:
```
gem 'embulk', '< 0.10'
```

----

We found the case thanks to @hiroyuki-sato.
https://gist.github.com/hiroyuki-sato/e3f277a05656888ef3e4810de799a946

### DIFF
--- a/embulk-core/src/main/java/org/embulk/jruby/JRubyInitializer.java
+++ b/embulk-core/src/main/java/org/embulk/jruby/JRubyInitializer.java
@@ -129,6 +129,10 @@ final class JRubyInitializer {
             }
 
             if (this.isEmbulkSpecific) {
+                // Loading 'embulk/version' triggers a version check with embulk.gem v0.10.25 or later.
+                // (Even in Embulk v0.10.24 or earlier, 'embulk/version' is loaded soon below from 'embulk/logger'.)
+                jruby.runScriptlet("require 'embulk/version'");
+
                 // Embulk's base Ruby code is loaded at last.
                 jruby.runScriptlet("require 'embulk/logger'");
                 jruby.runScriptlet("require 'embulk/java/bootstrap'");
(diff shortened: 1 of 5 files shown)


hiroyuki-sato:
LGTM 👍

| embulk  | gem       | command  | expected | result |
|---------|-----------|----------|----------|--------| 
| 0.10.24 | 0.10.25*1 | guess    | error    | OK     |
| 0.10.24 | 0.10.25   | run*2    | error    | OK     |
| 0.10.25 | 0.10.25   | guess    | no error | OK     |
| 0.10.25 | 0.10.25   | run      | no error | OK     |

* *1: embulk 0.10.25 snapshot (561d3a2d82b5aea4d1be5be1a31fcb48d852aa32)
* *2: add the filter plugin ruby_proc

```
filters:
- type: ruby_proc
````

ref: https://gist.github.com/hiroyuki-sato/62f2745da71e7b63322411d7f0d488ab

trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1372, opened 2021-01-22
Test embulk.gem's version check

dmikurube:
It tests "embulk.gem's version check" mechanism introduced in #1370 in `embulk-ruby`'s Ruby test code.


shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1373, opened 2021-01-22
Improve the error message against embulk.gem's version mismatch

dmikurube:
The error message, which was introduced in #1370 to show Embulk's version and `embulk.gem`'s version don't match, was not appropriate. This PR improves the error message.


trung-huynh (approved):

dmikurube:
Thanks!

PR #1357, opened 2020-12-09
Split embulk-standards to independent subprojects as plugins

dmikurube:
As the first step to externalize `embulk-standards`as separate plugins, this PR splits it as separate Gradle subprojects. All are still depended from `embulk-standards`, so no difference in the final build yet.

PR #1377, opened 2021-01-28
Reimplement CSV guess plugins in Java as embulk-guess-csv and embulk-guess-csv_all_strings

dmikurube:
This is an separeted pull request only for the CSV guess from #1366 because the CSV guess is so much complicated unlike Bzip2, Gzip, and JSON.

Note that this reimplementation is still intentionally in the "old-style". For example, it does not use `embulk-util-config`. It still depends on `embulk-core`'s `org.embulk.spi.util` classes, which have already been externalized as `embulk-util-file` and `embulk-util-text`.

It may seem redundant, but it is needed to be synchronized with `embulk-parser-csv`. `embulk-parser-csv` was externalized "as-is" in v0.10.26 (#1357) also in the "old-style" so that we can easily confirm we have no difference in the code level. `embulk-parser-csv` will be "modernized" in next v0.10.28. This `embulk-guess-csv` will be modernized at the same time in v0.10.28.

----

The biggest redundancy here is that this pull request has (almost) a copy of `embulk-util-guess` in it. It is because `embulk-util-guess` expects `embulk-util-config`. They will be removed when this is modernized in next v0.10.28. Those files have comments explaining they will be removed soon.

* These files will be removed in next v0.10.28, replaced by `embulk-util-guess`.
  * `CharsetGuess.java`
  * `LineGuessHelper.java`
  * `NewlineGuess.java`
  * `TimeFormatGuess.java`
  * `timeformat/*.java`
* This will remain in next v0.10.28 because `embulk-guess-csv` depends on `SchemaGuess#typesFromListRecords` which should have been `private`. We'll keep it tracked, and refactor it in the future. (It has originally been an issue from the old Ruby-based implementation. There was a comment about it.)
  * `SchemaGuess.java`

----

Btw, through this reimplementation, I found that `embulk-util-guess` had some bugs. They will be fixed in next v0.1.1 by the following pull requests. The copy in this pull request is the one after those fixes.
* https://github.com/embulk/embulk-util-guess/pull/10
* https://github.com/embulk/embulk-util-guess/pull/11
* https://github.com/embulk/embulk-util-guess/pull/12

----

It keeps a copy of the original Ruby-based implementation `csv.rb` in `src/test/` so that we can have more tests by comparing with the original Ruby-based implementation. It does nothing so far, though.


dmikurube:
The redandant files are removed in #1399 in v0.10.28.

trung-huynh (approved):
LGTM, checked in both `embulk-util-guess` and Ruby counter-part, there are some structural changes but logically it is similar.

PR #1382, opened 2021-02-25
Prepare in the top-level build.gradle for modernizing standard plugins

dmikurube:
It adds the `org.embulk.embulk-plugins` Gradle plugin without applying at the top-level.

Then, it moves the Gradle subproject list `subprojectNamesOfStandardPlugins` to a new list `subprojectNamesOfLegacyStandardPlugins` so that a new `subprojectNamesOfStandardPlugins` can include modernized standard plugins.

Finally, it sets subprojects in the new `subprojectNamesOfStandardPlugins` to be released to Maven Central through `publishMavenPublicationToMavenCentralRepository` to de defined in their own sub build.gradle.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,7 @@ plugins {
     id "jacoco"
     id "com.jfrog.bintray" version "1.8.4" apply false
     id "org.ajoberstar.grgit" version "4.0.1"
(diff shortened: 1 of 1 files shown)


shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1388, opened 2021-02-25
Modernize embulk-filter-rename with v0.11-ready API/SPI

dmikurube:


embulk-filter-rename/src/main/java/org/embulk/filter/rename/RenameFilterPlugin.java
```diff
@@ -441,12 +469,27 @@ private Schema applyUniqueNumberSuffixRule(Schema inputSchema, UniqueNumberSuffi
...
     }
 
-    private static final ImmutableMap<String, String> CHARACTER_TYPE_KEYWORDS =
-            new ImmutableMap.Builder<String, String>()
-                    .put("a-z", "a-z")
-                    .put("A-Z", "A-Z")
-                    .put("0-9", "0-9")
-                    .build();
+    private static String repeatStrings(final String string, final int count) {
+        final StringBuffer buffer = new StringBuffer();
```
huylenq, line 473: Nit: StringBuilder would suffice?

huylenq (approved):
LGTM

embulk-filter-rename/src/main/java/org/embulk/filter/rename/RenameFilterPlugin.java
dmikurube: Umm, I don't remember why I used `StringBuffer`, but `StringBuilder` could be better. I'll have a revisit here later. Thanks!

dmikurube:
Thanks!

PR #1406, opened 2021-03-03
Remove Joda-Time completely from Embulk

dmikurube:
We will no longer have Joda-Time in Embulk in v0.11 - v1. Some (deprecated) methods are totally removed to remove Joda-Time as they include Joda-Time's class in their method signatures -- it's unavoidable.


trung-huynh (approved):
LGTM, finally we can remove Joda time.

dmikurube:
Thanks!

Issue #1409, opened 2021-03-22
Embulk 0.9.23 raise multiple exception if v0.10 api use.


hiroyuki-sato:
Issue Type: Bug Report
-----------------------

* Write the following environmental information.
  * OS version: macOS 11.2.3
  * Java version: 1.8.251
  * Embulk version: build(0.10.26), run: 0.9.23
  * Your Embulk configuration: See below
  * Plugin versions: See below

```
https://github.com/hiroyuki-sato/embulk-parser-jsonpath
cd embulk-parser-jsonpath/
git checkout -b topic/try-v0_9-v0_10 origin/topic/try-v0_9-v0_10
./gradlew gem
```

## Expected behavior

```
embulk-10.0.27 run -I build/gemContents example/conf.yml
```

### Case1: with one input file

example/conf.yml

```yaml
in:
  type: file
  path_prefix: example/input2 # <-- Modifiy HERE
  parser:
    type: jsonpath
    root: "$.results"
    default_timezone: "Asia/Tokyo"
    columns:
      - { name: "name",          type: string }
      - { name: "city",          type: string }
      - { name: "street_name",   type: string }
      - { name: "zip_code",      type: string }
      - { name: "registered_at", type: timestamp, format: "%Y-%m-%d %H:%M:%S" }
      - { name: "vegetarian",    type: boolean }
      - { name: "age",           type: long }
      - { name: "ratio",         type: double }
out:
  type: stdout
```

```
embulk run -I build/gemContents/lib example/conf.yml
```

```
Error: java.lang.RuntimeException: java.lang.NoSuchMethodError: org.embulk.spi.Buffer: method <init>()V not found
```

### Case2: with trhee input files and "null" output

copy file

```
cp example/input2.json example/input3.json
```

```yaml
in:
  type: file
  path_prefix: example/input
  parser:
    type: jsonpath
    root: "$.results"
    default_timezone: "Asia/Tokyo"
    columns:
      - { name: "name",          type: string }
      - { name: "city",          type: string }
      - { name: "street_name",   type: string }
      - { name: "zip_code",      type: string }
      - { name: "registered_at", type: timestamp, format: "%Y-%m-%d %H:%M:%S" }
      - { name: "vegetarian",    type: boolean }
      - { name: "age",           type: long }
      - { name: "ratio",         type: double }
out:
  type: "null" # <-- MODIFY HERE stdout -> null
```

embulk run -I build/gemContents/lib example/conf.yml

```
[...]

PR #1410, opened 2021-03-23
Declare Buffer's constructor compatible with v0.9's

dmikurube:
#1409 needs a fix in `embulk-util-file` to construct its `EmptyBuffer`.

`embulk-util-file:0.1.1`'s `EmptyBuffer` expects a constructor `Buffer()` without arguments.
* https://github.com/embulk/embulk-util-file/blob/v0.1.1/src/main/java/org/embulk/util/file/EmptyBuffer.java

But it's only in Embulk v0.10's `Buffer`.
* https://github.com/embulk/embulk/blob/v0.10.27/embulk-api/src/main/java/org/embulk/spi/Buffer.java

Embulk v0.9.23 did not have `Buffer()`, but `Buffer(byte[], int, int)`.
* https://github.com/embulk/embulk/blob/v0.9.23/embulk-core/src/main/java/org/embulk/spi/Buffer.java

It's the reason why `embulk-util-file` cannot initialize `EmptyBuffer` with Embulk v0.9.23 in #1409. This is one of the fixes for the problem. The other fix will be in `embulk-util-file`.

### DIFF
--- a/embulk-api/src/main/java/org/embulk/spi/Buffer.java
+++ b/embulk-api/src/main/java/org/embulk/spi/Buffer.java
@@ -38,6 +38,11 @@ public abstract class Buffer {
     protected Buffer() {
     }
 
+    @Deprecated  // Declared only for embulk-util-file can refer this.
+    protected Buffer(final byte[] wrap, final int offset, final int capacity) {
+        this();
+    }
+
     /**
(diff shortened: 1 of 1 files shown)


shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1414, opened 2021-03-24
Set guess_sample_buffer_bytes and preview_sample_buffer_bytes from Embulk System Properties

dmikurube:
`guess_sample_buffer_bytes` and `preview_sample_buffer_bytes` are 32KB by default, and they could be too small as of the age of 2020s.

But, a decision to increase the default values is not easy. For instance, if users are reading just a local file, incrasing the default value does ot have any serious impact. But for another example, some users read files from S3. AWS charges for a size of S3 read. Updating the default value can affect such users silently.

As a better workaround than setting `exec: guess_sample_buffer_bytes: 32768` in every Embulk configuration, this change is to make it available to set their default values from Embulk System Properties, which could be configured from auto-loaded `.embulk/embulk.properties` (since v0.10.21).

### DIFF
--- a/embulk-core/src/main/java/org/embulk/exec/GuessExecutor.java
+++ b/embulk-core/src/main/java/org/embulk/exec/GuessExecutor.java
@@ -149,7 +149,6 @@ public class GuessExecutor {
         guessPlugins.removeAll(task.getExcludeGuessPlugins());
         final int guessParserSampleBufferBytes =
                 task.getSampleBufferBytes().orElse(this.systemGuessSampleBufferBytes.orElse(DEAULT_SAMPLE_BUFFER_BYTES));
-        System.out.println(guessParserSampleBufferBytes);
 
         return guessParserConfig(sample, inputConfig, guessPlugins, guessParserSampleBufferBytes);
     }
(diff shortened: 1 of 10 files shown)


embulk-core/src/main/java/org/embulk/exec/GuessExecutor.java
```diff
@@ -135,7 +147,9 @@ public ConfigDiff guessParserConfig(Buffer sample, ConfigSource inputConfig, Con
         final GuessExecutorTask task = loadGuessExecutorTask(execConfig);
         guessPlugins.addAll(task.getGuessPlugins());
         guessPlugins.removeAll(task.getExcludeGuessPlugins());
-        final int guessParserSampleBufferBytes = task.getSampleBufferBytes();
+        final int guessParserSampleBufferBytes =
+                task.getSampleBufferBytes().orElse(this.systemGuessSampleBufferBytes.orElse(DEAULT_SAMPLE_BUFFER_BYTES));
+        System.out.println(guessParserSampleBufferBytes);
```
shroman, line 152: nit: not descriptive

dmikurube: Ah, really. It was just remaining when I was in tries-and-errors. Thanks for the good catch! Removed it in 67d7bc22c998faae5b9c05865591883ad05f1470.

shroman (approved):
lgtm

dmikurube:
Thanks!

PR #1415, opened 2021-03-25
Add Ruby-based embulk-guess-csv_verify plugin to compare the old Ruby-based CSV guess v.s. the new Java-based CSV guess

dmikurube:
This is an "experimental" Ruby-based plugin to compare the old Ruby-based CSV guess v.s. the new Java-reimplemented CSV guess. This as a guess plugin returns the result from the old Ruby-based implementation as its own restult.

With Embulk v0.10.29+ (to be released with this), users should be able to check whether they return the same guess results by a partial configuration like below:

```
exec:
  exclude_guess_plugins: ['csv']
  guess_plugins: ['csv_verify']
in:
  type: file
  path_prefix: "..."
```


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1419, opened 2021-04-13
Dump actual diffs in "decoders" and "parser" in embulk-guess-csv_verify

dmikurube:
`embulk-guess-csv-verify` (#1415) originally dumped only different key names, without actual diffs, in order to avoid logging confidential information (e.g. AWS access token) unintentionally.

But, actual diffs are really helpful to investigate. This PR is to dump actual diffs of `decoders` and `parser` sections because those sections should not have any confidential information.

### DIFF
--- a/embulk-guess-csv_verify/build.gradle
+++ b/embulk-guess-csv_verify/build.gradle
@@ -7,7 +7,7 @@ repositories {
 
 group = "${rootProject.group}"
 version = "${rootProject.version}"
-description = "Verification-purpose Embulk CSV guess plugin to compare the old Ruby-based one and the new Java-based one (not for your production use)"
+description = "Verification-purpose Embulk CSV guess plugin to compare the old Ruby-based one and the new Java-based one (not for your production use; note that 'decoders' and 'parser' sections in your configuration can be logged even if they contain confidential information)"
 
 dependencies {
     compile project(":embulk-guess-csv")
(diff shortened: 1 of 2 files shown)


shroman:
> should not have any confidential information

Does it mean they CANNOT have such information, or is it still possible?

dmikurube:
Thanks for your comment. 

> Does it mean they CANNOT have such information, or is it still possible?
At least for the default guess, decoders, and parsers we have -- `bzip2,` `gzip,` `json` and `csv`, they cannot have. If a user adds their own special ones in their configuration, it can have.

But, I don't think It's a problem because :
* This `embulk-guess-csv_verify` is really a special plugin only for verifycation purpose (not to be bundled with the Embulk executable package). OSS Embulk users should use it at their own risk.
* It does not sound very natural to use confidential information in "decoding" and "parsing", not about accessing to a storage like S3, SFTP, ...
* (Just for ourselves) we plan to use it in our own service, but we only have known guess, decoders, and parsers. They have no concerns about confidential information.

shroman:
Thank you, it makes sense to me. I'd include a comment "Make sure no credentials are included" anyway, but I don't think it's a blocker.

shroman (approved):

dmikurube:
Thanks! Reasonable suggestion.

I pondered where to comment, but decided to note that in the RubyGem's "description" part. (To be available in https://rubygems.org/gems/embulk-guess-csv_verify)

shroman:
@dmikurube I'd probably add it to dumping messages.

PR #1420, opened 2021-04-14
Let embulk-guess-csv_verify to use the legacy CsvParserPlugin in the legacy embulk-standards

dmikurube:
`embulk-guess-csv_verify` has used the modern `embulk-parser-csv` internally, but it could be different from the original if the modern parser was not 100% compatible. To have a compliant comparison, using the legacy `CsvParserPlugin` copied from legacy `embulk-standards`.

### DIFF
--- a/embulk-guess-csv_verify/build.gradle
+++ b/embulk-guess-csv_verify/build.gradle
@@ -10,6 +10,12 @@ version = "${rootProject.version}"
 description = "Verification-purpose Embulk CSV guess plugin to compare the old Ruby-based one and the new Java-based one (not for your production use)"
 
 dependencies {
+    compile project(":embulk-api")
+    compile project(":embulk-spi")
+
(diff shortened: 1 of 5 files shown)


shroman (approved):
That's very hard to be sure it's all correct without unit tests, but since this is experimental (maybe README in the parent directory required?), let me approve it.

dmikurube:
Thanks! Yeah, its guess mechanism is a copy from the "original" Ruby implementation (see #1415) with some classloading preparation and the modern `embulk-parser-csv` replacement so that it "keeps" the original behavior. Then, this PR is to get the replacement back to the legacy `org.embulk.standards.CsvParserPlugin`.

We might want some tests especially about classloading and other hacks (not so much about the guess mechanism), but such testing wouldn't be very simple... Let's go ahead with releasing it as of now, and let's see how it goes.

PR #1422, opened 2021-04-15
Update the dependency lock of embulk-guess-csv_verify against Joda-Time removal

dmikurube:
Due to the commits and merge order of #1406 and #1420, `embulk-guess-csv_verify`'s dependency lock still contained dependencies to Joda-Time on `master`.


trung-huynh (approved):

dmikurube:
Thanks!

PR #1424, opened 2021-04-15
Use embulk-util-json's JsonParser inside org.embulk.spi.json.JsonParser

dmikurube:
This is one of the steps to hide Jackson behind embulk-deps.


dmikurube:
While we have some more PRs, let me merge this now. Thanks!

PR #1426, opened 2021-04-15
Stop handling @ConfigInject

dmikurube:
`@ConfigInject` has been like `@Config` in plugin `Task`, but assigned a value from the Guice injector, not from the config. As explained in https://dev.embulk.org/topics/catchup-with-v0.10.html, it is removed.

Typical use-cases were `BufferAllocator` and `ScriptingContainer` (JRuby) like below.

```
interface PluginTask extends Task {
    @Config("someConfig")
    String getSomeConfig();

    @ConfigInject
    BufferAllocator getBufferAllocator();

    @ConfigInject
    ScriptingContainer getJRuby();
}
```

`BufferAllocator` can be now retrieved by `Exec.getBufferAllocator()`. JRuby `ScriptingContainer` is no longer available for plugins.


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1428, opened 2021-04-15
Move Jackson Modules to org.embulk.config

dmikurube:
As a preparation to move all Jackson Modules to `embulk-deps`, moving all Jackson Modules to package `org.embulk.config` to be referred from `org.embulk.config.ModelManager` without `public`.


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1429, opened 2021-04-15
Register Jackson Modules inside ModelManager

dmikurube:
As a preparation to move all Jackson Modules to `embulk-deps`, registering all Jackson Modules inside `org.embulk.config.ModelManager` to have references in a closed loop.


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1432, opened 2021-04-15
SerDe TimestampFormat with a Jackson Module

dmikurube:
Several Embulk classes have had annotated with Jackson's annotations like `@JsonCreator`, `@JsonProperty`, and `@JsonValue`. To remove Jackson from `embulk-core`, those annotations need to be removed.

Instead of the annotations, the classes can still be serialized/deserialized by registering a Jackson Module to `ObjectMapper` with appropriate `JsonSerializer` and `JsonDeserializer` implemented. It works only for SerDe with `ModelManager`, though. (It won't work for arbitrary `ObjectMapper`, but we give it up. We haven't observed such a plugin so far.)

It is a replacement from annotations to a SerDe Module for `org.embulk.spi.time.TimestampFormat`.


trung-huynh (approved):
LGTM

dmikurube:
Thanks! Needs #1431 to be reviewed before merging this.

dmikurube:
Thanks!

PR #1436, opened 2021-04-15
Move ConfigLoader behind the embulk-deps DependencyClassLoader

dmikurube:
`ConfigLoader`, a big direct user of Jackson, is to load JSON/YAML into `ConfigSource`. It moves the main implementation to `embulk-deps` behind a sub class loader to hide Jackson from plugins.


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1437, opened 2021-04-15
Move all Jackson usage to embulk-deps, and remove Jackson dependencies from embulk-core

dmikurube:
Finally! It is to move all Jackson usage into `embulk-deps` behind a sub class loader to hide Jackson from plugins. It removes Jackson-related dependencies from `embulk-core`'s `build.gradle`.


trung-huynh (approved):
LGTM

dmikurube:
Resolved a conflict on `embulk-guess-csv_verify` (which has been removed in #1456). Thanks!

PR #1440, opened 2021-04-16
Remove Guice JRubyScriptingModule

dmikurube:
We're going to remove Guice from Embulk v0.10.33. This pull request is to remove Guice's `JRubyScriptingModule`, and to manage the JRuby instance directly in `ExecSessionInternal`.


shroman (approved):
lgtm within this repository scope, but tbh I'm not very sure about the scope of the impact of this removal to plugins.

dmikurube:
Thanks!

> the scope of the impact of this removal to plugins.

If a Java plugin calls JRuby's feature in it, it could have been impacted. But :

* As far as I searched OSS plugins, there are no such cases.
* We have announced "Stop using JRuby in your plugin." in the article: https://dev.embulk.org/topics/get-ready-for-v0.11-and-v1.0.html#using-jruby-orgjruby-classes

So, I believe it should be fine.

dmikurube:
Thanks!

PR #1362, opened 2020-12-15
Replace InjectedPluginSource to Guice-independent BuiltinPluginSource

dmikurube:
`embulk-standards` plugins, some built-in plugins, and some special plugins in testing had been managed in `InjectedPluginSource` through Guice.
* `embulk-standards` plugins are now loaded through `PluginClassLoader`.
* Some built-in plugins are loaded directly.
* Then, some special plugins for testing are remaining.

We're going to remove Guice soon. This PR replaces `InjectedPluginSource` to a new Guice-independent `BuiltinPluginSource` so that tests can register plugins without Guice.

PR #1442, opened 2021-04-16
Remove Guice ExtensionServiceLoaderModule

dmikurube:
We're going to remove Guice from Embulk v0.10.33. This pull request is to remove Guice's `ExtensionServiceLoaderModule`, which has been used to manage "ExtensionService" used for standard plugins.

Standard plugins are now loaded through `BuiltinPluginSource` (#1362), then no longer needed.


shroman (approved):

dmikurube:
Thanks!

PR #1443, opened 2021-04-16
Remove Guice ExecModule

dmikurube:
We're going to remove Guice from Embulk v0.10.33. This pull request is to remove Guice's `ExecrModule`, which has been used to manage types of Guice bindings -- logger, `BufferAllocator`, and `TempFileSpaceAllocator`. They will be managed directly in `ExecSessionInternal`, instead.


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1445, opened 2021-04-19
Move base classes for testing to embulk-junit4 from embulk-core:tests

dmikurube:
Some plugins have needed `embulk-core:tests` in their `testCompile` dependencies to test the plugins. But, that is not very healthy, nor reasonable.

This pull request moves classes that can be needed from plugin's tests to `embulk-junit4` from `embulk-core:tests`. It became possible after the Guice removal.


dmikurube:
Thanks!

PR #1448, opened 2021-04-20
Remove Guava

dmikurube:
Like we did move Jackson from embulk-core's top-level class loader in #1437, we'd also like to remove Guava so that embulk-core's Guava does not interfere plugin's Guava.

But unlike Jackson, all Guava usage in Embulk can be replaced with trivial use of Java 8 features. We decided to remove all Guava usage, then.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/spi/util/dynamic/LongColumnSetter.java
+++ b/embulk-core/src/main/java/org/embulk/spi/util/dynamic/LongColumnSetter.java
@@ -1,7 +1,5 @@
 package org.embulk.spi.util.dynamic;
 
-import com.google.common.math.DoubleMath;
-import java.math.RoundingMode;
 import java.time.Instant;
 import org.embulk.spi.Column;
 import org.embulk.spi.PageBuilder;
@@ -32,8 +30,15 @@ public class LongColumnSetter extends AbstractDynamicColumnSetter {
     public void set(double v) {
         long lv;
         try {
-            // TODO configurable rounding mode
-            lv = DoubleMath.roundToLong(v, RoundingMode.HALF_UP);
+            final double roundedDouble = Math.rint(v);
+            final double diff = v - roundedDouble;
+            if (diff == 0.5) {
+                lv = (long) (v + 0.5);
+            } else if (diff == -0.5) {
+                lv = (long) (v - 0.5);
+            } else {
+                lv = (long) roundedDouble;
+            }
         } catch (ArithmeticException ex) {
             // NaN / Infinite / -Infinite
             defaultValue.setLong(pageBuilder, column);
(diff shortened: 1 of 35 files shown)


embulk-core/src/main/java/org/embulk/spi/util/dynamic/LongColumnSetter.java
```diff
@@ -32,8 +30,15 @@ public void set(long v) {
...
-            lv = DoubleMath.roundToLong(v, RoundingMode.HALF_UP);
+            final double roundedDouble = Math.rint(v);
+            final double diff = v - roundedDouble;
+            if (diff == 0.5) {
+                lv = (long) (v + 0.5);
+            } else if (diff == -0.5) {
+                lv = (long) (v - 0.5);
+            } else {
+                lv = (long) roundedDouble;
+            }
```
legiangthanh, line 41: I think we have an issue if v has an even integer part, e.g 2.5.

dmikurube: What's the issue?  If `v` is exact `2.5`,

* `roundedDouble` would be `2.0` from the definition of `Math.rint()`.
* `diff` would be `0.5` from `2.5 - 2.0`.
* `lv` would be `3` from `2.5 + 0.5`.

Then it's the expected `HALF_UP` behavior.

legiangthanh: Sorry, I just have a mistake. Actually, the issue happens when v has an integer part is 1.5

`v` is exact `1.5`
rounded double would be `2.0`
`diff` would be `-0.5` from `1.5 - 2.0`
lv would be `1` from `1.5 -0 .5`.
I think we expect 2 in here.

dmikurube: Ah, probably you're right. I'll have another look into Guava, and have a comparison.

dmikurube: Thanks for your good catch. I should have used the sign of `v`, not the sign of `diff`, for `0.5`. Fixed it in: e65f6ecd21a728df893bee57dc47fc691f13470e

legiangthanh (approved):
LGTM

dmikurube:
Added the core team to review...

hiroyuki-sato (approved):
LGTM 👍

dmikurube:
We confirmed that the previous v0.10.38 worked well with almost all the plugins we have. It's time to merge this after the 1.5yr long interval, and to release v0.10.39!

dmikurube:
Oops, the test failed after merging the old pull request... The merge might have conflicted semantically with recent changes after this pull request got ready. Let me have a look. https://github.com/embulk/embulk/actions/runs/3627519586/jobs/6117488258

PR #1450, opened 2021-04-23
Cleanup "parent_first" series for Jackson and JRuby

dmikurube:
`parent_first_packages` and `parent_first_resources` have been controlling a class loading order in `PluginClassLoader` so that `embulk-core`'s direct dependencies are used.

It has not been a perfect mechanism to control dependency hell, in fact. We may remove this mechanism completely in the end of v0.10, but removing just related things in the meantime.

Some around JRuby should have already been removed beforehand, but forgotten. Removing them, too, now.

### `parent_first_packages`

* `com.fasterxml`: removed in v0.10.31
* `com.headius`: JRuby
* `com.jcraft.jzlib`: JRuby
* `com.kenai`: JRuby
* `com.martiansoftware.nailgun`: JRuby
* `javax.validation`: removed in v0.10.31
* `jnr`: JRuby
* `net.jcip.annotations`: JRuby
* `org.aopalliance`: JRuby
* `org.apache.bval`: removed in v0.10.31
* `org.apache.commons`: removed in v0.10.31
* `org.fusesource.hawtjni.runtime`: JRuby
* `org.jcodings`: JRuby
* `org.joni`: JRuby
* `org.jruby`: JRuby
* `org.yecht`: JRuby

### `parent_first_resources`

* `com/martiansoftware/nailgun/builtins`: `com.martiansoftware.nailgun` (JRuby)
* `jni`: `jni` (JRuby)
* `jruby`: JRuby
* `net/jcip/annotations`: `net.jcip.annotations` (JRuby)
* `okay`: JRuby
* `org/apache/bval/jsr303`: `org.apache.bval` (removed in v0.10.31)
* `tables`: ? (Only Apache License template in ancient Embulk v0.6)
* `templates`: `org.jcodings` (JRuby)


embulk-core/src/main/resources/embulk/parent_first_resources.properties
```diff
@@ -1,22 +1,5 @@
...
-jni/x86_64-SunOS
-jni/x86_64-Windows
-jruby
 msgpack
-net/jcip/annotations
-okay
-org/apache/bval/jsr303
-tables
-templates
+org/embulk
```
trung-huynh, line 5: Is it for those new embulk-util-* resources?

trung-huynh (approved):
LGTM

embulk-core/src/main/resources/embulk/parent_first_resources.properties
dmikurube: Ah, actually not. None of `embulk-util-*` are loaded in the top-level class loader, then they would never conflict.

This is "just in case". I might not need to add this here, but `embulk-core` has some resources in `org/embulk`.
https://github.com/embulk/embulk/tree/v0.10.31/embulk-core/src/main/resources

dmikurube:
Thanks!

PR #1453, opened 2021-04-23
Move dependency on javax.inject to embulk-api from embulk-core

dmikurube:
Note: we need to reconsider this because `javax.inject.Inject` is going to be modified to `jakarta.inject.Inject`. There will be no compatibility.
https://github.com/eclipse-ee4j/injection-api/blob/2.0.0/src/main/java/jakarta/inject/Inject.java


dmikurube:
We decided not to do this. We do #1461 instead.

PR #1459, opened 2021-05-20
Upgrade msgpack-core to 0.8.24

dmikurube:
Embulk's "JSON"-type values have been represented by msgpack-java's `org.msgpack.value.(Immutable)Value` since Embulk v0.7. The msgpack-java has set its roots deep inside Embulk, including its plugin SPI method signatures.

It has brought a heavy dependency hell between Embulk and msgpack-java. We have hesitated even just upgrading `msgpack-core` which can potentially cause a compatibility issue. As a result, a couple of plugins that use `msgpack-core` (such as `embulk-parser-msgpack`) had to stay with older `msgpack-core`. Keeping it old is definitely bad. 

The Embulk v0.10 "development" series is a good chance to catch-up. Let's upgrade `msgpack-core` to 0.8.24, which is the latest v0.8 -- said to be still "compatible" expectedly.

However, msgpack-java v0.9 added a support for a new MessagePack type. Compatibility issues may happen more likely. We do not go with msgpack-java v0.9.

----

Even though we stay with v0.8 now, it wouldn't continue for good. Instead of future "catch-ups" continued, we'll drop msgpack-java from Embulk in the long term. 

Indeed, the Embulk core has never used `MessagePacker` nor `MessageUnpacker` from `msgpack-core`. Embulk has used only `org.msgpack.value.(Immutable)Value` just as a JSON-like data container. It's nonsense!

----

The long-term plan would be:

1. We upgrade `msgpack-core` to v0.8.24 in Embulk v0.10.42 in this pull request #1459.
2. We introduce another set of classes for "JSON" type (expected to be `org.embulk.spi.json.JsonValue`) in v0.10.42 in another pull request #1462.
    * We'll keep SPI methods for both (`org.msgpack.value.Value` and `org.embulk.spi.json.JsonValue`) for a long time -- even after Embulk v1.0.
    * It'd be similar to "TIMESTAMP" type -- `org.embulk.spi.time.Timestamp` and `java.time.Instant`.
    * Plugins are expected to migrate to use `org.embulk.spi.json.JsonValue` through Embulk v0.11 - v1.0.
3. At some point in the future, after Embulk v1.0, `msgpack-core` will be removed.

We'll also write up an EEP for this.

### DIFF
--- a/embulk-api/build.gradle
+++ b/embulk-api/build.gradle
@@ -10,6 +10,14 @@ configurations {
 }
 
 dependencies {
-    api "org.msgpack:msgpack-core:0.8.11"
     api "org.slf4j:slf4j-api:1.7.30"
+
+    // It will be removed in the future, but not very soon from Embulk v1.0.0.
+    //
+    // Embulk still needs to have the entire "msgpack-core" in "embulk-api" because:
+    // * Signatures of some methods in "embulk-api" depend on "org.msgpack.value.Value".
+    //   It needs time for mitigating impacts to remove them.
+    // * "msgpack-core" will not split their "value" classes to another artifact very soon.
+    //   https://github.com/msgpack/msgpack-java/pull/571
+    api "org.msgpack:msgpack-core:0.8.24"
 }
--- a/embulk-api/gradle/dependency-locks/compileClasspath.lockfile
+++ b/embulk-api/gradle/dependency-locks/compileClasspath.lockfile
@@ -1,5 +1,5 @@
 # This is a Gradle generated file for dependency locking.
 # Manual edits can break the build and are not advised.
 # This file is expected to be part of source control.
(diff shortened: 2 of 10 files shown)


dmikurube:
Hmm, even in a draft pull request, the Core Team is forcibly assigned... Let me think. Please ignore the review request.

hiroyuki-sato:
It is marked as 🌙 It seems that the meaning is the following.

> embulk/core-team will be requested when the pull request is marked ready for review.

lock

dmikurube:
Ah, I see, the 🌙 mark means like that...

hiroyuki-sato (approved):
LGTM👍

vietnguyen-td (approved):
LGTM

dmikurube:
Thanks!

PR #1460, opened 2021-05-20
Remove @org.embulk.exec.ForGuess

dmikurube:
`@org.embulk.exec.ForGuess` is no longer used, especially after Guice is removed.


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1461, opened 2021-05-20
Remove dependency on javax.inject

dmikurube:
We had originally planned to keep only `javax.inject` in dependencies so that plugins will be able to get some instances injected. `javax.inject` has been considered to be a semi-standard artifact, unlike Guice as a third-party library.

But, `javax.inject` will no longer continue. It will change to `jakarta.inject`. They change the Java packages, too, then they have no compatibility. It will cause dependency problems in the future again.

Finally, we decided to remove `javax.inject`, too.


shroman (approved):
lgtm
do we want to mention it in some developer/release notes?

dmikurube:
Thanks!

> do we want to mention it in some developer/release notes?
I basically don't think so -- no OSS plugins have never used injections. Just a changelog in Git annotations would be fine.

PR #1444, opened 2021-04-16
Remove Guice

dmikurube:
We're going to remove Guice from Embulk v0.10.33. All Guice `Module`s have been removed in #1440, #1441, #1442, and #1443. This pull request is finally remove Guice.

PR #1466, opened 2021-07-29
Remove Guava from embulk-deps

dmikurube:
Guava (18.0) is already in the top-level class loader (`embulk-core`), then `embulk-deps` should not have Guava (also, it's 15.0!).

In fact, 0.10.32's `embulk-deps` should not have included Guice, too. (It's anyway removed in 0.10.32 at: https://github.com/embulk/embulk/pull/1444/files#diff-a0a01b2cd5bc9398f71f2c5e71d140d35a353e0a5b72605152be03718f78d971L35)


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1467, opened 2021-07-30
Fix SelfContainedJarAwareURLClassLoader against "findResources" for non-self-contained embulk-deps

dmikurube:
`DependencyClassLoader#findResources` (not `findResource`) could fail when `embulk-deps` was deployed directly on the file system, not embedded as self-contained JARs in Embulk's executable JAR.

* This problem does not affect normal OSS Embulk users. It could affect only for users who deploy JARs by themselves.
* This problem does not affect before v0.10.32 because `embulk-deps` did not include libraries calling `findResources`. It affected only v0.10.32.

This PR is to fix the problem by catching its related Exception(s).

### DIFF
--- a/embulk-core/src/main/java/org/embulk/deps/SelfContainedJarAwareURLClassLoader.java
+++ b/embulk-core/src/main/java/org/embulk/deps/SelfContainedJarAwareURLClassLoader.java
@@ -16,6 +16,8 @@ import java.util.Optional;
 import java.util.Vector;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


shroman:
`Holder.logger.debug` not INFO at least to know what is happening with loading?
DEBUG will most likely be always hidden.

dmikurube:
Thanks for taking a look!

I at first thought that logging them only when debugging locally is okay, but changed my mind better to have them always. Updated it to `WARN` in 0e1b4333892cd85fd66fedd82c3e85536429497b.

shroman:
Yeah, it's always better to make them silent later if needed or via configurations instead of missing this, IMO. Thank you!

shroman (approved):

PR #1470, opened 2021-08-13
Fix schema guess for so large a long integer value

dmikurube:
The same fix with https://github.com/embulk/embulk-util-guess/pull/14


shroman (approved):

dmikurube:
Thanks!

PR #1476, opened 2021-09-29
Fix the TestingEmbulk#runOutput problem of not skipping the header even though the column name:type needs to be defined in the header.

civitaspo:
`TestingEmbulk#runOutput` requires a CSV file as an input data source and the CSV file must have the header like `columnName:columnType,columnName:columnType,...`, but the csv parser setting does not include `skip_header_lines=1`, so I added.


dmikurube (approved):
LGTM! Thanks for working on it.

But sorry that we'll need some time to release the next version v0.10.34.

civitaspo:
@dmikurube thanks for merging it!!

> But sorry that we'll need some time to release the next version v0.10.34.

np:)

PR #1480, opened 2021-10-11
Update standard plugins into v0.10.35 to embed in Embulk v0.10.34

dmikurube:
`embulk-standards` plugins have been moved into https://github.com/embulk/embulk-standards, and we have already released their v0.10.35.

The next Embulk v0.10.34 would contain the latest `embulk-standards` v0.10.35.


dmikurube:
Thanks!

PR #1481, opened 2021-10-12
Fallback to "plugins.default.<category>.<type>" if no plugin is found from "plugins.<category>.<type>"

dmikurube:
Users (especially ourselves) sometimes want to set a "default" plugin version from Embulk System Properties, while sometimes wanted to override the default version with another version.

Users can just override the default Properties, of course, but it often happens that their deployment tests things only with the "default" version, while the "overridden" version is not tested very well.

Even if they fail to load the overridden version, they often want to fallback to the default version, which is well tested.

----

This pull-request is realize such a want. For example, a user may set `plugins.default.input.s3=maven:org.embulk:0.5.2` and `plugins.input.s3=maven:org.embulk:0.6.0`.

Then, `0.6.0` is loaded if `maven:org.embulk:embulk-input-s3:0.6.0` exists in the deployment.

Otherwise if `maven:org.embulk:embulk-input-s3:0.6.0` does not exist, `0.5.2` is loaded.


trung-huynh (approved):
LGTM.

dmikurube:
Thanks!

PR #1482, opened 2021-10-12
Support `override_dependencies` and `exclude_dependencies` in Maven-based `type`

dmikurube:
This is an alternative to JDBC plugins' `driver_path` configuration built in Embulk's core.

Users sometimes want to switch plugin's dependencies at runtime, without hacking the plugin. This configuration allows users to switch with Maven artifact names.

For example,

```
type:
  source: maven
  group: org.embulk
  name: mysql
  version: 0.10.2
  override_dependencies:
    - mysql:mysql-connector-java:8.0.11
    # Overrides existing mysql:mysql-connector-java:5.1.44 which has the same groupId and artifactId.
```

```
type:
  source: maven
  group: org.embulk
  name: sqlserver
  version: 0.10.2
  exclude_dependencies:
    - com.microsoft.sqlserver:mssql-jdbc
  override_dependencies:
    - net.sourceforge.jtds:jtds:1.3.1
    # It cannot override existing com.microsoft.sqlserver:mssql-jdbc, so excluding explicitly above.
```

### DIFF
--- a/embulk-core/src/main/java/org/embulk/plugin/MavenPluginType.java
+++ b/embulk-core/src/main/java/org/embulk/plugin/MavenPluginType.java
@@ -2,18 +2,39 @@ package org.embulk.plugin;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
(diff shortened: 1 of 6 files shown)


trung-huynh (approved):
nice one! LGTM.

dmikurube:
Thanks!

PR #1484, opened 2021-10-15
Cache Maven-based plugin's loaded classes in MavenPluginRegistry (Fix #1483)

### LINKED ISSUE #1483: Class-load Maven-based plugins only once per ExecSession
Maven-based plugin's classes loaded via `MavenPluginSource` have not been cached, and it has caused unnecessary repeats of loading the same classes via different `PluginClassLoader`s. It could have caused out of Java Metaspace more easily.

RubyGem-based plugin's (Ruby/Java) classes loaded via `plugin.rb`, `java_plugin.rb`, `plugin_registry.rb`, and `JRubyPluginSource` have actually been cached in `Embulk::PluginRegistry`. The Maven-based plugins should have followed it.

~From a File Input plugin (via `FileInputRunner`), its configured Parser plugin and Decoder plugins are class-loaded many times, estimated to be per file. The same for File Output : Formatter and Encoder.~

~Due to this, Java Metaspace can be consumed up. It didn't happen so often because the major CSV/JSON Parsers were on the core side until v0.10.28. But, after moving them to ordinary `PluginClassLoader`, it is now revealed as a problem.~

~https://github.com/embulk/embulk/blob/v0.10.33/embulk-core/src/main/java/org/embulk/spi/FileInputRunner.java#L95-L98~

dmikurube:
See #1483 for the context. It caches Maven-based plugin's classes loaded by the same `MavenPluginSource` instance (shared per the same `ExecSessionInternal` instance).

Note that it caches only classes (`java.lang.Class` such as `S3InputPlugin.class`). Any instance created from the class (such as `new S3InputPlugin()`) is not cached.

### DIFF
--- a/embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginRegistry.java
+++ b/embulk-core/src/main/java/org/embulk/plugin/maven/MavenPluginRegistry.java
@@ -23,6 +23,8 @@ import org.embulk.spi.GuessPlugin;
 import org.embulk.spi.InputPlugin;
 import org.embulk.spi.OutputPlugin;
 import org.embulk.spi.ParserPlugin;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Caches Maven-based plugin's loaded classes such as embulk-ruby's Embulk::PluginRegistry.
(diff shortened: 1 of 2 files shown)


dmikurube:
Thanks, both @himadieievsv and @trung-huynh. Let me merge this, and release v0.10.34. Then, I'll have a backport release v0.10.29.1, too.

PR #1486, opened 2021-10-18
Cache self-contained plugin's loaded classes in SelfContainedPluginRegistry

dmikurube:
A cache similar to #1484, but for self-contained plugins, which are in Embulk OSS distributions.


dmikurube:
Thanks!

PR #1490, opened 2021-11-24
Backport v0.9.24

dmikurube:
### Change 1: Back-port new v0.10 APIs to v0.9

Back-porting some new methods into v0.9, which were added during v0.10, so that "modernized" plugins can keep working with v0.9 without a hack (ex. call this method if it exists, otherwise call another...) for a reasonable duration.

We haven't done the back-port so far because it was difficult to declare that the API/SPI is stable fixed. But now, we are approaching v0.11, then we are now ready to declare that the API/SPI is almost "fixed". 

```
org.embulk.spi.Exec.getTransactionTimeInstant()
org.embulk.spi.Exec.getPageBuilder(BufferAllocator, Schema, PageOutput)
org.embulk.spi.Exec.getPageReader(Schema)
```

```
org.embulk.spi.ExecSession#getTransactionTimeInstant()
org.embulk.spi.ExecSession#getTransactionTimeString()
org.embulk.spi.ExecSession#getPageBuilder(BufferAllocator, Schema, PageOutput)
org.embulk.spi.ExecSession#getPageReader(Schema)
```

```
org.embulk.spi.PageBuilder#setTimestamp(Column, Instant)
org.embulk.spi.PageBuilder#setTimestamp(int, Instant)
```

```
org.embulk.spi.PageReader#getTimestampInstant(Column)
org.embulk.spi.PageReader#getTimestampInstant(columnIndex)
```

```
org.embulk.spi.time.Timestamp.ofString(String)
```

### Change 2: Add `com.fasterxml.jackson.datatype.jdk8` in `parent_first_packages`

This is the same with #1489 for `v0.9`.

`embulk-core` has not have `com.fasterxml.jackson.datatype.jdk8` in `parent_first_packages.properties` while :
* `parent_first_packages.properties` has had other `com.fasterxml.jackson.???`.
* `embulk-core` has had `jackson-datatype-jdk8` in dependencies until v0.10.31.

It has caused no problems when all "v0.10-ready" plugins are using only Jackson 2.6.7. But, `embulk-output-td` needed to use Jackson 2.8 and 2.9.

`embulk-output-td` has caught up with Embulk v0.10 API/SPI in its `0.8.0`, and it needed to have `jackson-datatype-jdk8` in addition.
* https://github.com/treasure-data/embulk-output-td/pull/111

----

By this addition, Embulk v0.9 - v0.10.31 loads `jackson-databind:2.6.7` from `embulk-core` while `embulk-output-td` tries to use its embedded `jackson-datatype-jdk8:2.9.10`. We observed that the mismatch caused `NoSuchMethodError` like :

```
Caused by: java.lang.NoSuchMethodError: com.fasterxml.jackson.databind.type.ReferenceType.upgradeFrom(Lcom/fasterxml/jackson/databind/JavaType;Lcom/fasterxml/jackson/databind/JavaType;)Lcom/fasterxml/jackson/databind/type/ReferenceType;
	at com.fasterxml.jackson.datatype.jdk8.Jdk8TypeModifier.modifyType(Jdk8TypeModifier.java:40)
        ...
```

----

A "caught-up" should basically work with the latest Embulk v0.9 and recent v0.10s so that users/plugin-developers do not need a one-way jump.

This hotfix is to make it work with v0.9.24, the new latest stable v0.9.

We release only an executable binary (`embulk-v0.9.24.jar`) on this for OSS community users.


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1491, opened 2021-11-24
Add com.fasterxml.jackson.datatype.jdk8 in parent first packages v0.10.31

dmikurube:
`embulk-core` has not have `com.fasterxml.jackson.datatype.jdk8` in `parent_first_packages.properties` while :
* `parent_first_packages.properties` has had other `com.fasterxml.jackson.???`.
* `embulk-core` has had `jackson-datatype-jdk8` in dependencies until v0.10.31.

It has caused no problems when all "v0.10-ready" plugins are using only Jackson 2.6.7. But, `embulk-output-td` needed to use Jackson 2.8 and 2.9.

`embulk-output-td` has caught up with Embulk v0.10 API/SPI in its `0.8.0`, and it needed to have `jackson-datatype-jdk8` in addition.
* https://github.com/treasure-data/embulk-output-td/pull/111

----

By this addition, Embulk v0.9 - v0.10.31 loads `jackson-databind:2.6.7` from `embulk-core` while `embulk-output-td` tries to use its embedded `jackson-datatype-jdk8:2.9.10`. We observed that the mismatch caused `NoSuchMethodError` like :

```
Caused by: java.lang.NoSuchMethodError: com.fasterxml.jackson.databind.type.ReferenceType.upgradeFrom(Lcom/fasterxml/jackson/databind/JavaType;Lcom/fasterxml/jackson/databind/JavaType;)Lcom/fasterxml/jackson/databind/type/ReferenceType;
	at com.fasterxml.jackson.datatype.jdk8.Jdk8TypeModifier.modifyType(Jdk8TypeModifier.java:40)
        ...
```

----

A "caught-up" should basically work with the latest Embulk v0.9 and recent v0.10s so that users/plugin-developers do not need a one-way jump.

This hotfix is to make it work with `v0.10.31(.2)`.

We release only Maven artifacts (`embulk-core`, `embulk-deps`, ...) for v0.10.31.2.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ repositories {
 }
 
 group = "org.embulk"
-version = "0.10.31.1"
+version = "0.10.31.2"
 
 def subprojectNamesOfCoreArtifacts = [
     "embulk-api",
--- a/embulk-core/src/main/resources/embulk/parent_first_packages.properties
+++ b/embulk-core/src/main/resources/embulk/parent_first_packages.properties
@@ -4,6 +4,7 @@ com.fasterxml.jackson.annotation
 com.fasterxml.jackson.core
 com.fasterxml.jackson.databind
 com.fasterxml.jackson.datatype.guava
+com.fasterxml.jackson.datatype.jdk8
 com.fasterxml.jackson.module.guice
 com.google.common.annotations
 com.google.common.base


trung-huynh (approved):

dmikurube:
Thanks!

PR #1492, opened 2021-11-26
Throw more descritive Exceptions from PageBuilder in case of type mismatch

dmikurube:
We observed a case that `PageBuilder(Impl)` threw `IllegalStateException("Not reach here")`.

* https://github.com/embulk/embulk/blob/v0.10.29.1/embulk-core/src/main/java/org/embulk/spi/PageBuilderImpl.java#L356-L378

To debug such a case more easily, this pull-request makes the Exception messages more descriptive.


minidragon88 (approved):
LGTM

dmikurube:
Thanks!

PR #1498, opened 2022-03-31
Disable publishing a Gradle Module Metadata

dmikurube:
Similar to https://github.com/embulk/embulk-input-gcs/pull/52. We basically trust Maven `.pom` files. Having two types of dependency declarations would cause confusions.

See: https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html#sub:disabling-gmm-publication


trung-huynh (approved):
LGTM

dmikurube:
Thanks!

PR #1500, opened 2022-04-04
Deprecate `embulk selfupdate`

### LINKED ISSUE #1495: Deprecate `embulk selfupdate`
Background
-----------

`embulk selfupdate` had been useful in earlier days for casual users running Embulk on their local laptop occasionally. But recently, many of Embulk users run Embulk on the server side, and they apply careful version controls (both of the Embulk core and plugins) by themselves because compatibility issues are serious.

On the other hand, keeping `selfupdate` is a burden for us, the Embulk maintainers.

* Each Embulk binary has to embed the URL to download. (Such as `dl.embulk.org`)
* The Embulk team has to keep `dl.embulk.org` maintained to be updated from older Embulk.
* Maintaining `dl.embulk.org` is a pain. We have to maintain not just the code and releases, but also servers/services.

Because of those reasons, we're going to remove the `selfupdate` feature from Embulk at some point of v0.11.*.

Plan
-----

* We have already unsupported single `selfupdate` to update to the "latest" version in v0.10.0.
* During `v0.10.*` (expected to be in `v0.10.36`), we start warning messages against `embulk selfupdate` to notify that `selfupdate` will be unsupported at some point during `v0.11.*`. Start maintaining versions and deployment by yourself.
* At some point in `v0.11.*`, we actually remove the self-executable support.

Warning Messages
-----------------

```
================================== [ NOTICE ] ==================================
 Embulk `selfupdate` will not be available at some point in v0.11.*.
 Be prepared by controlling Embulk versions with your own deployment strategies.
 See https://github.com/embulk/embulk/issues/1495 for the details.
================================================================================
```

dmikurube:
Fixes #1495

### DIFF
--- a/embulk-core/src/main/java/org/embulk/cli/SelfUpdate.java
+++ b/embulk-core/src/main/java/org/embulk/cli/SelfUpdate.java
@@ -29,6 +29,12 @@ import org.slf4j.LoggerFactory;
  */
 class SelfUpdate {
     static void toSpecific(final String runningVersion, final String targetVersion, final boolean forced) throws IOException {
+        logger.info("================================== [ NOTICE ] ==================================");
+        logger.info(" Embulk `selfupdate` will not be available at some point in v0.11.*.");
+        logger.info(" Be prepared by controlling Embulk versions with your own deployment strategies.");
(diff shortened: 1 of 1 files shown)


trung-huynh (approved):

dmikurube:
Thanks!

PR #1503, opened 2022-04-22
Backport v0.10.32.2

dmikurube:
It backports #1467 to v0.10.32.

The issue was considered to be unrelated in v0.10.32 (as described in #1467), but actually, it affects when Embulk is called via `EmbulkEmbed`, not from a command-line.


trung-huynh (approved):

dmikurube:
Thanks!

Issue #1485, opened 2021-10-15
Eliminate `PluginSource`-based plugin lookup, do things more integrated


dmikurube:
While Embulk used Guice to look through `PluginSource`s, they needed to inherit `PluginSource`. But it has been making the error messages cryptic.

We can make the plugin lookup mechanism more organized through the whole, and make error messages more understandable.

PR #1523, opened 2022-12-02
Fix caching in MavenPluginRegistry

dmikurube:
We found that #1481 (v0.10.35) broke the plugin class cache by #1484 (v0.10.34).

I believe that the plugin lookup mechanism should be totally refactored as filed in #1485, so that we can avoid this kind of bugs in the future. But as of now, let me fix just this before #1448, the planned next big removal.


vietnguyen-td (approved):
LGTM

legiangthanh (approved):
LGTM

dmikurube:
Thanks!

PR #1527, opened 2022-12-07
Clean up and comment on Guava dependencies

dmikurube:
* Removing "exclude guava" in a dependency to "embulk-deps" for the executable binary.
* Commenting to remove Guava from "embulk-core" test dependencies.
* Re-adding Guava in "embulk-junit4" dependencies, which was just "compileOnly" because "embulk-core" had Guava.


vietnguyen-td (approved):
LGTM

dmikurube:
Thanks!

PR #1532, opened 2022-12-09
Log the version of the loaded JRuby instance. (Fix #1521)

### LINKED ISSUE #1521: Logging JRuby Version useful for debugging Ruby Plugins.
A User chooses the JRuby version explicitly in Embulk 0.10.x.
Logging output JRuby Version helps debug JRuby plugins.

hiroyuki-sato:
This PR outputs log the JRuby version Fix #1521

embulk run

```
2022-12-09 11:22:10.411 +0900 [INFO] (0001:transaction): Loaded JRuby 9.4.0.0 Compatible MRI: 3.1.0
```

embulk gem

```
2022-12-09 11:24:36.747 +0900 [INFO] (main): Loaded JRuby 9.4.0.0 Compatible MRI: 3.1.0
```

### DIFF
--- a/embulk-core/src/main/java/org/embulk/jruby/LazyScriptingContainerDelegate.java
+++ b/embulk-core/src/main/java/org/embulk/jruby/LazyScriptingContainerDelegate.java
@@ -17,13 +17,15 @@ public final class LazyScriptingContainerDelegate extends ScriptingContainerDele
             final ClassLoader classLoader,
             final LocalContextScope delegateLocalContextScope,
             final LocalVariableBehavior delegateLocalVariableBehavior,
-            final JRubyInitializer initializer) {
+            final JRubyInitializer initializer,
+            final Logger logger) {
         this.impl = null;
 
         this.classLoader = classLoader;
         this.delegateLocalContextScope = delegateLocalContextScope;
         this.delegateLocalVariableBehavior = delegateLocalVariableBehavior;
         this.initializer = initializer;
+        this.logger = logger;
     }
 
     public static LazyScriptingContainerDelegate withGems(
@@ -104,7 +106,8 @@ public final class LazyScriptingContainerDelegate extends ScriptingContainerDele
                             ? ScriptingContainerDelegate.LocalContextScope.SINGLETON
                             : ScriptingContainerDelegate.LocalContextScope.SINGLETHREAD,
                     ScriptingContainerDelegate.LocalVariableBehavior.PERSISTENT,
-                    initializer);
+                    initializer,
+                    logger);
             if (useGlobalRubyRuntime) {
                 // In case the global JRuby instance is used, the instance should be always initialized.
                 // Ruby tests (in embulk-ruby/) are examples.
@@ -315,6 +318,7 @@ public final class LazyScriptingContainerDelegate extends ScriptingContainerDele
                     this.classLoader, this.delegateLocalContextScope, this.delegateLocalVariableBehavior);
             if (this.initializer != null) {
                 this.initializer.initialize(this.impl);
+                this.logger.info("Loaded JRuby runtime {}", this.impl.getJRubyVersion());
             }
         }
         return this.impl;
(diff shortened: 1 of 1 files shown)


embulk-core/src/main/java/org/embulk/jruby/LazyScriptingContainerDelegate.java
```diff
@@ -315,6 +317,7 @@ synchronized ScriptingContainerDelegateImpl getInitialized() {
                     this.classLoader, this.delegateLocalContextScope, this.delegateLocalVariableBehavior);
             if (this.initializer != null) {
                 this.initializer.initialize(this.impl);
+                this.logger.info(String.format("Loaded JRuby %s Compatible MRI: %s",this.impl.getJRubyVersion(),this.impl.getRubyVersion()));
```
dmikurube, line 320: * You can use `"{}"` for string replacement in SLF4J loggers.
* The "Ruby version" from JRuby is often confusing. Let's just have the JRuby version.

```suggestion
                this.logger.info("Loaded JRuby runtime {}", this.impl.getJRubyVersion()));
```

embulk-core/src/main/java/org/embulk/jruby/LazyScriptingContainerDelegate.java
```diff
@@ -104,7 +106,7 @@ private static LazyScriptingContainerDelegate of(
                             ? ScriptingContainerDelegate.LocalContextScope.SINGLETON
                             : ScriptingContainerDelegate.LocalContextScope.SINGLETHREAD,
                     ScriptingContainerDelegate.LocalVariableBehavior.PERSISTENT,
-                    initializer);
+                    initializer,logger);
```
dmikurube, line 109: We have a space, or a newline, after a comma. In this case, preceding parameters are newline-separated, then let's follow them.

```suggestion
                    initializer,
                    logger);
```

dmikurube (commented):
Left some minor comments.

embulk-core/src/main/java/org/embulk/jruby/LazyScriptingContainerDelegate.java
hiroyuki-sato: Thanks! done.

embulk-core/src/main/java/org/embulk/jruby/LazyScriptingContainerDelegate.java
hiroyuki-sato: Thanks! fixed.

The output MRI version helps check the RubyGems site.
I don't know any website or document mapping JRuby and MRI versions.

スクリーンショット 2022-12-09 13 25 51

dmikurube: It's usually mentioned in their release notes : 
* "Compatibility is updated to Ruby 3.1 level." - https://www.jruby.org/2022/11/23/jruby-9-4-0-0.html
* "JRuby 9.3.x is compatible with Ruby 2.6.x and stays in sync with C Ruby." - https://www.jruby.org/2022/10/24/jruby-9-3-9-0.html

But, the information is sometimes not perfectly accurate. Some recent Ruby features might be implemented earlier, and some might delay. Even if we show "Compatible Ruby version", that can be wrong. Users would ask us "It should it's 3.X.X-compatible, but does not work!", and we have to investigate by ourselves... I'd like to avoid that.

As the Embulk project, I don't have intention (and resources) to take care of JRuby versions. Choosing an appropriate JRuby version would be user's responsibility. Users have responsibility to look into https://www.jruby.org/ by themselves, and to think which version they should use.

hiroyuki-sato: Thanks! I'll create a compatible table by myself. (For personal use)

dmikurube (approved):
Thanks! At last, could you : 
* Squash the commits in the PR, and
* Update the Subject to "Log the version of the loaded JRuby instance"

hiroyuki-sato:
Thank you for your comment! squashed and updated the subject.

dmikurube:
Thanks. Merging this for v0.10.40...

PR #1534, opened 2022-12-09
Add methods in org.embulk.config.DataSource to handle Lists and Maps (nested attributes) better

dmikurube:
Since v0.10.32, we have split the core Jackson, and plugin's Jackson. It had caused some compatibility issues even if the plugin has "caught up".

For example, if a plugin calls `ConfigDiff#get()` like below, it stopped working.

```
configDiff.get(JsonNode.class, "attributes");
```

The `configDiff` object is based on the core Jackson, and `JsonNode.class` in the first parameter is plugin's Jackson. The core Jackson does not recognize plugin's Jackson.

Plugin developers sometimes had to do like this because the `org.embulk.config.DataSource` interface was poor. There were no way to retrieve an arbitrary tree node from `DataSource`. Then, we had to get the background `JsonNode` from `DataSource` (`ConfigDiff`).

----

This pull request tries to mitigate this problem by making `DataSource` more powerful by itself. It adds methods :

* `DataSource#hasList(String)` to check if an attribute is a list.
* `DataSource#hasNested(String)` to check if an attribute is a nested `DataSource`.
* `DataSource#getListOf(Class<?>, String)` to get listed values with type restriction.
* `DataSource#toMap()` to get the tree reconstructed with `Map`, `List`, and primitives.

----

This is changing (adding) the SPI, so we may want an EEP even for this according to [EEP-1](https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0001.md). Wanted to hear reviewer's thoughts.

### DIFF
--- a/embulk-deps/src/main/java/org/embulk/deps/config/DataSourceImpl.java
+++ b/embulk-deps/src/main/java/org/embulk/deps/config/DataSourceImpl.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import org.embulk.config.ConfigDiff;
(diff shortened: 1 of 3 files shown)


embulk-deps/src/main/java/org/embulk/deps/config/DataSourceImpl.java
```diff
@@ -88,6 +115,22 @@ public <E> E get(Class<E> type, String attrName, E defaultValue) {
...
+        final JsonNode json = this.data.get(attrName);
+        if (json == null) {
+            return Collections.emptyList();
+        }
+        if (!json.isArray()) {
+            throw new ConfigException("Attribute " + attrName + " must be an array");
+        }
+        final ArrayList<E> list = new ArrayList<>();
+        for (final JsonNode element : (Iterable<JsonNode>) () -> json.elements()) {
+            model.readObject(type, element.traverse());
```
vietnguyen-td, line 129: seems missing add this object into `list`?

vietnguyen-td (commented):
general looks good,  left an concern, do we need unit test for these new methods?

embulk-deps/src/main/java/org/embulk/deps/config/DataSourceImpl.java
dmikurube: Oops, exactly. Good catch!

hiroyuki-sato (approved):
LGTM👍

dmikurube:
Yeah, it's a good point that we need unit tests -- it's clear from my mistake. ;)

vietnguyen-td (approved):
LGTM

dmikurube:
Thanks. Let me merge this, and release v0.10.41.

I think I'll write up a retrospective EEP including this to summarize the Jackson removal and the Configuration API.

PR #1540, opened 2023-01-18
Rename JsonValue.Type to JsonValue.EntityType, and JsonValue#is... to JsonValue#isJson...

dmikurube:
It clarifies the context of `JsonValue.(Entity)Type` by adding and updating Javadocs, where `EntityType` should be 1:1 corresponding to the implementation classes of `JsonValue`.

Just `Type` was not very understandable, then wanted to clarify its purpose and context. We define that `(Entity)Type` should be 1:1 corresponding to each implementation class, to be "implementation-specific", not by JSON's value type. A number could be integer, or decimal. The word "entity" comes from that purpose.

In contrast, `javax.json` (https://docs.oracle.com/javaee/7/api/javax/json/JsonValue.html) has had `JsonValue#ValueType` which is JSON's value type. A number is number there.

Embulk's `JsonValue` classes are designed to be mostly compatible with `javax.json`, so that they can "implement" `javax.json` or `jakarta.json` classes in the future. In that case, `javax(jakarta).json.JsonValue.ValueType` would play the role of JSON's value type.

It is a follow-up to #1462.

Its corresponding EEP is draft at #1538.


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1541, opened 2023-01-18
Clarify the equality of JsonValue implementations

dmikurube:
It is a follow-up to: #1462

#1462 was not very clear about equality (`#equals(Object)`). If a developer defines their own `JsonValue` subclass, the new class may behave inconsistent with the "standard" `JsonValue` classes. Some are `true`, some are `false`, and some throw an Exception. The second commit 1f248a9c59e8bceb7b78c3e4b79459399968bc0d tests it, and confirms it fails.

This pull request clarifies the behavior of `#equals(Object)`. The standard `JsonValue` classes should not be "equal" to any non-standard `JsonValue` subclass. The third commit bb3f497bf155ae788c8a4f4a3a43490617a96d70 fixes `#equals(Object)`, and confirms it passes.

Its corresponding EEP is draft at https://github.com/embulk/embulk/pull/1538.

### DIFF
--- a/embulk-api/src/main/java/org/embulk/spi/json/JsonArray.java
+++ b/embulk-api/src/main/java/org/embulk/spi/json/JsonArray.java
@@ -189,6 +189,9 @@ public final class JsonArray extends AbstractList<JsonValue> implements JsonValu
     /**
      * Compares the specified object with this JSON array for equality.
      *
+     * <p>Note that it can return {@code true} only when {@link JsonArray} is given. It checks the equality as a JSON array.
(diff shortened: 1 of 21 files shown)


hiroyuki-sato (approved):
LGTM 👍

Minor: Copyright is 2023?

dmikurube:
`FakeJson....java` are almost copies of the main source codes `Json....java`. `Json....java` were written in 2022.

dmikurube:
Anyway, thanks!

PR #1542, opened 2023-01-20
Rename JsonInteger to JsonLong, and JsonDecimal to JsonDouble, and insert JsonNumber as a common interface of them

dmikurube:
It is another follow-up to: #1462

#1462 named integral numbers as `JsonInteger`, and decimal numbers as `JsonDecimal`. But, after some thoughts, I changed my mind that `JsonLong` and `JsonDouble` could be better.

The reasons are :
* Those implementation classes are coupled more strongly with their internal representations: `long` and `double`. Developers cannot easily imagine that `JsonInteger` is based on `long`, then it has limitations by the type.
* Embulk has had the `LONG` column type and the `DOUBLE` column type. It should be acceptable for Embulk-related developers to use names like `Long` and `Double` in such class names.
* JSON values in JSON columns sometimes need to be interoperable with Embulk's native columns. For example, [`embulk-filter-expand_json`](https://github.com/embulk/embulk-filter-expand_json) expands JSON values into Embulk columns. They need to be named consistently for that purpose.

The discussion is included in the draft EEP: #1538


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1543, opened 2023-01-20
Add JsonValue#presumeReferenceSizeInBytes to presume an approximate size to be occupied in Page

dmikurube:
It is a follow-up to: #1462

`PageBuilder` has had a mechanism to estimate size of each record in order to make decisions when to "flush" the `PageBuilder` to `Page`. But, it has been just assumed as `256` for JSON columns.

https://github.com/embulk/embulk/blob/v0.10.41/embulk-core/src/main/java/org/embulk/spi/PageBuilderImpl.java#L181

Introducing the new `JsonValue` is a good chance to make it better. `JsonValue#presumeReferenceSizeInBytes` is the mechanism to presume the approximate size.

### DIFF
--- a/embulk-api/src/main/java/org/embulk/spi/json/JsonArray.java
+++ b/embulk-api/src/main/java/org/embulk/spi/json/JsonArray.java
@@ -115,11 +115,24 @@ public final class JsonArray extends AbstractList<JsonValue> implements JsonValu
     /**
      * Returns the approximate size of this JSON array in bytes presumed to occupy in {@link org.embulk.spi.Page} as a reference.
      *
+     * <p>This approximate size is used only as a threshold whether {@link org.embulk.spi.PageBuilder} is flushed, or not.
+     * It is not accurate, it does not need to be accurate, and it is impossible in general to tell an accurate size that
+     * a Java object occupies in the Java heap. But, a reasonable approximate would help to keep {@link org.embulk.spi.Page}
+     * performant in the Java heap.
+     *
+     * <p>It is better to flush more frequently for bigger JSON value objects, less often for smaller JSON value objects,
+     * but no infinite accumulation even for empty JSON value objects.
+     *
      * @return the approximate size of this JSON array in bytes presumed to occupy in {@link org.embulk.spi.Page} as a reference
+     *
+     * @see "org.embulk.spi.PageBuilderImpl#addRecord"
      */
     @Override
     public int presumeReferenceSizeInBytes() {
+        // No clear reason for the number 4.
+        // But at least, it should not be 0 so that the approximate size of an empty array would not be 0.
         int sum = 4;
+
         for (int i = 0; i < this.values.length; i++) {
             sum += this.values[i].presumeReferenceSizeInBytes();
         }
--- a/embulk-api/src/main/java/org/embulk/spi/json/JsonArray.java
+++ b/embulk-api/src/main/java/org/embulk/spi/json/JsonArray.java
@@ -113,6 +113,20 @@ public final class JsonArray extends AbstractList<JsonValue> implements JsonValu
     }
 
     /**
+     * Returns the approximate size of this JSON array in bytes presumed to occupy in {@link org.embulk.spi.Page} as a reference.
+     *
+     * @return the approximate size of this JSON array in bytes presumed to occupy in {@link org.embulk.spi.Page} as a reference
+     */
+    @Override
+    public int presumeReferenceSizeInBytes() {
+        int sum = 4;
+        for (int i = 0; i < this.values.length; i++) {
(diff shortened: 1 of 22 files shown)


embulk-api/src/main/java/org/embulk/spi/json/JsonString.java
```diff
@@ -84,6 +84,17 @@ public JsonString asJsonString() {
...
 
+    /**
+     * Returns the approximate size of this JSON string in bytes presumed to occupy in {@link org.embulk.spi.Page} as a reference.
+     *
+     * @return the approximate size of this JSON string in bytes presumed to occupy in {@link org.embulk.spi.Page} as a reference
+     */
+    @Override
+    public int presumeReferenceSizeInBytes() {
+        // Indeed, null is stored in Page as a special form as |nullBitSet|, but considered as 1 here in JsonValue just for ease.
+        return this.value.length() * 2 + 4;
```
hiroyuki-sato, line 95: Minor question why return twice the length?

embulk-api/src/main/java/org/embulk/spi/json/JsonArray.java
```diff
@@ -112,6 +112,20 @@ public JsonArray asJsonArray() {
...
     }
 
+    /**
+     * Returns the approximate size of this JSON array in bytes presumed to occupy in {@link org.embulk.spi.Page} as a reference.
+     *
+     * @return the approximate size of this JSON array in bytes presumed to occupy in {@link org.embulk.spi.Page} as a reference
+     */
+    @Override
+    public int presumeReferenceSizeInBytes() {
+        int sum = 4;
```
hiroyuki-sato, line 122: Minor question why 4-byte? (I know it represents the type of JsonArray. But I didn't know Why four-byte)

hiroyuki-sato (commented):
Could you confirm FakeJson{Array|Object} initial sum size? If It's OK, LGTM

embulk-api/src/test/java/org/embulk/spi/json/FakeJsonArray.java
```diff
@@ -36,6 +36,15 @@ public EntityType getEntityType() {
         return EntityType.ARRAY;
     }
 
+    @Override
+    public int presumeReferenceSizeInBytes() {
+        int sum = 0;
```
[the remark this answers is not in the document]
dmikurube: Ah, I fixed only `JsonArray`, and forgot `FakeJsonArray`. In fact, `FakeJsonArray`'s `presumeReferenceSizeInBytes` is never used, but some implementation is needed to get it extended from `JsonValue`. (Just returning 0 should work for `FakeJsonArray`.)

But anyway, better to fix it to have the same one. Let me fix it.

embulk-api/src/test/java/org/embulk/spi/json/FakeJsonObject.java
```diff
@@ -50,6 +50,15 @@ public EntityType getEntityType() {
         return EntityType.OBJECT;
     }
 
+    @Override
+    public int presumeReferenceSizeInBytes() {
+        int sum = 0;
```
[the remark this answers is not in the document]
dmikurube: Ditto.

embulk-api/src/main/java/org/embulk/spi/json/JsonString.java
dmikurube: It comes from :
> https://github.com/embulk/embulk/blob/v0.10.41/embulk-core/src/main/java/org/embulk/spi/PageBuilderImpl.java#L173

It's a difficult question in general to tell "how many bytes are consumed by a string". It depends on JVM's internal representation, and that could be different per a lot of perspective, including JVM versions. But it would be some reasonable persumption because JVM's internal representation of strings has been UTF-16 at least so far. (It may change, though.)

embulk-api/src/main/java/org/embulk/spi/json/JsonArray.java
dmikurube: Noted in a comment there in e86e09cb7fa3e00e16bd0ca97641ec681c15c9c0

dmikurube (commented):
Thanks for your comments! Addressed in e86e09cb7fa3e00e16bd0ca97641ec681c15c9c0

hiroyuki-sato (approved):
LGTM 👍

embulk-api/src/main/java/org/embulk/spi/json/JsonString.java
dmikurube: Ah, indeed, Strings are not always UTF-16 already since Java 9. ;)

https://openjdk.org/jeps/254

dmikurube:
Thanks!

PR #1544, opened 2023-01-20
Remove JsonValue#equals because it's in java.lang.Object

dmikurube:
It is a follow-up to: #1462

It's simple. As `Object#equals` exists already, we don't need `JsonValue#equals`. It came from `equals` in msgpack-java's `Value`.


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1545, opened 2023-01-20
Define a public constant NULL in JsonNull

dmikurube:
It is a follow-up to: #1462 

It is just to make `JsonNull` consistent with `JsonBoolean` which has `FALSE` and `TRUE` as constants.


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1547, opened 2023-01-22
Stop String#intern when instantiating JsonString

dmikurube:
Follow-up to: #1462 

I at first used `String#intern()` forcibly when creating `JsonString` to save memory, but at the same time, it needs time to `String#intern()`.

It should be plugin developer's choice. Going to remove the `intern()`.


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1548, opened 2023-01-22
Represent JsonObject's internal keys by String, not JsonString

dmikurube:
Follow-up to: #1462

`JsonObject` started from a single array containing both keys and values because it started from a clone of msgpack-java's [`ImmutableMapValueImpl`](https://github.com/msgpack/msgpack-java/blob/v0.8.24/msgpack-core/src/main/java/org/msgpack/value/impl/ImmutableMapValueImpl.java).

But, unlike msgpack-java's `MapValue` taking any `Value` as a key, `JsonObject` takes only strings as a key. It would be more reasonable and straightforward to have `JsonObject`'s keys as plain `String`s.

It does not change any behavior observed from outside. No changes in tests, then.

### DIFF
--- a/embulk-api/src/main/java/org/embulk/spi/json/JsonObject.java
+++ b/embulk-api/src/main/java/org/embulk/spi/json/JsonObject.java
@@ -18,7 +18,6 @@ package org.embulk.spi.json;
 
 import java.util.AbstractMap;
 import java.util.AbstractSet;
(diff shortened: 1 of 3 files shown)


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1549, opened 2023-01-23
Fix cases of JsonDouble, and reject NaN and the infinity

dmikurube:
Follow-up to: #1462

First, `JsonDouble` had a bug that it checked in `*ValueExact()` in an opposite way. Changed them to use `is*Value()` to check. Added tests for it and more first, then fixed it.

Second, `JsonDouble` accepted NaN and the infinity although the original JSON format does not have definitions of NaN or the infinity. `BigInteger` and `BigDecimal` do not support NaN nor the infinity on the other hand, then `#bigIntegerValue(Exact)` and `#bigDecimalValue` would have needed some special handling. After some thoughts, we concluded that `JsonDouble` should not accept NaN nor the infinity.

Embulk's `DOUBLE` column type accepts NaN and the infinity, indeed. They would not be interoperable at that point, but introducing NaN or the infinity in JSON would be more confusing. We may revisit it when we receive a strong request for NaN and/or the infinity.

### DIFF
--- a/embulk-api/src/test/java/org/embulk/spi/json/TestJsonDouble.java
+++ b/embulk-api/src/test/java/org/embulk/spi/json/TestJsonDouble.java
@@ -18,12 +18,14 @@ package org.embulk.spi.json;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.lang.reflect.Modifier;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
 import org.junit.jupiter.api.Test;
 
 public class TestJsonDouble {
@@ -31,6 +33,17 @@ public class TestJsonDouble {
     public void testFinal() {
         // JsonDouble must be final.
         assertTrue(Modifier.isFinal(JsonDouble.class.getModifiers()));
+    }
+
+    @Test
+    public void testNaN() {
+        assertThrows(ArithmeticException.class, () -> JsonDouble.of(Double.NaN));
+    }
+
+    @Test
+    public void testInfinity() {
+        assertThrows(ArithmeticException.class, () -> JsonDouble.of(Double.NEGATIVE_INFINITY));
+        assertThrows(ArithmeticException.class, () -> JsonDouble.of(Double.POSITIVE_INFINITY));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


embulk-api/src/test/java/org/embulk/spi/json/TestJsonDouble.java
```diff
@@ -77,4 +90,718 @@ public void testBasicDoubleValue() {
...
+        assertEquals((short) -1, jsonDouble.shortValue());  // Overflow
+        assertThrows(ArithmeticException.class, () -> jsonDouble.shortValueExact());
+        assertEquals((int) 2147483647, jsonDouble.intValue());  // Overflow
+        assertThrows(ArithmeticException.class, () -> jsonDouble.intValueExact());
+        assertEquals(9223372036854775807L, jsonDouble.longValue());  // Overflow
+        assertThrows(ArithmeticException.class, () -> jsonDouble.longValueExact());
+        assertEquals(new BigInteger("9223372036854776000000"), jsonDouble.bigIntegerValue());
+        assertEquals(new BigInteger("9223372036854776000000"), jsonDouble.bigIntegerValueExact());
+        assertEquals((float) 9.223372036854776E21, jsonDouble.floatValue(), 0.0001);
+        assertEquals(9.223372036854776E21, jsonDouble.doubleValue(), 0.0001);
```
hiroyuki-sato, line 700: Minor: Just for my study: Is this value(`0.0001`) based on something specification?

embulk-api/src/test/java/org/embulk/spi/json/TestJsonDouble.java
```diff
@@ -77,4 +90,718 @@ public void testBasicDoubleValue() {
...
+        assertThrows(ClassCastException.class, () -> jsonDouble.asJsonObject());
+        assertEquals(8, jsonDouble.presumeReferenceSizeInBytes());
+        assertTrue(jsonDouble.isIntegral());
+        assertFalse(jsonDouble.isByteValue());
+        assertFalse(jsonDouble.isShortValue());
+        assertFalse(jsonDouble.isIntValue());
+        assertFalse(jsonDouble.isLongValue());
+        assertEquals((byte) -1, jsonDouble.byteValue());  // Overflow
+        assertThrows(ArithmeticException.class, () -> jsonDouble.byteValueExact());
+        assertEquals((short) -1, jsonDouble.shortValue());  // Overflow
```
hiroyuki-sato, line 691: Minor: Do you know any reference why `(double)9_223_372_036_854_775_807_000.0` cast to (short)-1?
It seems that the `(double)9223372036854775807000.0` value encode to `100010001111111010000000000000000000000000000000000000000000000`
I didn't know why it encode to `-1`.

hiroyuki-sato (approved):
LGTM 👍

I'll appreciate it if you comment on my questions.

embulk-api/src/test/java/org/embulk/spi/json/TestJsonDouble.java
dmikurube: Java's primitive conversion has some complex rules. Nice to have a look at: https://docs.oracle.com/javase/specs/jls/se7/html/jls-5.html#jls-5.1.3

embulk-api/src/test/java/org/embulk/spi/json/TestJsonDouble.java
dmikurube: Not specially. That `0.0001` is an acceptable calculation eplison ("error"; 誤算) to compare the floating point numbers. It is just a "small number" to accept some calculation errors.

dmikurube (commented):
Thanks!

hiroyuki-sato:
Thank you for your comment!

dmikurube:
Thanks. Merging this!

PR #1550, opened 2023-01-23
Add JsonObject#of(...) methods

dmikurube:
Follow-up to: #1462

It just adds a lot of convenient creator methods `JsonObject.of(String, JsonValue, ...)` that are similar to `Map` in Java 9+.

https://docs.oracle.com/javase/9/docs/api/java/util/Map.html


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1551, opened 2023-01-23
Check null specified for JsonObject

dmikurube:
Follow-up to: #1462

`JsonObject` should not be created with `null`.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1552, opened 2023-01-23
Check null specified for JsonString

dmikurube:
Follow-up to: #1462

`JsonString` should not be created with `null`.


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1554, opened 2023-01-25
Change JsonValue classes to back with msgpack-java classes

dmikurube:
This is the step 3 in the draft EEP: JSON Column Type: #1538 

> "Modify the new JSON classes to represent their inner values by msgpack-java"

`#toMsgpack()` is used only for running with the existing methods in `Page`, `PageBuilder`, `PageReader`. It will be reverted when we finally remove msgpack-java from the Embulk SPI after v1.0.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1557, opened 2023-01-27
Add some Javadoc and some tests for JsonString

dmikurube:
Follow-up to: #1462

Some addition in Javadoc, and add some test cases that are exampled in the EEP Draft: #1538

### DIFF
--- a/embulk-api/src/test/java/org/embulk/spi/json/TestJsonString.java
+++ b/embulk-api/src/test/java/org/embulk/spi/json/TestJsonString.java
@@ -140,6 +140,15 @@ public class TestJsonString {
         assertFalse(string.equals(FakeJsonString.of(string.getString())));
     }
 
+    @SuppressWarnings("checkstyle:IllegalTokenText")
+    @Test
+    public void testExample() {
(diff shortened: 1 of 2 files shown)


embulk-api/src/test/java/org/embulk/spi/json/TestJsonString.java
```diff
@@ -140,6 +140,15 @@ public void testEscape() {
         assertFalse(string.equals(FakeJsonString.of(string.getString())));
     }
 
+    @SuppressWarnings("checkstyle:IllegalTokenText")
+    @Test
+    public void testExample() {
```
hiroyuki-sato, line 145: Minor: It seems reasonable the name `testLiteral` or something instead of `testExample`.

hiroyuki-sato (approved):
LGTM 👍
I left a little comment.

embulk-api/src/test/java/org/embulk/spi/json/TestJsonString.java
dmikurube: Thanks for your comment. Actually, these cases are "examples" contained in the EEP Draft: #1538, then the `Example` intended that.

PR #1559, opened 2023-01-31
Let JsonLong and JsonDouble can be "equal" when JsonDouble is integral

dmikurube:
Follow-up to: #1462

In terms of the `JSON` context, `JsonLong.of(12345L)` should be "equal" to `JsonDouble.of(12345.0)` because both are "JSON numbers".


civitaspo (approved):
I've reviewed that the comparison condition is fine because `isIntegral` is called inside `isLongValue`. Tests are also fine.

dmikurube:
Thanks!

PR #1563, opened 2023-02-21
Upgrade embulk-api's slf4j-api to 2.0.6, and executable's logback-classic to 1.3.5

dmikurube:
Let's update SLF4J and Logback to the (available) latest before we go into Embulk v0.11!

SLF4J v2.0 should basically have a compatible interface with SLF4J v1.7, just with :
* Fluent logging interface in addition, and
* Better support for Java 9+ modules (Jigsaw).

Logback needs to be updated to v1.3 to follow. (I believe that v1.3 will work with Java 9+ whiel v1.4 does not work with Java 8.)

Plugins should be depending on `embulk-api` just with `compileOnly`. The actual SLF4J and Logback artifacts are bundled only with the Embulk executable binary. The dependency on `embulk-api` from plugins should not be a big problem.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -221,7 +221,13 @@ dependencies {
     implementation project(':embulk-core')
 
     // Logback and jansi are included only in the executable package. (jansi for logback colors to work on Windows.)
-    implementation "ch.qos.logback:logback-classic:1.2.3"
+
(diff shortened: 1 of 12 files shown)


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks! Let me merge this, and release v0.10.43.

PR #1566, opened 2023-03-13
Merge embulk-api to embulk-spi

dmikurube:
We have been separated `embulk-api` and `embulk-spi` like following.

* `embulk-api` contains accessors for Embulk core features. Utility libraries for Embulk plugins may also depend on `embulk-api` to help plugins accessing Embulk core features.
* `embulk-spi` consists of interfaces and abstract classes for Embulk plugins to inherit.

But, as far as we observed so far, this separation has not contributed to better experiences, but just increased a little bit labor.

Before going into v0.11, and also before splitting `embulk-spi` to a separate Git repository, let's merge `embulk-api` into `embulk-spi` for simplicity.


hiroyuki-sato (approved):
LGTM 👍

dmikurube:
Thanks!

PR #1570, opened 2023-03-24
Upgrade the standard plugins to 0.10.40

dmikurube:
`embulk-guess-csv` started to use `embulk-util-guess:0.3.0` with a copy of ICU4J from 0.10.40. By this upgrade, Embulk will be able to remove ICU4J from the core (`embulk-deps`).


dmikurube:
Thanks!

PR #1574, opened 2023-03-25
Remove shared build configurations among build.gradle in subprojects

dmikurube:
Build configurations have been shared among subprojects by `subprojects { ... }`. But, it has made the build configurations complicated, and hard to understand.

As the first step to simplify, this pull request removes the share although it will cause some duplications.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1576, opened 2023-03-27
Remove Jansi

dmikurube:
Jansi had been included in Embulk since v0.6.12 to "color" Logback logs even in Windows.
See commit: 34d238494e05231ecb5fe5e10910119035767492

We, however, do not see a big benefit to keep it. It is not critical, but taking a certain amount in Embulk's executable size. Let's remove it.

We may get it back if we receive many requests to do it again in the future.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,15 +45,11 @@ configurations {
 dependencies {
     implementation project(':embulk-core')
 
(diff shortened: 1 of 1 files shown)


hiroyuki-sato (approved):
LGTM👍

hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1577, opened 2023-03-27
Remove JLine

dmikurube:
JLine has been included in Embulk since v0.10.20 along with refactoring the command line parser.
See pull request: #1333

We, however, do not see a big benefit to keep it. It is not critical, but taking a certain amount in Embulk's executable size. Let's remove it.

We may get it back if we receive many requests to do it again in the future.


dmikurube:
Thanks!

PR #1589, opened 2023-04-04
Remove conflicts in the executable all-in-one JAR file

dmikurube:
`META-INF/INDEX.LIST` and `META-INF/versions/9/module-info.class` have conflicted in the executable all-in-one JAR file. Gradle 7+ started to fail when there is a conflict in building a JAR file.

Better not to have a conflict in the JAR file, anyway.

They can be just removed while Embulk basically expects Java 8.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1590, opened 2023-04-05
Reimplement preview without Jackson in embulk-core

dmikurube:
`org.embulk.deps.*` packages in `embulk-core` may conflict with `embulk-deps` for the same reason with #1589. Wanted to move `org.embulk.deps.*` classes in `embulk-core` to other packages.

Along with that, I noticed that `PreviewPrinter` does not essentially need Jackson. It can be implemented practically purely in `embulk-core`. Then this pull request moves them to `org.embulk.exec` in `embulk-core` without Jackson.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1591, opened 2023-04-05
Move delegate shim classes in embulk-core for embulk-deps out of org.embulk.deps

dmikurube:
`org.embulk.deps.*` packages in `embulk-core` may conflict with `embulk-deps` for the same reason with #1589 and #1590. Wanted to move `org.embulk.deps.*` classes in `embulk-core` to other packages.

After #1590, this pull request moves other `org.embulk.deps.*` classes to appropriate packages.


embulk-core/src/main/java/org/embulk/spi/json/JsonParserDelegate.java
```diff
@@ -7,8 +7,8 @@
 import org.embulk.deps.EmbulkDependencyClassLoaders;
 import org.msgpack.value.Value;
 
-public abstract class DepsJsonParser {
-    public static DepsJsonParser of() {
+public abstract class JsonParserDelegate {
+    public static JsonParserDelegate of() {
         try {
```
hiroyuki-sato, line 12: The implementation class is inside `embulk-spi`. So, this class added the suffix `Delegate`. Correct?
It is just confirmation about the naming rule. (Class naming is my weak point.)

hiroyuki-sato (approved):
LGTM👍 I have a question.

embulk-core/src/main/java/org/embulk/spi/json/JsonParserDelegate.java
dmikurube: IIUC, its implementatino is not inside `embulk-spi`?

This `JsonParserDelegate` is not provided for plugins, so actually, any name is okay. Let be explain some backgrounds.

* In the past (-v0.9), `org.embulk.spi.json.JsonParser` had been provided for plugins from `embulk-core`. Some plugins used `org.embulk.spi.json.JsonParser`.
* From v0.10, we decided to provide such a utility feature as an external library, not from `embulk-core`. This `JsonParser` is now in `[`embulk-util-json`](https://github.com/embulk/embulk-util-json)`. New (caught-up) plugins should already be using `embulk-util-json`, instead of `org.embulk.spi.json.JsonParser`.
* But, if we remove `org.embulk.spi.json.JsonParser` now, older plugins are easily impacted immediately. To avoid such an impact, we still have `org.embulk.spi.json.JsonParser`, but its implementation is based on `embulk-util-json`, located in `embulk-deps` behind the hidden class loader.
* This `JsonParserDelegate` is a bridge to `embulk-util-json`'s `org.embulk.util.json.JsonParser`.
* In summary, old plugins may call `org.embulk.spi.json.JsonParser`, and the call is delegated like:
    * `org.embulk.spi.json.JsonParser`
    * `org.embulk.spi.json.JsonParserDelegate`
    * `org.embulk.spi.json.JsonParserDelegateImpl` (in `embulk-deps`)
    * `org.embulk.util.json.JsonParser` (in `embulk-util-json`)

hiroyuki-sato: Thank you for your comment! This summary very useful for me.

PR #1592, opened 2023-04-06
Have singleton org.embulk.EmbulkDependencyClassLoader merged from EmbulkDependencyClassLoaders and DependencyClassLoader

dmikurube:
Wanted not to have classes under `org.embulk.deps` in `embulk-core`, for the same motivation with #1589, #1590, and #1591.

At the same time,
* There **were** multiple `DependencyClassLoader`s when we had multiple `embulk-deps-*`. But now, there **is** only `embulk-deps`, and there **is** only a single `DependencyClassLoader`. The structure is redundant that `EmbulkDependencyClassLoaders` maintain only a single instance of `DependencyClassLoader`.
* `(Embulk)DependencyClassLoader(s)` could be used from some `EmbulkEmbed` users. Such classes have been located at `org.embulk` prefixed with `Embulk-`. (Such as `EmbulkEmbed`, `EmbulkSystemProperties`, ...)

Then, this pull request merges `org.embulk.deps.EmbulkDependencyClassLoaders` and `org.embulk.deps.DependencyClassLoader` into a singleton `org.embulk.EmbulkDependencyClassLoader`.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1593, opened 2023-04-06
Move classes about self-contained JAR files to org.embulk.cli

dmikurube:
Wanted not to have classes under org.embulk.deps in embulk-core, for the same motivation with #1589, #1590, #1591, and #1592.

"Self-contained JAR files" are used for the all-in-one executable JAR file, then those classes are used with CLI, therefore nice to locate them in `org.embulk.cli`.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1594, opened 2023-04-06
Move all embulk-junit4 classes to org.embulk.test

dmikurube:
In the end of package movings with #1589, #1590, #1591, and #1592, wanted not to have classes of `embulk-junit4` in the same packages with other artifacts.

Then, moving `org.embulk.*` and `org.embulk.spi.*` classes in `embulk-junit4` into `org.embulk.test`. It would break tests in many plugins when they start using `org.embulk:embulk-junit4:0.10.47`, but they're just tests. No production releases broken.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1609, opened 2023-04-17
Remove config validations with BVal from org.embulk.config and org.embulk.deps.config

dmikurube:
BVal has been used for validating configurations with Embulk core's embedded config processor. But, BVal uses JAXB in it. Initializing BVal in embulk-core/embulk-deps makes Embulk not to work with Java 9+, especially with Java 11+.

Embulk v0.11+ does basically expect "caught-up" plugins. The "caught-up" plugins should basically process configurations with their own embulk-util-config, not with Embulk core's embedded config processor.

Therefore, in v0.10.49 just before v0.11, they can be removed from embulk-core/embulk-deps.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1611, opened 2023-04-20
Draft EEP: Expose Task report

legiangthanh:
This EEP proposes a change to export task report information to outside.

### DIFF
--- a/docs/eeps/draft-expose-task-report.md
+++ b/docs/eeps/draft-expose-task-report.md
@@ -1,0 +1,192 @@
+---
+EEP: unnumbered
+Title: Expose Task report
+Author: thanhle
+Status: draft
+Type: Standards Track
+Created: 2023-04-14
+---
+
+Introduction
+=============
+
+Currently, `TaskReport` information is only used internally by `InputPlugin` and `OutputPlugin`. This EEP propose a change to expose `TaskReport` to outside also add a new SPI, change in `Formatter` and `Parser` help them able to provide a TaskReport information.
+
+Motivation
+===========
+
+The TaskReport now mainly is used by `InputPlugin` and `OutputPlugin` for creating a `ConfigDiff`. Moreover, the Formatter and Parser Plugin can't provide TaskReport.
+
+This behavior limit plugin and application able to add and get these reports relate to task, e.g. The InputPlugin can provide statistical information of task such as num of records, num of error, file size etc.
+
+Plugin SPI Definitions
+=======================
+
+This section explains the additional Embulk plugin SPI definition
+
+Definitions
+------------
+
+### Changing in interface `ParserPlugin`
+
+```java
+package org.embulk.spi;
+public interface ParserPlugin {
+    // ... (snip) ...
+    /**
+     * Runs each parsing task and return TaskReport
+     *
+     * @param taskSource  a configuration processed for the task from {@link ConfigSource}
+     * @param schema  {@link Schema} to be parsed to
+     * @param input  {@link FileOutput} that is read from a File Input Plugin, or a Decoder Plugin
+     * @param output  {@link PageOutput} to write parsed input so that the input is read from an Output Plugin, or
+     *     another Filter Plugin
+     * @return the {@link TaskReport} in {@link Optional}
+     */
+    default Optional<TaskReport> runWithTaskReport(TaskSource taskSource, Schema schema, FileInput input, PageOutput output) {
+        this.run(taskSource, schema, input, output);
+        return Optional.empty();
+    }
+}
+```
+
+Changing in Embulk
+=======================
+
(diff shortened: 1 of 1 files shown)


dmikurube (changes requested):
At first, please follow the steps written in [EEP-1](https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0001.md). I saw you submitted the idea in https://github.com/orgs/embulk/discussions/1, and I left some comments there. But the discussion does not seem over yet, and nobody is nominated as a sponsor.

> Following the initial discussion above, the workflow varies based on whether any of the EEP's co-authors are in the core committers. If one or more of the EEP's co-authors are core committers, they are responsible for following the process outlined below. Otherwise (i.e. none of the co-authors are core committers), then the EEP author(s) will need to find a sponsor for the EEP.
> https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0001.md

Second, at least address the points we discussed in https://github.com/orgs/embulk/discussions/1. For example,
* Formatter
    * You mentioned about Formatter in the text, but nothing in SPI changes?
* Method name
    * Discussed, but no conclusion yet. Do you want to go with `runWithTaskReport`? How's my counter proposal?
* The sturecture of `TaskReport`, and which component is responsible?
    * Discussed, but no conclusion yet.
* Decoder/Encoder and Filters,
    * I said in the Discussions that "we may not implement it for decoders/encoders", but also said "include the discussions (listing possible approaches, and pros/cons for each) in your upcoming EEP".
    * Imported the Discussion into the draft EEP, but may not be sufficient. This part would affect the future "structure of `TaskReport`, and which component would be responsible (above). Then, you'll need to leave an expected `TaskReport` structure that includes consideration about Decoder/Encoder, and Filters.

Last, just a note, changes in `embulk-spi` are mandatory, but you don't need to include the "code" changes out of `embulk-spi`. You don't need to remove them, but you'll want to explain more narratively about "how things are processed", not just by the code. The code out of embulk-core could change anytime, then explaining by the code does not work.

legiangthanh:
Thank @dmikurube  for your comment.
> At first, please follow the steps written in [EEP-1](https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0001.md). I saw you submitted the idea in https://github.com/orgs/embulk/discussions/1, and I left some comments there. But the discussion does not seem over yet, and nobody is nominated as a sponsor.
Sorry, I think the discussion is finished. Could you please give me instructions on finishing a discussion and asking a sponsor? 

> 
> Formatter
> You mentioned about Formatter in the text, but nothing in SPI changes?
ForrmaterPlugin will create `PageOutput`. I think we don't need to change the SPI of `PageOutput`, I will check if Formatter returns a `TransactionalPageOutput` to get a TaskReport by calling a commit method. You can view it at https://github.com/embulk/embulk/pull/1612/files#diff-4b50d5a981be1817f1d21ae524996cc0e8a62f6db12545a8879a0e54f02e3d4aR168-R177

> Method name
> Discussed, but no conclusion yet. Do you want to go with runWithTaskReport? How's my counter proposal?
Sorry, I miss to update it in EEP. I will update it later.

> The sturecture of TaskReport, and which component is responsible?
> Discussed, but no conclusion yet.
I just give a comment on the discussion. The `FileInput/OutputRunner` will get a TaskReport from Parser/Decoder(TransactionalPageOutput) and create a new attribute with the name `parser` or `formatter`.

> Decoder/Encoder and Filters,
> I said in the Discussions that "we may not implement it for decoders/encoders", but also said "include the discussions (listing possible approaches, and pros/cons for each) in your upcoming EEP".
> Imported the Discussion into the draft EEP, but may not be sufficient. This part would affect the future "structure of TaskReport, and which component would be responsible (above). Then, you'll need to leave an expected TaskReport structure that includes consideration about Decoder/Encoder, and Filters.
> Last, just a note, changes in embulk-spi are mandatory, but you don't need to include the "code" changes out of embulk-spi. You don't need to remove them, but you'll want to explain more narratively about "how things are processed", not just by the code. The code out of embulk-core could change anytime, then explaining by the code does not work
I will update EEP as your comment.

dmikurube:
Code changes in pull requests are "supportive" information for proposals, but they're just helping, not any main contents of proposals. The code base always changes. Explanations only by code may not help in the future.

An EEP is a document to be kept for future maintainers, not only to make a quick change for you just now. Consider it mandatory to explain your proposal, both in Discussions and in an EEP, in text, not just by code. It means that "I made a change in the latest code ..." is not a good explanation.

> Could you please give me instructions on finishing a discussion and asking a sponsor?

No formal process/instructions. But you could just ask me (and else) in the Discussion simply like: "Can I finish the Discussion, and move forward to EEP? Can anybody sponsor my proposal?"

PR #1615, opened 2023-05-15
Upgrade standard plugins to v0.10.42, including the fix for explicit null in CSV quote and escape

dmikurube:
Activating this update https://github.com/embulk/embulk-standards/pull/43 in the Embulk release.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1621, opened 2023-06-15
EEP: JRuby as Optional

dmikurube:
This is an "retrospective" EEP (Embulk Enhancement Proposal) for the decision to make JRuby to be just optional in Embulk, including the decision to remove JRuby from Embulk's executable distribution.


dmikurube:
This is the last one planned to finish in June. ;)

hiroyuki-sato:
I'm still reviewing this PR.

How about writing the liquid template future?
It is a heavy dependency on JRuby.

dmikurube:
Hmm, we actually have no any decision about Liquid for the future as of now. And, Liquid is just Liquid. Embulk is just simply calling it.

I have no idea what to write about Liquid, to be honest. Do you have any specific topic about Liquid?

hiroyuki-sato:
I mean how about writing "no any decision about Liquid for the future as of now"?

I think a reader worries about Liquid future. 

Embulk uses a Liquid template engine. It depends on JRuby. It is necessary to replace an alternate way, but we actually have no any decision about Liquid for the future as of now.

dmikurube:
I basically don't think so.

* This EEP just explains making JRuby "optional", not to be deprecated, nor removed. Liquid is not impacted.
* EEPs are basically for (core and plugin) developers, not for users. The Liquid integration is only for users.

When we write another EEP for templating configurations in the future, Liquid would be mentioned there. But here in this EEP, I feel that Liquid would be just a noise against the main topic.

hiroyuki-sato (approved):
Thanks! I got it. LGTM👍

dmikurube:
Thanks! Let's merge it as EEP-6 as Final.

headius:
I know this is already merged, but I'd love to work with you to provide better JRuby support going forward. Development has increased significantly over the last 2 years, and the latest JRuby runs most benchmarks much faster than previous versions. We know of at least one large customer who depends on the JRuby support.

hiroyuki-sato:
Hello, @headius. Thank you for your comment.

Embulk v0.11 runs without JRuby, but some plugins still use JRuby.
Embulk ecosystem needs plugin maintenance consistently.
We need many helpers, and [we are looking for long-time maintainers](https://www.embulk.org/articles/2025/04/11/dmikurube-is-stepping-down.html).
But no one has yet come forward to be a maintainer.

Additionally, [@dmikurube is stepping down... in a part](https://www.embulk.org/articles/2025/04/11/dmikurube-is-stepping-down.html)

It's touch and go. (I want to use JRuby as a plugin development language)

headius:
@dmikurube @hiroyuki-sato I would love to help support this effort from the JRuby perspective! Perhaps you can explain what's involved in getting JRuby support back into Embulk, and what work is required to keep it that way?

hiroyuki-sato:
Hello, @headius.
 If I have a problem with JRuby, I'll ask you for help like this. 
[[Q] About Found multiple supers in java ctor](https://github.com/jruby/jruby/issues/7241)

The plugin developer can use JRuby as a plugin language for a while.
We are still looking for long-term maintainers for Embulk itself.

dmikurube:
Thanks @headius for your help!

Indeed, the reason why we made JRuby "optional" in Embulk was not the speed of JRuby, but it was hard for us to keep Embulk being maintained to run upon "2 platforms" -- Java and (J)Ruby -- at the same time, with many Embulk plugins developed by outside developers.

Supporting multiple platforms would make the complexity of investigations exponentially. For example, we sometimes have to bump up the base Java (like this time 8 to later), then we have to investigate any possible problem there with JRuby versions. We also have to bump up JRuby as well, the we also have to investigate the combinations of Java versions.

Therefore, we made Embulk to be able to run only on Java, and made JRuby "optional". However, note that it does not mean we dropped the JRuby support. It still runs with JRuby for Ruby-based plugins, then I'd say that it's still "supported".

On the contrary, users achieved the benefit of "choosing their preferred JRuby version" by themselves, by making JRuby optional, which was realized by loading JRuby dynamically. In conclusion, I have no intention to change this dynamic-loading architecture and no intention to get it back to "mandatory", not optional.

----

On the other hand, however, I have to say that the option to "drop" JRuby is still on our table. The reason is because :

1. As written in the post mentioned by @hiroyuki-sato, I left the company, and I'd spend only part of my personal time on Embulk.
2. The difficulty for us is not the performance of JRuby, but almost always the compatibility between JRuby versions. This "compatibility" is not only about the compatibility as the Ruby language, but including the compatibility of the Java-Ruby interface. We have no bandwidth to investigate such the compatibility problems in the future.

So, I'd say that fine-grained up-to-date explaination and announcement about any possible incompatibility in JRuby would help us to keep JRuby supported in Embulk.

dmikurube:
> I'd also like to understand the status of Embulk. Is anyone actively working on it these days?

Not very active, indeed. By chance, we were about to start a kind of discussions and/or decision-making about it.

PR #1635, opened 2023-09-25
Split dependency handling for embedded deps and each embedded plugin

dmikurube:
Until v0.11.0, the dependency handling of embedded libraries in executable JAR was not perfect. If different plugins have different libraries, their versions were aligned to one version. It might potentially make some embedded plugins broken.

This pull request is to split dependency handling for each. It would allow all embedded plugins to work with their expected dependencies.

As a side effect of this change, it turned out that `embulk-guess-csv_all_strings` had a little bit older dependencies than others, indeed... ;)  (See `gradle.lockfile`.)


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1636, opened 2023-09-26
Upgrade standard plugins to use org.embulk.spi.json.JsonValue, instead of org.msgpack.value.Value

dmikurube:
It upgrades embulk-util-json in embulk-deps along with standard plugins.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1641, opened 2023-11-17
Warn against use of org.embulk.spi.json.JsonParser

dmikurube:
`org.embulk.spi.json.JsonParser` is deprecated, and we may remove it in the future. Let's dump a duplication message for use of that.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1642, opened 2023-11-17
Unuse embulk-util-json's JsonParser, instead have the implementation in embulk-deps

dmikurube:
We externalized the implementation of `org.embulk.spi.json.JsonParser` as [`embulk-util-json`](https://github.com/embulk/embulk-util-json), then we have used it in `embulk-deps`. 

However, `embulk-util-json` will switch to `JsonValueParser` with `JsonValue`, and its latest version requires the latest Jackson. On the other hand, `JsonParser` in `embulk-deps` would not be expanded as it is deprecated.

Then, we get the implementation back in `embulk-deps`.


hiroyuki-sato (approved):
LGTM👍

PR #1654, opened 2024-03-12
EEP: Class Loading, and Executable JAR File Structure

dmikurube:
This is a new "retrospective" EEP to explain the new class loading mechanism and the executable JAR file structure introduced throughout the Embulk v0.10 series.

Read the rendered version at:

https://github.com/embulk/embulk/blob/eep-class-loading/docs/eeps/draft-class-loading.md


docs/eeps/draft-class-loading.md
```diff
@@ -0,0 +1,524 @@
...
+```
+
+The `embulk-core` and `embulk-deps` artifacts are tightly coupled by their nature. Different versions of them must not be paired. They are separated only to be loaded by different class loaders.
+
+Embulk's executable JAR file
+-----------------------------
+
+Almost all the Embulk users use Embulk's packaged executable JAR file while it is possible for users to embed `embulk-core` and `embulk-deps` by themselves. The next question is how to structure the executable JAR file.
+
+The hidden dependency libraries and `embulk-deps` cannot be extracted into the executable JAR file so that they are not loaded by the top-level class loader. Instead, we decided to include these library JAR files "as-is" nested in the executable JAR file. This is called the "JAR-in-JAR" technique, traditionally known for [One-JAR](https://one-jar.sourceforge.net/).
```
hiroyuki-sato, line 313: What do you think about adding a built-in plugin explanation?

ex)
we decided to include these library
-> 
we decided to include these libraries(include built-in plugins)


IIUC, `*.jar` files contain (1) embulk-deps.jar itself, (2) hidden dependency libraries which load by embulk-deps (3) built-in plugins.

docs/eeps/draft-class-loading.md
```diff
@@ -0,0 +1,524 @@
...
+
+Second, this mechanism applies to each class, not each dependency library artifact JAR file. Therefore, a mix-up can occur within a single library. For example, recent Guava 30.0+ has a new class `com.google.common.util.concurrent.ClosingFuture`, while Guava 18.0 used in the Embulk core (up to v0.10.38) did not. If a plugin includes Guava 30.0+, `ClosingFuture` is loaded from that Guava 30.0+. `ClosingFuture` uses another Guava class, `com.google.common.collect.ImmutableList`. `ImmutableList` is loaded from Guava 18.0 in Embulk v0.10.38 or earlier because it existed in Guava 18.0 and `com.google.common` was included in `parent_first_packages` of Embulk v0.10.38 or earlier. When `ClousingFuture` called a new method in `ImmutableList` that was added after Guava 18.0, `NoSuchMethodError` would be thrown.
+
+Because of these two problems, a different mechanism for isolating library artifact JAR files other than `parent_first_packages` was needed.
+
+The `parent_first_resources` file is a similar mechanism for non-class resources, which has the same problems.
+
```
hiroyuki-sato, line 171: `embulk-deps` replaced `parent_first_packages` and `parent_first_resources` mechanisms?
Or It still uses in v0.11.x?

hiroyuki-sato (approved):
LGTM👍 
I left two minor comments.


Sorry, I wasn't feeling well 😢

docs/eeps/draft-class-loading.md
dmikurube: Sounds reasonable. I added 

> (as well as the built-in plugins explained below)

in the sentense.

----

> IIUC, `*.jar` files contain (1) embulk-deps.jar itself, (2) hidden dependency libraries which load by embulk-deps (3) built-in plugins.

and (4) dependency libraries that are used from built-in plugins, too.

(More precisely, "hidden dependency libraries which load by embulk-deps" is not 100% correct. Thease hidden libraries are loaded "with" embulk-deps. embulk-deps does not have any class loading feature by itself.)

docs/eeps/draft-class-loading.md
dmikurube: Added the following sentense.

> The `parent_first_packages` and `parent_first_resources` mechanisms still remain in Embulk as of v0.11, but they will go away once they are no longer needed.

dmikurube (commented):
@hiroyuki-sato Thanks! Addressed your comments and updated the draft.

dmikurube:
> Sorry, I wasn't feeling well 😢

Sorry to hear that. Take care...

dmikurube:
Thanks! Finalizing this as EEP-9 to be Final.

PR #1661, opened 2024-05-14
Upgrade built-in plugins that depend on Jackson 2.16.2

dmikurube:


gradle.lockfile
```diff
@@ -3,18 +3,25 @@
...
+com.fasterxml.jackson.core:jackson-annotations:2.16.2=embulk-decoder-bzip2,embulk-decoder-gzip,embulk-encoder-bzip2,embulk-encoder-gzip,embulk-filter-remove_columns,embulk-filter-rename,embulk-formatter-csv,embulk-guess-bzip2,embulk-guess-csv,embulk-guess-csv_all_strings,embulk-guess-gzip,embulk-guess-json,embulk-input-config,embulk-input-file,embulk-output-file,embulk-output-null,embulk-output-stdout,embulk-parser-csv,embulk-parser-json
+com.fasterxml.jackson.core:jackson-annotations:2.6.7=deps
+com.fasterxml.jackson.core:jackson-core:2.16.2=embulk-decoder-bzip2,embulk-decoder-gzip,embulk-encoder-bzip2,embulk-encoder-gzip,embulk-filter-remove_columns,embulk-filter-rename,embulk-formatter-csv,embulk-guess-bzip2,embulk-guess-csv,embulk-guess-csv_all_strings,embulk-guess-gzip,embulk-guess-json,embulk-input-config,embulk-input-file,embulk-output-file,embulk-output-null,embulk-output-stdout,embulk-parser-csv,embulk-parser-json
+com.fasterxml.jackson.core:jackson-core:2.6.7=deps
```
hiroyuki-sato, line 9: Why jackson **2.6.7** contains? Is it OK?

hiroyuki-sato (commented):
It seems `gradle.lockfile` contains jackson **2.6.7**. Could you check just in case?

gradle.lockfile
dmikurube: Jfyi, it would work even if these versions are different as they are loaded in different class loaders. Thanks to [EEP-7](https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0007.md) and [EEP-9](https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0009.md). ;)

dmikurube: #1663 is the upgrade of `embulk-deps`.

hiroyuki-sato: Thanks! Does it mean below?

* `gradle.lockfile` allow multiple jackson versions. (As a hard work result for Embulk v0.11)
* `embulk-deps` will load jackson 2.6.7 
* other plugins will load jackson 2.16.2

hiroyuki-sato (approved):
LGTM👍

gradle.lockfile
dmikurube: > `gradle.lockfile` allow multiple jackson versions. (As a hard work result for Embulk v0.11)

`gradle.lockfile` is just generated by Gradle. Gradle accepts different versions of the same library between `configuration`s, then `gradle.lockfile` allows multiple versions as a natural result.

`=embulk-decoder-bzip2,embulk-decoder-gzip,...` means that is used by the configurations embulk-decoder-bzip2, embulk-decoder-gzip, and else. `=deps` means that is used by the configuration (embulk-)deps.
```

> `embulk-deps` will load jackson 2.6.7

Yes, as of this revision.

> other plugins will load jackson 2.16.2

Yes for the "built-in" plugins. For other non-built-in plugins, they would have their own versions.

dmikurube: If you're interested, you may want to build `./gradlew executable` locally for this revision: 189bb879bb0a3363f77477e9f714d90cabe70b62

The result executable JAR includes multiple versions of Jackson, and its `META-INF/MANIFEST.MF` explains which uses what.

dmikurube:
Thanks!

gradle.lockfile
hiroyuki-sato: Thanks! I'll try it.

PR #1662, opened 2024-05-14
Set the ":checkDependencies" task to check for all projects

dmikurube:
Its `build.gradle` has had the `:checkDependencies` task to check dependencies (to fail when dependencies are invalid), but it has not iterated over subprojects. It was not very convenient.

This pull request makes the `:chedkDependencies` task to iterate over all subprojects.

(Note: we may consider applying this to other repositories, but only in this repository as of now. We may find another better approach.)


hiroyuki-sato (approved):
LGTM👍

PR #1674, opened 2024-06-17
EEP-0: Index of Embulk Enhancement Proposals (EEPs)

dmikurube:
We didn't have an index for EEPs while we now have 10 EEPs. It's inconvenient when we look into published EEPs.

Like [Python's PEP-0](https://peps.python.org/pep-0000/), it'd be nice to have EEP-0 as the index.

### DIFF
--- a/docs/eeps/eep-0000.md
+++ b/docs/eeps/eep-0000.md
@@ -39,7 +39,7 @@ EEP Status Key
 
 * A — Accepted: Normative proposal fixed and accepted for implementation, to be ready for implementation.
 * A — Active: Currently valid informational guidance, or an in-use process.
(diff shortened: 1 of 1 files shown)


hiroyuki-sato (approved):
LGTM👍

What do you think about copying [`PEP Type Key`](https://peps.python.org/pep-0000/#pep-types-key) and [`PEP Status Key`](https://peps.python.org/pep-0000/#pep-status-key) parts?
(I didn't understand Type and Status means)

dmikurube:
Thanks! Good idea, but may not be just a copy. Let me try to add.

dmikurube:
Added EEP Types Key and EEP Status Key based on the explanation in EEP-1.

hiroyuki-sato:
Thanks! LGTM👍

PR #1675, opened 2024-06-18
Resolve forward references in EEPs until EEP-10

dmikurube:
EEPs so far had some "forward references" for future EEPs, such as:

> ... will be explained in another EEP.

> ... is another effort to be explained in another EEP.

> ... will be explained in another upcoming EEP.

This pull request is to resolve those forward references in past EEPs.

### DIFF
--- a/docs/eeps/eep-0003.md
+++ b/docs/eeps/eep-0003.md
@@ -114,7 +114,7 @@ It may, or may not work unfortunately. It often works, but it can accidentally f
 
 This is a type of so-called leaky abstraction about dependency libraries. The Embulk core has used Jackson for itself, but the Jackson had been leaking to plugins.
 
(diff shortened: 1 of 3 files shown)


hiroyuki-sato (approved):
LGTM👍

Just an aside, It is necessary to add a link to the [EEP Index](https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0000.md) on www.embulk.org.

hiroyuki-sato:
Thanks!

This point was it is hard to find the EEP link in documents.
(www.embulk.org and dev.embulk.org)

I just found it in this page only.
https://www.embulk.org/articles/2023/03/10/embulk-maintenance-gets-open.html

Let's discuss another place.

PR #1676, opened 2024-06-19
Update the "Post-History" headers of updated EEPs

dmikurube:
Forgot to update the `Post-History` headers of the updated EEPs in #1675 

(`Post-History` is explained in https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0001.md#eep-header-preamble )


hiroyuki-sato (approved):
LGTM👍

PR #1681, opened 2024-09-11
Discard "parent_first_*.properties" files, and hard-code them in PluginClassLoader

dmikurube:
`parent_first_packages.properties` and `parent_first_resources.properties` have been used to suppress some classes from being loaded by plugin class loaders. However, it was not working properly, then we have planned to remove this mechanism, as declared in EEP-9:

https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0009.md#parent_first_packages-and-parent_first_resources

The time has come. As the first step, this pull request removes those files, and hard-code them directly in `PluginClassLoader`.

----

In addition, we needed `PluginClassLoader` to be more flexible so that the class loading mechanism could work with testing a plugin, hopefully with a new test helper `embulk-junit5`.


dmikurube:
Thanks!

PR #1683, opened 2024-09-12
Hard-code PluginClassLoader.isParentFirstPackage and isParentFirstPath

dmikurube:
This may seem like a stupid change, but intentional. I am going to make another change to check not only with `startsWith`, but some more detailed check. This is a preparation for the change.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1682, opened 2024-09-12
Test PluginClassLoader.isParentFirstPackage and PluginClassLoader.isParentFirstPath

dmikurube:
To be prepared for changing the logic of "parent_first", adding tests to detect unexpected behaviors.

PR #1684, opened 2024-09-12
Exclude some org.embulk.* packages from "parent_first"

dmikurube:
This is the goal of #1682 and #1683.

Some packages under "org.embulk." are intentionally excluded from "parent_first", for example,

- Packages for plugins, such as "org.embulk.input." and "org.embulk.parser."
- Packages under "org.embulk.base.", "org.embulk.deps.", and "org.embulk.util."


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1685, opened 2024-09-12
Exclude "msgpack/" resources from "parent_first"

dmikurube:
Resources under `msgpack/` have been included in "parent_first", but actually, the existing `msgpack-core` does not contain any resource under `msgpack/`, then the Embulk packages does not include any resource under `msgpack/`.

It's no longer needed. Let's remove it.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1686, opened 2024-09-12
Add PluginClassLoader#loadClassInThisClassLoader, for internal use

dmikurube:
In implementing a new testing framework for Embulk plugins, we needed a way to load a specified class forcibly in `PluginClassLoader`, with ignoring the parent class loader.

This is only a helper method for the testing framework.


hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1687, opened 2024-09-13
Fix the comment on ClassLoader#findLoadedClass(String) in PluginClassLoader

dmikurube:
Found that the comment in `PluginClassLoader#loadClass(String, boolean)` was wrong.

`ClassLoader#findLoadedClass(String)` does not find the class from its parent class loader indeed. I confirmed it at: https://gist.github.com/dmikurube/ac1480b4eeaf5e925ead3aec719a5441


dmikurube:
Follow-up to v0.11.5 at #1681, #1683, and #1684. Can you take a look? @embulk/core-team

hiroyuki-sato (approved):
LGTM👍

dmikurube:
Thanks!

PR #1703, opened 2026-09-17
release: Use existing `github.ref_name` instead of computing tag name manually

### LINKED ISSUE #1700: release: Use `GITHUB_REF_NAME` in `.github/workflows/publish.yml`
We can use `GITHUB_REF_NAME` instead of `${GITHUB_REF#refs/*/}` to get tag name.

kou:
Fix #1700

We also use more descriptive name than `TAG_NAME`: `RELEASE_VERSION_TAG`


komainu8 (approved):

kou:
@hiroyuki-sato Could you take a look at this?

hiroyuki-sato (approved):
LGTM👍
Please give me a little time. I'm checking with @dmikurube about the PR merge policy.


For my own reference:
`github.ref_name` is the short name of the branch, tag, or other ref that triggered the workflow.
`github.ref_name` is not specific to tags. For a branch push, it contains the branch name. For a pull request event, it may contain a value such as `1703/merge`.
However, this GitHub Actions workflow is configured to run only when a tag is pushed.
Therefore, in this workflow, we can assume that `github.ref_name` always contains the tag name.

For example, when the `v0.11.5` tag is pushed, `github.ref_name == "v0.11.5"`.

```
name: Publish
on:
  push:
    tags:
      - "v0.*"
```

kou:
Context: https://github.com/orgs/embulk/discussions/42

.github/workflows/publish.yml
dmikurube, line 15:
```suggestion
    env:
      # github.ref_name shoule be a tag because this workflow is triggered by: "on: push: tags:"
      RELEASE_VERSION_TAG: ${{ github.ref_name }}
    steps:
```

.github/workflows/publish.yml
```diff
@@ -34,11 +30,11 @@ jobs:
             const response = await github.rest.repos.createRelease({
               draft: false,
               generate_release_notes: false,
-              name: 'Release ${{ env.GIT_TAG }}',
+              name: 'Release ${{ github.ref_name }}',
```
dmikurube, line 33:
```suggestion
              name: `Release ${process.env.RELEASE_VERSION_TAG}`,              
```

.github/workflows/publish.yml
```diff
@@ -34,11 +30,11 @@ jobs:
...
               owner: context.repo.owner,
               prerelease: false,
               repo: context.repo.repo,
-              tag_name: '${{ env.GIT_TAG }}',
+              tag_name: '${{ github.ref_name }}',
```
dmikurube, line 37:
```suggestion
              tag_name: process.env.RELEASE_VERSION_TAG,
```

.github/workflows/publish.yml
```diff
@@ -56,7 +52,7 @@ jobs:
         script: |
           const fs = require('fs');
           const path = require('path');
-          const version = '${{ env.GIT_TAG }}'.substring(1);
+          const version = '${{ github.ref_name }}'.substring(1);
```
dmikurube, line 55:
```suggestion
          const version = process.env.RELEASE_VERSION_TAG.substring(1);
```

dmikurube (commented):
Thanks for your efforts on it!  I've read your Discussion at https://github.com/orgs/embulk/discussions/42 and replied there in addition to this PR. Let's try working together.

The PR looks correct to me, but let me suggest using it via (still) an environment variable (while not via GITHUB_REF like before the change) for the following reasons. 

1. Embedding `${ ... }` directly there has a potential risk about [injections](https://docs.github.com/en/actions/concepts/security/script-injections). Yeah, I know that this problem has been there even before this PR, and this has a very low risk because external contributors do not have the permissions to push tags, but this PR could be a good chance to address this at the same time.  
2. Using `github.ref_name` directly there would lose the context that this should be a tag, like @hiroyuki-sato -san commented above. Storing this as another "variable" named `RELEASE_VERSION_TAG` would help keeping this context.

dmikurube:
We may have an option to have a guard to check `process.env.GITHUB_REF_TYPE` is `'tag'`, but it could be too much...

kou:
If we prefer environment variable, we can use `process.env.GITHUB_REF_NAME` without defining our `RELEASE_VERSION_TAG`.

If we want a name for the tag variable, how about simpler `VERSION` instead of `RELEASE_VERSION_TAG`?
(`GITHUB_REF_NAME` is widely used in GitHub Actions. So I don't think that we need a new name for it.)

dmikurube:
I meant that, by using the variable name `RELEASE_VERSION_TAG`, all the following code readers (including coding agents) can understand that this should be a tag name, not a branch name, through reading only this part.

If `process.env.GITHUB_REF_NAME` or `VERSION` is used there, readers have to get back to the trigger (`on:`) part, and check that this workflow is triggered only by pushing tags, not pushing branches.

Both human beings and coding agents tend to read only the local parts that are related to their interest. Embedding the context in the local part would help everyone understanding the intention. On the other hand, shorter or simpler names are not helpful at all.

kou:
OK. I've applied your suggestions.

dmikurube (approved):
LGTM! Thanks for working on it.

btw, one comment about the Issue.

Filing an issue before submitting a pull request is a good custom. However, I believe that the issue should explain "the problem to solve", not just "what the upcoming pull request would do".

From the Issue title "release: Use `GITHUB_REF_NAME` in `.github/workflows/publish.yml`" and the description "We can use `GITHUB_REF_NAME` instead of `${GITHUB_REF#refs/*/}` to get tag name.", I understood "we can do it", but I didn't understand "why we should do it -- what is the problem with the existing way".

kou:
It makes sense. I've added "why" to existing my issues.

I've also rebased on main to resolve conflicts.

PR #1705, opened 2026-09-17
ci: use the latest Intel-based macOS

### LINKED ISSUE #1704: ci: cannot run check workflow on macOS 13
## Problem

Currently, the check workflow does not run in GitHub Actions, so we cannot run tests on macOS.

## Causes

The check workflow currently uses macOS 13.
However, the macOS 13 runner image is no longer available in GitHub Actions as of September 19, 2025.

See: https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

 ## Suggestion

We should use the latest version of macOS supported by GitHub Actions.

komainu8:
Fix: GH-1704

macOS 26 is the latest Intel-based macOS version supported by GitHub Actions, so we use it.

See: Available images on GitHub Actions
https://github.com/actions/runner-images/tree/main#available-images


komainu8:
Confirmed that this change passes with https://github.com/komainu8/embulk/actions/runs/35178786120.

kou (approved):
+1

@hiroyuki-sato Could you take a look at this?

hiroyuki-sato:
@komainu8 @kou Thank you for your work. Let's try it.

hiroyuki-sato (approved):
LGTM👍 (I'm waiting @dmikurube review before merge this)

kou:
Context: https://github.com/orgs/embulk/discussions/42

dmikurube (approved):
Thanks for your efforts on it! I've read your Discussion at https://github.com/orgs/embulk/discussions/42 and replied there in addition to this PR. Let's try working together.

This change looks good for the time being, so approving this. On the other hand, macos-26 would be deprecated very soon. So, `macos-26-intel` could be obsolete soon, too.

If we (you) restart the maintenance of Embulk, we may want to consider to give up testing with `macos`, or to give up building/testing with Java 8.

dmikurube:
> macos-26 would be deprecated very soon. So, `macos-26-intel` could be obsolete soon, too.

Sorry, it was my wrong understanding. macos-26 should be covered by security updates still for a while. I misunderstood this because I remembered that my Intel Macbook is going to lose its macos support soon...

However, I still think we should consider the following part.

> If we (you) restart the maintenance of Embulk, we may want to consider to give up testing with `macos`, or to give up building/testing with Java 8.

hiroyuki-sato:
@dmikurube Thank you for your review. Just to confirm, is it okay for me to merge this?
macOS 27 was just released, so macOS 26 will probably be supported for at least another year. Let's migrate to the new CI platform before then.

dmikurube:
Yeah, good to merge this, but I wanted that a discussion on the future macos or Java would happen.

hiroyuki-sato:
Thanks. Let's create a separate GitHub issue or discussion for this. Which would be better, a GitHub issue or a discussion?

In my opinion, if we can migrate to Java 11 within a year, we should move to the Apple Silicon platform. Otherwise, if we can't complete the migration before `macos-26-intel` is dropped, we should drop macOS CI.

(Another option)
It seems that Zulu supports Java 8 on Intel-based Macs. (I mentioned Zulu as one of the Java distributions other than Temurin.)
スクリーンショット 2026-09-24 10 36 57

komainu8:
@dmikurube @hiroyuki-sato 
Thank you for reviewing and merging.

> If we (you) restart the maintenance of Embulk, we may want to consider to give up testing with macos, or to give up building/testing with Java 8.

These are still open issues that have not yet been concluded, I feel it would be better to discuss them in the discussion thread.

dmikurube:
@hiroyuki-sato Under the context of https://github.com/orgs/embulk/discussions/42, I believe that such kind of discussions should be led by the ClearCode folks. ;)

@komainu8 Thanks! I'd suggest the "General" section in the Discussions here, but if you prefer another option, that'd be considered.

komainu8:
@dmikurube 
Thank you for the quick response. I’ll create it in the "General" section in the Discussions( https://github.com/orgs/embulk/discussions/categories/general ) today.

